PR #1, opened 2017-06-20
Initial project structure and CI routines

armiol:
This PR brings some initial project configuration, including Gradle build and CI/quality metrics configuration.

In particular

* Travis, Codecov and Codacy integration files created (including the encrypted sensitive data);
* `spine-model-compiler` and `spine-tools-common` were copied to this repository and renamed to fit.
* Gradle, Protobuf and gRPC versions were updated to the latest releases;
* `ext.gradle` now holds library versions, that are used across the modules and are meant to be the same;
* model compiler integration tests were updated to use the same versions from the `ext.gradle` file (see above).

In the next PRs the dependencies of the model compiler plugin should be switched to `io.spine:spine-base` artifact. Also, model compiler and tools-common modules will be removed from `tools` repo.


armiol:
@alexander-yevsyukov PTAL. This is also a "base" for your efforts on moving the modules from `core-java`.

build.gradle
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    task javadocJar(type: Jar, dependsOn: 'javadoc') {
+        from ("$projectDir/build/docs/javadoc")
+        classifier "javadoc"
+    }
+
+    apply from: filterInternalJavadocsScript
+
+    idea.module.iml {
```
alexander-yevsyukov: Can we have these several sections defining `idea.module` merged into one like this?

```groovy
idea {
    module {
        iml {
        }
    }
}
```

build.gradle
```diff
@@ -0,0 +1,271 @@
...
+
+    idea {
+        module {
+            downloadJavadoc = true
+            downloadSources = true
+        }
+    }
+}
+
+idea.project.ipr {
```
alexander-yevsyukov: A similar question applies here: can we have `project` under `idea`?

gradle-plugins/model-compiler/src/main/java/io/spine/gradle/compiler/Indent.java
```diff
@@ -0,0 +1,42 @@
...
+    private final int indent;
+
+    public Indent(int indent) {
+        this.indent = indent;
+    }
+
+    public String getJavaPoetIndent() {
+        final String result = Strings.repeat(" ", indent);
+        return result;
+    }
```
alexander-yevsyukov: How about having `hashCode()`, `equals()` in this class?

gradle-plugins/model-compiler/src/main/java/io/spine/gradle/compiler/Indent.java
```diff
@@ -0,0 +1,42 @@
...
+}
```
alexander-yevsyukov: Why do we have the `poet` thing in the method name? What if `toString()` does the job instead?

gradle-plugins/model-compiler/src/main/java/io/spine/gradle/compiler/validate/ClassNames.java
```diff
@@ -0,0 +1,134 @@
...
+import java.util.Collection;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static io.spine.gradle.compiler.message.fieldtype.ProtoScalarType.getBoxedScalarPrimitive;
+import static io.spine.gradle.compiler.message.fieldtype.ProtoScalarType.getJavaTypeName;
+import static io.spine.util.Exceptions.newIllegalArgumentException;
+
+/**
+ * Utility class for working with the {@code ClassName}s.
+ */
```
alexander-yevsyukov: Who's the author?

gradle-plugins/model-compiler/src/main/java/io/spine/gradle/compiler/validate/MetadataAssembler.java
```diff
@@ -0,0 +1,339 @@
...
+        final Set<VBMetadata> fieldMetadata = obtainFieldMetadata(metadataItems);
+        result.addAll(fieldMetadata);
+        result.addAll(metadataItems);
+        log().debug("The metadata is obtained, {} validating builder(s) will be generated.",
+                    fieldMetadata.size());
+        return result;
+    }
+
+    @SuppressWarnings("MethodWithMultipleLoops")    // It's fine, as for the utility plugin.
+    private Set<VBMetadata> obtainFileMetadata(
```
alexander-yevsyukov: Can the method decl. fit into one line here?

gradle-plugins/model-compiler/src/main/java/io/spine/gradle/compiler/validate/MetadataAssembler.java
```diff
@@ -0,0 +1,339 @@
...
+            result.addAll(metadataSet);
+
+        }
+        log().trace("The file-level metadata is obtained.");
+        return result;
+    }
+
+    @SuppressWarnings("MethodWithMultipleLoops")    // It's fine, as for the utility plugin.
+    private Set<VBMetadata> obtainFieldMetadata(
+            Iterable<VBMetadata> metadataSet) {
```
alexander-yevsyukov: Can the method decl. fit into one line here?

gradle-plugins/model-compiler/src/main/java/io/spine/gradle/compiler/validate/MetadataAssembler.java
```diff
@@ -0,0 +1,339 @@
...
+        log().trace("Obtaining the metadata for the validating builders, " +
+                            "which will be generated for the Message fields.");
+        final Set<VBMetadata> result = newHashSet();
+
+        for (VBMetadata metadata : metadataSet) {
+
```
alexander-yevsyukov: Do we really need this empty line here?

gradle-plugins/model-compiler/src/main/java/io/spine/gradle/compiler/util/UnknownOptions.java
```diff
@@ -0,0 +1,261 @@
...
+ * Currently, we do not need options with several properties.
+ * So, only first option property values are obtained, and others are ignored.
+ *
+ * @author Alexander Litus
+ */
+public class UnknownOptions {
+
+    private static final Pattern PATTERN_COLON = Pattern.compile(":");
+    @SuppressWarnings("HardcodedLineSeparator")
+    private static final Pattern PATTERN_NEW_LINE = Pattern.compile("\n");
```
armiol: No the issues aren't related. It's just a simple way to split the multiline string into lines. 

Basically, `String.split(regexString)` does the same, but compiles the `regexString` into `Pattern` internally each time. Instead of doing that, we just compile the pattern ourselves, but once and for all.

Interestingly, whatever solution we use (including Guava's `Splitter`) we would end up with `"\n"` in the codebase anyway. So the suppression makes sense to me.

armiol:
@alexander-yevsyukov the issues are addressed. PTAL again.

gradle-plugins/model-compiler/src/main/java/io/spine/gradle/compiler/Indent.java
```diff
@@ -35,8 +36,26 @@ public Indent(int indent) {
...
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        Indent indent1 = (Indent) o;
```
alexander-yevsyukov: Please rename the variable to `other`.

PR #3, opened 2017-06-23
Migrate to `spine-base` dependency instead of `spine-client-core` + `ThrowableMessage`

armiol:
This PR addresses a few things.

1. The plugins no longer depend on `spine-client-core`. Instead, they depend `spine-base`.
2. The test code depends on `spine-testutil-base` instead of `spine-testutil-core`.
3. `ThrowableMessage` is used for code generation instead of `FailureThrowable`.
4. The generation folder for Spine failures is now set to `... /generated/ (main|test) /spine` instead of ` ... /generated/ (main or test) /java`.
5. The dependency on `tools` module is set to `0.9.25-SNAPSHOT` (the one, where there is no model compiler already).

The version for `base` modules is increased to `0.9.29-SNAPSHOT`.

### DIFF
--- a/ext.gradle
+++ b/ext.gradle
@@ -25,7 +25,7 @@
  *  as we want to manage the versions in a single source.
  */
 
-def final SPINE_VERSION = '0.9.25-SNAPSHOT'
+def final SPINE_VERSION = '0.9.29-SNAPSHOT'
 
 ext {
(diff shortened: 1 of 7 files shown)


armiol:
@alexander-yevsyukov PTAL.

PR #4, opened 2017-06-27
Code from `core`

alexander-yevsyukov:
This PR adds the code of the `base` module, which used to be a part of `core-java` repository.

The version of Spine in this PR is one number ahead from that in `core-java` (the related PR there is: https://github.com/SpineEventEngine/core-java/pull/532).

It done so to allow dependency on `spine-base` from `spine-core` with a following PR in `core-java`.


alexander-yevsyukov:
@armiol, pleases pick it up from this point. The build of `model-compiler` does not pass because it depends on `spineVersion` which is not yet published.

PR #8, opened 2017-07-10
Move `options` to the top of the proto files.

alexander-aleksandrov:
Move
```protobuf
import "spine/options.proto";
``` 
at the top of the import section in proto files.


alexander-aleksandrov:
@armiol PTAL

PR #15, opened 2017-07-24
Rename `Failure` to `Rejection`

alexander-yevsyukov:
This PR makes the Spine Framework use the term `Rejection` instead of `Failure` to avoid the terminology clash with the term from [Reactive Manifesto](http://www.reactivemanifesto.org/glossary#Failure).


alexander-yevsyukov:
@armiol, PTAL.

PR #17, opened 2017-07-25
Logging utilities

alexander-yevsyukov:
This PR introduces logging utilities based on the state of the `Environment`. The `Environment` class was copied from `core-java`, and is going to replace the class used in that sub-project.


alexander-yevsyukov:
@armiol, PTAL.

PR #33, opened 2017-07-27
Small base fixes and improvements

alexander-yevsyukov:
This PR introduces the following improvements:
 * `MessageClass.toString()` now return class name of the enclosed class instead of `String.valueOf()`. This caused annoying `class` prefix in the returned string.
* `Messages` got `ensureMessage()` utility.
* Fixed IDEA-Gradle integration issue in the `plugin-base` module.


alexander-yevsyukov:
@armiol, PTAL.

PR #34, opened 2017-07-31
Rename IDEA project

dmdashenkov:
Currently, we have both IDEA projects `base` and `core-java` with the name `spine`. This is both confusing and uncomfortable when navigating through the projects.

Let's call this project `base` in consistency with the root directory and git repo name.


dmdashenkov:
@armiol, PTAL.

.idea/.name
```diff
@@ -1 +0,0 @@
-spine
```
armiol: This should be `spine-base`.

PR #35, opened 2017-08-02
`MessageField` and `ThrowableMessage` improvements

alexander-yevsyukov:
This PR
  * Introduces `MessageFieldException` is introduced, which improves diagnostics for error cases.
  * Adds the ability to set producer ID for a `ThrowableMessage`.

### DIFF
--- a/base/src/main/java/io/spine/protobuf/MessageFieldException.java
+++ b/base/src/main/java/io/spine/protobuf/MessageFieldException.java
@@ -1,0 +1,56 @@
+package io.spine.protobuf;
+
+import com.google.protobuf.Message;
+
+import javax.annotation.Nullable;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * Signals an error working with a protobuf message field.
+ *
+ * @author Alexander Yevsyukov
+ */
+public class MessageFieldException extends RuntimeException {
+
+    private final Message protobufMessage;
+
+    /**
+     * Constructs a {@code MessageFieldException} with the formatted message text.
+     *
+     * @param protobufMessage    a Protobuf message object working with a field of which caused
+     *                           an error
+     * @param errorMessageFormat a formatted string for error message
+     * @param params             error message parameters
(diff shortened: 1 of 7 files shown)


alexander-yevsyukov:
@armiol, PTAL.

base/src/main/java/io/spine/protobuf/MessageFieldException.java
```diff
@@ -0,0 +1,56 @@
+package io.spine.protobuf;
+
+import com.google.protobuf.Message;
+
+import javax.annotation.Nullable;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * Signals an error working with a protobuf message field.
```
armiol: Let's call it `Protobuf`, with a capital.

base/src/main/java/io/spine/protobuf/MessageFieldException.java
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * Obtains a Protobuf message working with a field of which caused an error
+     */
+    public Message getProtobufMessage() {
+        return protobufMessage;
+    }
+
+    private static String formatMessage(String format, @Nullable Object... params) {
+        if (params == null) {
```
armiol: This method is reported to be only partially tested.

base/src/main/java/io/spine/protobuf/MessageFieldException.java
```diff
@@ -0,0 +1,56 @@
...
+public class MessageFieldException extends RuntimeException {
+
+    private final Message protobufMessage;
+
+    /**
+     * Constructs a {@code MessageFieldException} with the formatted message text.
+     *
+     * @param protobufMessage    a Protobuf message object working with a field of which caused
+     *                           an error
+     * @param errorMessageFormat a formatted string for error message
```
armiol: I would not call that a `formatted string`. That's a string template used for error message formatting.

PR #38, opened 2017-08-04
Fix typename diagnostics

alexander-yevsyukov:
This PR:
  * Improves diagnostic messages in the `TypeUrl` class.
  * Temporarily disables some plug-in tests that result in out-of-memory build errors.


alexander-yevsyukov:
@armiol, PTAL.

.idea/.name
```diff
@@ -1 +1 @@
-spine-base
\ No newline at end of file
+spine
```
armiol: I think the change in this file should be reverted, as we have renamed the root project to `spine-base` intentionally.

base/src/main/java/io/spine/protobuf/MessageFieldException.java
```diff
@@ -12,28 +13,31 @@
...
+     *        a format string for the error message
+     * @param params
+     *        error message parameters
      */
     public MessageFieldException(Message protobufMessage,
                                  String errorMessageFormat,
                                  Object... params) {
         super(format(checkNotNull(errorMessageFormat), params));
-        this.protobufMessage = checkNotNull(protobufMessage);
+        this.protobufMessage = (GeneratedMessageV3)checkNotNull(protobufMessage);
```
armiol: A single-space character seems to be missing after `(GeneratedMessageV3)`.

base/src/main/java/io/spine/type/KnownTypes.java
```diff
@@ -253,13 +253,17 @@ public boolean apply(@Nullable TypeUrl input) {
      * @throws IllegalArgumentException if the name does not correspond to any known type
      */
     static GenericDescriptor getDescriptor(String typeName) {
+        checkNotNull(typeName);
+        checkArgument(!typeName.isEmpty(), "Type name cannot be empty");
```
armiol: Seems that `Validate.checkNotEmptyOrBlank` may fit better than both of `check...` statements.

Issue #39, opened 2017-08-07
JaCoCo should collect coverage data from other JVMs


dmytro-grankin:
`JaCoCo` currently can collect coverage data only from a single JVM. If during a test execution other JVMs are spawned, the code executed in these JVMs is not included to a coverage report.

We should provide a tool to help `JaCoCo` collect coverage data from all JVMs, which were created during a test execution.

This feature is required, because the test code of `Gradle` plugins spawns JMVs (see `GradleProject`), as a result, the plugins code is reported uncovered.

PR #40, opened 2017-08-07
Decrease the memory consumption during the test execution

dmytro-grankin:
This PR reduces the memory consumption during execution of tests for `Gradle` plugins.

Configuration of `GradleProject`  was changed to reduce the memory consumption. But in the opposite side, the test coverage was plenty decreased. For more details, please take a look [here](https://github.com/SpineEventEngine/base/issues/39).

Also, if you want to debug the code executed by `GradleProject`, you should manually set `GradleProject.DEBUG_ENABLED` to `true` for a debug session.


dmytro-grankin:
@armiol PTAL.

PR #42, opened 2017-08-10
Abstract class value objects base

alexander-yevsyukov:
This PR:
 * Introduces `ClassTypeValue` as the abstract base for classes that hold a `Class` value.
 * Reduces string duplication.


alexander-yevsyukov:
@armiol, PTAL.

PR #43, opened 2017-08-11
Specific interfaces for messages

dmdashenkov:
This PR addresses issue #19.

The mechanism is as follows:
 - `protoc` plugin generates marker interfaces derived from `Message` by the `(every_is)` and `(is)` options;
 - a Gradle plugin enables the Proto compiler and the compiler plugin integration (with Protobug Gradle Plugin).

After applying Spine Model Compiler plugin, the user of the framework will not need to configure the `protobuf { }` anymore (see `protoc_config.gradle`).

PR #44, opened 2017-08-16
Protoc plugin module

dmdashenkov:
This PR is a part of #43 PR which introduces the marker interface generation upon `(is)` and `(every_is)` options.

In order to be run and tested on CI properly, it is split into two separate pull requests.


dmdashenkov:
@armiol, PTAL.

PR #52, opened 2017-10-11
Update and cleanup before `0.10.0` release

armiol:
This PR updates the `base` version before the upcoming `0.10.0` release.

Also, it addresses an issue, raised in C++ Spine client library. `Currency.BSD` enum value causes a compile-time conflict with system header `sys/param.h`, which also defines `BSD` in some *nix operating systems. For instance:
```cpp
//...
#define	BSD	199103		/* March, 1991 system version (year & month) */
// ...
```
(_taken from a [random `sys/param.h`](http://unix.superglobalmegacorp.com/Net2/newsrc/sys/param.h.html) in illustrative purposes_)

Therefore `Currency.BSD` was renamed to `Currency.BSD_`.

The library version has been set to `0.9.77-SNAPSHOT`.

### DIFF
--- a/base/src/main/proto/spine/money/money.proto
+++ b/base/src/main/proto/spine/money/money.proto
@@ -178,8 +178,14 @@ enum Currency {
         numeric_code: 986,
         exponent_digits: 2 }];
 
-    // This identifier caused naming conflicts in C++ version of Spine client.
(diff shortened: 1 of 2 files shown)


armiol:
@alexander-yevsyukov PTAL.

armiol:
@alexander-yevsyukov issues addressed. Please take a look again.

PR #53, opened 2017-11-06
Fix Spine directory name

dmdashenkov:
Fix the `.spine` directory generation name. Because of an error, it was generated as `..spine` (with 2 dot symbols instead of 1).

Also, in this PR we change the logging level in the `Extension` getters and setters from `trace` to `debug`, since `trace` is not visible from the command line.


dmdashenkov:
@armiol, PTAL.

PR #54, opened 2017-11-13
Fix Validating Builder generation

dmdashenkov:
Before this PR we generated the `ValidatingBuilder`s for both the types from `.proto` definitions and from types of fields in that definitions. Now it's only the `.proto` definitions.

This is done to ensure unambiguity of the resulting `...VBuilder` class.


dmdashenkov:
@armiol, PTAL.

PR #68, opened 2018-02-07
Fix `(is)` option description

### LINKED ISSUE #66: Make options `is` and `every_is` multi-platform
The documentation of the `is` option now says:

```proto
// Specifies the name of a Java interface to be implemented by this message.
...
```

Although this is true in the current implementation, this is not exactly what we want to do when defining a type. We way to say that a type annotated in this way is a specific kind of message _and_ for Java this fact would be reflected in generated code in Java-specific way.

When we have code generation implemented for another programming language, the option description would be extended for that programming language.

The Model Compiler should have an API that would allow to match values from the `is` options to a value specific for the programming language. For Java it would a fully-qualified interface name.

dmdashenkov:
Until now, we had a Java-oriented documentation of the `(is)` and `(every_is)` options.

Now the documentation is changed to explain the role of the options when describing the model, rather than the role in code generation.

The Java-oriented doc part persists but is made secondary.

Fixes #66.


dmdashenkov:
@alexander-yevsyukov, PTAL.

base/src/main/proto/spine/options.proto
```diff
@@ -363,15 +363,28 @@ extend google.protobuf.MessageOptions {
...
-    // Specifies the name of a Java interface to be implemented by this message.
+    // Specifies a characteristic inherent in the the given message type.
     //
-    // The interface is generated into the Protobuf generation destination directory. The interface
-    // itself is a marker interface derived from `com.google.protobuf.Message`. Unless the interface
-    // is specified by its FQN, it will have the same Java package as the message.
+    // Example:
+    // ```
+    // message CreateProject {
+    //     option (is) = "CommandMessage";
```
alexander-yevsyukov: `CommandMessage` is going to be a built-in interface. How about having `ProjectCommand` instead?

base/src/main/proto/spine/options.proto
```diff
@@ -478,15 +491,28 @@ extend google.protobuf.FileOptions {
...
 
-    // Specifies the name of a Java interface to be implemented by all the messages in this file.
+    // Specifies a characteristic common for all the message types in the given file.
     //
-    // The interface is generated into the Protobuf generation destination directory. The interface
-    // itself is a marker interface derived from `com.google.protobuf.Message`. Unless the interface
-    // is specified by its FQN, it will have the same Java package as the message.
+    // Example:
+    // ```
+    // option (is) = "CommandMessage";
```
alexander-yevsyukov: Again, let's have `ProjectCommand` instead.

Issue #67, opened 2018-02-06
Provide missing package descriptions


alexander-yevsyukov:

dmytro-grankin:
Closing the issue since it was resolved in #70.

PR #70, opened 2018-02-07
Code generation improvements

alexander-yevsyukov:
This PR introduces the following changes.

### Code structure improvements and fixes
  * A `codegen` module with common code-generation classes was introduced.
  * Utility classes for working with code-related names were refactored with introduction of value objects `FieldName` (for both `java` and `proto`), `PackageName` (for `java`), `SimpleClassName`, etc.
  * Fixed missing package descriptions (#67) and non-nullity annotations.
  * Fixed data isolation when passing collections as constructor arguments.

### Interfaces for message types
The following interfaces were introduced:
  * `CommandMessage`
  * `EventMessage`
  * `RejectionMessage`.
Using these interfaces in code generation will be done in one of the following PRs.

### Other notable changes
  * `StringValueType` was made public in order to be used as the base for code name classes.
  * `CodePreconditions` class was introduced for avoiding dependency on the `Validate` class (which did not fit the need fully anyway).

### Dependency changes
   * Protobuf Gradle Plugin -> `0.8.4`
   * Protobuf -> `3.5.1` 
   * gRPC -> `1.9.0`

Spine version advanced to `0.10.24-SNAPSHOT`.


alexander-yevsyukov:
@dmdashenkov, please add description of the steps required to advance the version in `ext.gradle`.

base/src/main/java/io/spine/type/CommandMessage.java
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     * The name suffix for proto files containing command message declarations.
+     */
+    String PROTO_FILE_SUFFIX = "commands.proto";
+
+    /**
+     * Returns {@code true} if the passed file defines command messages, {@code false} otherwise.
+     */
+    Predicate<FileDescriptor> FILE_PREDICATE = new Predicate<FileDescriptor>() {
```
dmdashenkov: Let's not put the file predicate into the interface itself.
The interface is a public API and the framework users do not want to know how do we find those files.
This is also relevant to the `PROTO_FILE_SUFFIX` constant.

dmdashenkov: Please apply here and in other interfaces.

codegen/src/main/java/io/spine/tools/java/FileName.java
```diff
@@ -0,0 +1,85 @@
...
+    private static final String EXTENSION = ".java";
+
+    private FileName(String value) {
+        super(value);
+    }
+
+    /**
+     * Obtains the name for the passed type.
+     */
+    public static FileName forType(String typeName) {
```
dmdashenkov: Since it is a **Java** file name, can we use `Class` instead of `String`?

If so, the implementation may become not so simplistic (e.g. support for inner types etc.).

dmdashenkov: If the class is not yet generated when referring to its name, this comment should not be applied.

codegen/src/main/java/io/spine/tools/proto/FieldName.java
```diff
@@ -0,0 +1,87 @@
...
+     */
+    public String toCamelCase() {
+        final String result = CamelCase.convert(this);
+        return result;
+    }
+
+    /**
+     * Obtains the field name in {@code javaCase}.
+     */
+    public String javaCase() {
```
dmdashenkov: Can we have consistent names for `toCamelCase()` and `javaCase()`? I believe that it's obvious for a reader without such an example.

codegen/src/test/java/io/spine/tools/proto/UnderscoredNameShould.java
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * @author Alexander Yevsyukov
+ */
+public class UnderscoredNameShould {
+
+    @Test
+    public void have_utility_ctor_for_CamelCase() {
+        Tests.assertHasPrivateParameterlessCtor(UnderscoredName.CamelCase.class);
+    }
+
```
armiol: This line is redundant.

codegen/src/main/java/io/spine/tools/java/FileName.java
alexander-yevsyukov: Since we're dealing mostly with code generation, we don't have a class yet.

codegen/src/main/java/io/spine/tools/proto/FieldName.java
alexander-yevsyukov: It's done intentionally, since in `toCamelCase` we have `CamelCase`, which starts from capital `C`, and in `javaCase()`, it's small `j`.

alexander-yevsyukov:
@armiol, PTAL.

base/src/main/java/io/spine/type/CommandMessage.java
```diff
@@ -0,0 +1,74 @@
+/*
+ * Copyright 2017, TeamDev Ltd. All rights reserved.
```
dmytro-grankin: For some newly added files, a copyright year is still `2017`.

gradle-plugins/model-compiler/src/main/java/io/spine/gradle/compiler/protoc/package-info.java
```diff
@@ -0,0 +1,28 @@
...
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * This package provides {@link io.spine.gradle.compiler.annotation.ProtoAnnotatorPlugin
+ * ProtoAnnotatorPlugin}.
+ */
```
dmytro-grankin: This package provides `ProtocPluginImporter`.

PR #75, opened 2018-02-07
Add message types

alexander-yevsyukov:
This PR introduces specific interfaces for message types.


alexander-yevsyukov:
@dmdashenkov, this is what we've discussed today. But let's finish #70 first.

PR #77, opened 2018-02-08
Describe version changing

dmdashenkov:
This PR contains changes to docs in `ext.gradle`.

Now, we have an explanation of how to increment the version.


dmdashenkov:
@alexander-yevsyukov, PTAL.

PR #87, opened 2018-02-23
Extract date/time-related code to separate module

mdrachuk:
This PR removes the types and services related to date and time from the `spine-base`.

The code is extracted to [`spine-time`](https://github.com/SpineEventEngine/time) repository.


mdrachuk:
@alexander-yevsyukov PTAL.

base/src/main/java/io/spine/validate/MessageFieldValidator.java
```diff
@@ -131,6 +131,14 @@ private static boolean isTimeInvalid(Timestamp timeToCheck, Time whenExpected, T
         return isInvalid;
     }
 
+    private static boolean isLaterThan(Timestamp t1, Timestamp t2) {
```
alexander-yevsyukov: Please have this method in `Timestamps` utility file.

Please also have `fromMillis()` also in this utility class.

mdrachuk:
This was rethought and redone in PR #88 to prevent history loss.

PR #88, opened 2018-02-23
Extract date/time-related code to separate module

mdrachuk:
This PR removes the types and services related to date and time from `spine-base`. 

The code is extracted to [`spine-time`](https://github.com/SpineEventEngine/time) repository.

`Time.java` itself is moved to `io.spine.base` and is restricted in functionality to only providing time. This is done to leave a simple way to substitute system clock for tests.


scripts/upload-artifacts.sh
```diff
@@ -6,8 +6,13 @@
...
 # Prepare the test and coverage reports for the upload.
 mkdir reports
-zip -r reports/test-reports.zip **/build/reports
-zip -r reports/jacoco-reports.zip **/build/jacoco
+
+# Find all directories matching path to add to archive. 
+BUILD_REPORTS=$(find . -type d -path "*build/reports*")
+JACOCO_REPORTS=$(find . -type d -path "*build/jacoco*")
+
+zip -r reports/test-reports.zip $BUILD_REPORTS
```
mdrachuk: From what it looks like, previously it was intended to upload the reports for all subprojects. But it consumed only the reports for `codegen` and `base`. Seems like the globstar (`**`) didn’t work as intended — it depends on the shell environment and its settings. 
Instead, a search with `find` is performed now picking up all the build reports to add to the archive.

mdrachuk:
@alexander-yevsyukov PTAL.

PR #93, opened 2018-03-09
Fix Gradle scripts

dmdashenkov:
Currently, the generated `spine-protoc.gradle` script has errors in it. By an (un-)happy coincidence, `core-java` still works with those errors. Though, when changing the `core-java` version, the build fails.

This PR fixes errors in generating the `spine-protoc.gradle` file.

Also, in this PR we migrate the Gradle distribution from `bin` to `all`, since the lack of programming errors is more important than the build time.


dmdashenkov:
@armiol, PTAL.

PR #94, opened 2018-03-09
Update dependencies versions

dmdashenkov:
In this PR we update the Spine plugin versions.

This PR overrides the `0.10.30-SNAPSHOT` version.


dmdashenkov:
@armiol, PTAL.

PR #95, opened 2018-03-10
More improvements in code generation

alexander-yevsyukov:
This PR removes classes that are used only for code generation from the public API.

Other notable changes:
 * Names of resources generated by Spine plug-ins are now gathered under the `Resources` class to avoid constant duplication under `base` and `core-java`.
 * `debug` level is used instead of `trace` since `debug` is the minimum level which can be set for Gradle.
* Enrichment map generation was refactored to multiple classes.
* `OptionParser` was renamed to `OptionValue` with corresponding changes to the descendants and method names.

Spine version advanced to `0.10.31-SNAPSHOT`.


alexander-yevsyukov:
@armiol, PTAL.

base/src/main/java/io/spine/util/PropertyFiles.java
```diff
@@ -94,17 +89,28 @@ private static Properties loadPropertiesFile(URL resourceUrl) {
...
         return properties;
     }
 
     private static ClassLoader getContextClassLoader() {
-        return Thread.currentThread().getContextClassLoader();
+        return Thread.currentThread()
+                     .getContextClassLoader();
+    }
+
+    private static void warn(Throwable e, String errorFormat, Object... params) {
```
armiol: This looks a part of some utility.

codegen/src/main/java/io/spine/tools/properties/PropertiesWriter.java
```diff
@@ -54,11 +54,17 @@ public PropertiesWriter(String folder, String fileName) {
     }
 
     /**
-     * Updates the {@code .properties} file rewriting its contents if it already exists.
+     * Updates the content of the file with the passed map.
      *
-     * @param propertiesMap a map containing properties to write to the file
+     * <p>If the file already exists, the passed entries will be merged with the file content.
+     * If the file already has an key from the passed map, its value will <strong>NOT</strong>
```
armiol: We tend to use `<b>` and `<strong>` for the same goal. Let's settle on `<b>`, as it's shorter.

gradle-plugins/model-compiler/src/main/java/io/spine/tools/compiler/enrichment/ByOption.java
```diff
@@ -0,0 +1,153 @@
...
+     * the given field.
+     */
+    private List<String> parse() {
+        final List<FieldReference> fieldRefs = FieldReference.allFrom(field);
+
+        final ImmutableList.Builder<String> result = ImmutableList.builder();
+        for (FieldReference fieldRef : fieldRefs) {
+            if (fieldRef.isWildcard() && fieldRefs.size() > 1) {
+                // Multiple argument `by` annotation can not contain wildcard reference onto
+                // the event type if the type was not specified with a `enrichment_for` annotation
```
armiol: The dot is missing at the end of this sentence.

gradle-plugins/model-compiler/src/main/java/io/spine/tools/compiler/enrichment/EnrichmentMap.java
```diff
@@ -0,0 +1,226 @@
...
+    private final TypeNameValue enrichmentType;
+
+    EnrichmentMap(String packagePrefix) {
+        this.packagePrefix = packagePrefix;
+        this.eventType = new TypeNameValue(enrichmentFor, packagePrefix);
+        this.enrichmentType = new TypeNameValue(enrichment, packagePrefix);
+    }
+
+    /**
+     * Obtains enrichment information found in the passed messages.
```
armiol: In scope of clarifying things for the ease of writing the documentation, it would be good to describe what are the keys and values.

codegen/src/main/java/io/spine/tools/properties/PropertiesWriter.java
alexander-yevsyukov: Using `<em>` is more correct in this case as it's a single word in the sentence.

base/src/main/java/io/spine/util/PropertyFiles.java
alexander-yevsyukov: Well spotted. I've moved it to `Logging`.

PR #99, opened 2018-03-19
Move tools back

alexander-yevsyukov:
*WIP*


alexander-yevsyukov:
@armiol, PTAL.

tools/codegen/src/main/java/io/spine/tools/proto/TypeName.java
```diff
@@ -0,0 +1,88 @@
...
+ * A fully-qualified Protobuf type name.
+ *
+ * @author Alexander Yevsyukov
+ */
+public class TypeName extends StringTypeValue {
+
+    /**
+     * The separator character for package names in a fully qualified proto type name.
+     */
+    static final char PACKAGE_SEPARATOR = '.';
```
armiol: Not sure, why this constant and the next one are package-private. I would make them private and have a static accessors, if those are really needed somewhere else in the package.

tools/gradle-plugins/gcs-plugin/src/main/java/io/spine/tools/gcs/GcsPlugin.java
```diff
@@ -0,0 +1,63 @@
+/*
+ * Copyright 2017, TeamDev Ltd. All rights reserved.
```
armiol: In some of files there is still `2017`, while it should be `2018`.

PR #106, opened 2018-05-08
Performance Tuning

dmdashenkov:
In this PR we change the way `Json` tool assembles its `TypeRegistry`.

Before, the registry was assembled by calling `TypeUrl.getDescriptor()` for all the known types. The issue with this approach is that each call to `TypeUrl.getDescriptor()` implies a reflective call.

Now, we assemble the type registry from the `.desc` file, which is constructed when compiling `.proto`s. To do that, we add the descriptor files into classpath (by placing them into resources) and parse those files on application startup.

## Motivation

The main (and only) motivation for this change is performance. On a _not-such-a-big_ domain model, the process of assembling the type registry takes a few (~5) seconds on an AppEngine instance (B4). Although this change clips this time up to 2-3 seconds, this place is still a potential improvement point.

The next bottleneck in assembling the type registry for `Json` utility is parsing the `DescriptorProto`s into `Descriptor`s.

## Side Effect Changes

The `codegen` module is merged into the `base` module.

## Misc

In this PR some usages of Guava substitutes are replaced with the Java 8 API usages (such as `Stream`s, `Functuion`s, etc.).


dmdashenkov:
@armiol, PTAL.

The CI builds are to be restored soon.

dmdashenkov:
@armiol, PTAL.

The `codegen` module is inlined into the `base` module under package `io.spine.tools`.

dmdashenkov:
Closing this PR as it's a part of #125.

PR #115, opened 2018-05-16
Introduce Error Prone

alexander-yevsyukov:
This PR adds Error Prone to the project as
 1. `javac` plug-in;
 2. annotations library.

All `package-info.java` were annotated with `@CheckReturnValue`.
Minor code adjustments were made to satisfy the `CheckReturnValue` bug pattern checker.
Most notable modification is  using `ExpectedException` in tests instead of annotation value with an exception class.


alexander-yevsyukov:
@armiol, PTAL.

base/src/main/java/io/spine/util/Exceptions.java
```diff
@@ -59,6 +61,7 @@ private Exceptions() {
      * @return nothing ever
      * @throws UnsupportedOperationException always
      */
+    @CanIgnoreReturnValue
```
armiol: That's odd. IDEA gives a warning if the returned exception is not thrown:
```java
...
Exceptions.unsupported(...);   // the warning is emitted.

...
throw Exceptions.unsupported(...); // It's fine.
```
I don't know if `throw` counts as a usage of a returning value, but to my mind it should.

Therefore I wouldn't annotate methods like this one with `@CanIgnoreReturnValue`.

armiol:
@alexander-yevsyukov the rest LGTM.

alexander-yevsyukov:
@armiol, I guess you're right. I've reverted throwing utils.

alexander-yevsyukov:
@armiol, please merge on my behalf after the build is done.

PR #116, opened 2018-05-17
Fix Protobuf Compiler Plugin on Windows

dmdashenkov:
This PR brings a hotfix of a crash in the protoc plugin when running on the Windows OS.

The crash is related to the path format used in Windows. From now, we ensure that the path is in the Unix-like (URI-style) format (with `/` instead of `\`).

Also, in this PR we delete a generated Gradle script which has got to the Git repository accidentally.


dmdashenkov:
@armiol, PTAL.

PR #118, opened 2018-05-27
Improve treating returned values

alexander-yevsyukov:
This PR improves the way we treat returned values in response to introducing Error Prone checking.

Utility `Files2` was also introduced.

Spine version advances to `0.10.39-SNAPSHOT`.


alexander-yevsyukov:
@armiol, PTAL.

base/src/main/java/io/spine/io/Files2.java
```diff
@@ -0,0 +1,86 @@
...
+import com.google.errorprone.annotations.CanIgnoreReturnValue;
+
+import java.io.File;
+import java.io.IOException;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.io.Files.createParentDirs;
+
+/**
+ * Additional utilities for working with files.
```
armiol: Together with this wording, I'd append another sentence or two on why the utilities listed below aren't enough for us.

I am trying to get the potential confusion addressed. I.e. "why for the rest of the world those utilities are enough and we have some very specific scenarios, that we need such an extension".

PR #120, opened 2018-05-29
Add class storing common display names for JUnit 5 tests

dmitrykuzmin:
This PR adds storage for the common JUnit 5 test [display names](https://junit.org/junit5/docs/5.0.3/api/org/junit/jupiter/api/DisplayName.html) to the `testlib` module.

This class can provide assistance in avoiding string literal duplication when assigning `@DisplayName` to the common test cases like null tolerance check or utility constructor check.

`DisplayNames` is represented as a class and not an enumeration because `@DisplayName` annotation requires constant `String` expression as its parameter, so we can't retrieve such expression from enum constant in any reasonable way.

Spine version advances to `0.10.40-SNAPSHOT`.

### DIFF
--- a/ext.gradle
+++ b/ext.gradle
@@ -38,7 +38,7 @@
  *
  * This is also the version which the artifacts are published with.
  */
-def final SPINE_VERSION = '0.10.39-SNAPSHOT'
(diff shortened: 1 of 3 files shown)


dmitrykuzmin:
@armiol PTAL.

PR #122, opened 2018-05-31
Delete GCS Plugin

dmdashenkov:
In this PR we delete the Gradle `gcloud-storage` plugin, a.k.a. Spine GCS plugin.

The plugin has become obsolete, as there is a way to clean up a GCS bucket by configuring the bucket itself. See [this doc](https://cloud.google.com/storage/docs/lifecycle) for details.

The process of uploading the artifacts to GCS is left untouched.

The `spine-dev.appspot.com` bucket is configured to delete all objects older than 14 days.


dmdashenkov:
@armiol, PTAL.

PR #125, opened 2018-06-14
Known Types Based On Descriptors

### LINKED ISSUE #124: Resolve remaining direct dependency constants
See `ext.gradle` for remaining dependency variables. 

The goal is to use dependencies from `config/gradle/dependencies.gradle` instead.

dmdashenkov:
# Known Types

Before, we used a number of special compile-time tools to process the message types. This approach showed to be not agile enough.
Since now, we use the `FileDescriptorSet` object stored in a file/files in the classpath in order to discover known types at runtime.

## Implementation details

It is expected that all the known types are mentioned in at least one file `known_types.desc` found in the runtime classpath.

By default, the Protobuf compiler assembles such a file when processing `.proto` sources and then the assembled file is added to the module compilation output by the Spine model compiler automatically. When necessary, the user may as well add such files manually. Duplicate types are ignored when loading the types.

On startup, the `known_types.desc` files are read from the classpath and the descriptor linking is performed. Note that there can be many files with such name in the classpath; in this case all of them are parsed. However, for better startup performance it is recommended to exclude all the `known_types.desc` files except for the "last" one (the one built in the leaf of the project hierarchy tree).

# Build Changes

Since the `ProtoToJavaMapper` compile-time tool is no longer present, the `base` module is now built without the help from `model-compiler`.

Instead, there is some custom build logic in the `base/build.gradle`, which substitutes the configurations performed by the model compiler.

Also, since `base` is unable to build its own `VBuilder`s, Spine `core` module will be configured to build validating builders from the classpath.

## Smoke Tests

In order to destroy the "two-step" build completely, the smoke tests for the Gradle plugins are now collected under an independent Gradle project hierarchy. 

The __main__ hierarchy includes `base`, `model-compiler` with its helper modules, and `testlib`. All the test modules are moved to this new __smoke-tests__ hierarchy.

__smoke-tests__ depend on __main__ via [composite builds](https://docs.gradle.org/current/userguide/composite_builds.html#included_build_declaring_substitutions). It may require some [additional settings](https://blog.jetbrains.com/idea/2016/10/intellij-idea-2016-3-eap-gradle-composite-builds-and-android-studio-2-2/) to make the IDEA treat the __smoke-tests__ sources as actual source code.

# Instructions For The Framework Users

It is now not recommended to configure `protobuf` Gradle plugin manually unless custom requirements from the user side apply. The model compiler now performs all the _required_ configurations automatically.

Though the Protobuf Gradle plugin configuration is generally safe and OK, it is not a good idea to change the settings concerning the descriptor file generation. In the worst case, it may lead to a not obvious runtime error.

# Other Changes

## `UnknownOptions`

The `UnknownOptions` class is deleted. The correct way to parse any kind of a Protobuf message from binary is `MyMessage.parse(binary, Options.registry())`. This way, all the Spine custom options are resolved and can be read well. For reading the options, see `Options` utility.

## `validators-gen`

The `validators-gen` smoke test module is deleted, since it had no actual tests, only some `.proto` definitions.

## Misc

 - Deprecated `compile` Gradle dependency configuration is substitured with `api`/`implementation`.
 - Since the Protobuf version is updated to `3.6`, the script removing the `@Deprecated` annotation from the generated code is deleted.
 - Custom `ext.gradle` is reduced (fixes #124).


dmdashenkov:
@armiol, PTAL at the PR general idea.

There are still issues to address in this PR including the code and doc issues, coverage and (last, but not least) descriptor file merging technics.

ext.gradle
```diff
@@ -25,7 +25,7 @@
  * as we want to manage the versions in a single source.
  */
 
-final def SPINE_VERSION = '0.10.41-SNAPSHOT'
+final def SPINE_VERSION = '0.10.42.1-SNAPSHOT'
```
alexander-yevsyukov: Why not `43`?

dmdashenkov: This is still WIP. In order to avoid conflicts with of my locally published versions with the globally accessible ones, I use non-typical versioning.

alexander-yevsyukov:
> known_types.desc

Can we keep using the names `main.desc` and `test.desc`?

alexander-yevsyukov:
> ... is not built without the help from model-compiler.

Is it `now` instead of `not`? 

If not, please avoid the double negation in the wording. It's hard to understand.

dmdashenkov:
@armiol, PTAL.

alexander-yevsyukov:
@dmdashenkov, it think it would be appropriate if you add your authorship to files which you changed much.

base/src/main/java/io/spine/type/TypeUrl.java
```diff
@@ -193,12 +193,28 @@ private static String prefixFor(GenericDescriptor descriptor) {
...
+    /**
+     * Returns a message {@link Class} corresponding to the Protobuf message type represented
+     * by this type URL.
+     *
+     * <p>This is a convenience method. Use it only when sure that the {@link TypeUrl} represents
+     * a message (i.e. not an enum).
+     *
+     * @throws IllegalStateException if the type URL represents an enum
+     */
+    public <T extends Message> Class<T> getMessageClass() throws UnknownTypeException {
```
armiol: This is an exact duplicate of the same method in `TypeName`. And the one in `TypeName` is not used in `base`. Please review the latter. Maybe it should be removed.

base/src/main/java/io/spine/validate/rules/ValidationRules.java
```diff
@@ -38,6 +39,8 @@
  */
 public class ValidationRules {
 
+    private static final Splitter OPTION_SPLITTER = Splitter.on(',');
```
armiol: Please have a description.

base/src/main/java/io/spine/option/Options.java
```diff
@@ -0,0 +1,222 @@
...
+
+import java.util.Optional;
+
+/**
+ * A utility class for working with custom Protobuf options.
+ *
+ * @author Dmytro Dashenkov
+ */
+@Internal
+public final class Options {
```
alexander-yevsyukov: @dmdashenkov, please file an issue to refactor this class so that particular kinds of options are separate classes.

tools/model-compiler/build.gradle
```diff
@@ -31,42 +31,40 @@ ext {
...
 
     implementation deps.build.protobuf
     implementation deps.gen.javaPoet
-    
+
     // A library for parsing Java sources.
     // Used for parsing Java sources generated from Protobuf files
     // to make their annotation more convenient.
-    compile (group: 'org.jboss.forge.roaster', name: 'roaster-api', version: roasterVersion) {
+    implementation (group: 'org.jboss.forge.roaster', name: 'roaster-api', version: roasterVersion) {
```
armiol: As long as you do this, you may do the `compile` -> `implementation` / `api` migration in the rest of modules.

base/src/main/java/io/spine/option/Options.java
alexander-yevsyukov: The problem with the current code is that it's too much `option` in the text. You have `Options.option(..., option).

tools/model-compiler/src/main/java/io/spine/tools/compiler/enrichment/TypeNameParser.java
```diff
@@ -0,0 +1,99 @@
...
+final class TypeNameParser {
+
+    /**
+     * A splitter for message types.
+     *
+     * <p>Splits strings by the {@code ,} (comma) character.
+     *
+     * @see #joiner
+     */
+    static final Splitter splitter = Splitter.on(',');
```
armiol: This field can be `private`.

tools/model-compiler/src/main/java/io/spine/tools/gradle/compiler/DescriptorSetMergerPlugin.java
```diff
@@ -0,0 +1,110 @@
...
+        createMainTask(project);
+        createTestTask(project);
+    }
+
+    private void createMainTask(Project project) {
+        logDependingTask(MERGE_DESCRIPTOR_SET,
+                         GENERATE_VALIDATING_BUILDERS,
+                         GENERATE_PROTO);
+        newTask(MERGE_DESCRIPTOR_SET,
+                createMergingAction(configuration(project, "runtime"),
```
armiol: I'd have `Configurations` enum for `runtime` and `testRuntime` values. To reuse in future.

tools/protoc-plugin/src/main/java/io/spine/tools/protoc/Plugin.java
```diff
@@ -59,9 +59,10 @@ public static void main(String[] args) {
     }
 
     private static CodeGeneratorRequest readRequest() {
-        final CodedInputStream stream = CodedInputStream.newInstance(System.in);
         try {
-            final CodeGeneratorRequest request = CodeGeneratorRequest.parseFrom(stream);
+            CodeGeneratorRequest request = CodeGeneratorRequest.parseFrom(
```
armiol: Please move the content after `=` to the next line.

tools/smoke-tests/external-known-types/build.gradle
```diff
@@ -0,0 +1,20 @@
...
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
```
armiol: Please have a comment saying this file is left blank on purpose.

alexander-yevsyukov:
@dmdashenkov,
 
> Also, since `base` is unable to build its own VBuilders, Spine `core` module will be configured 
> to build validating builders from the classpath.

Aren't we going to have a split package issue in Java 9 then?

tools/smoke-tests/settings.gradle
```diff
@@ -18,27 +18,23 @@
...
-    generatedFilesBaseDir = generatedRootDir
-    protoc {
-        artifact = deps.build.protoc
-    }
-
-    generateProtoTasks {
-        all().each { final task ->
-            task.descriptorSetOptions.includeSourceInfo = true
+// :base root project
+includeBuild("$rootDir/../../") {
```
armiol: Please describe the trick in details, so that an out-of-context person could understand and re-use for own purposes or education.

tools/smoke-tests/validation-rules/build.gradle
```diff
@@ -0,0 +1,19 @@
...
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
```
armiol: Please have it commented due to the same reasons as above.

alexander-yevsyukov:
In the rest, it's a big and good piece of work, and I'm happy it's hearing the completion.

ext.gradle
```diff
@@ -29,10 +29,10 @@ final def SPINE_VERSION = '0.10.43-SNAPSHOT'
 
 ext {
     spineVersion = SPINE_VERSION
+    versionToPublish = SPINE_VERSION
 
     protobufVersion = '3.5.1'
```
armiol: Please try to use the [dependencies.gradle](https://github.com/SpineEventEngine/config/blob/409b966f2784f12b6a94fda95570018b879e5524/gradle/dependencies.gradle) instead.

There is a related issue, see #124. The problem is that the dependencies in this file are referenced from a bunch of integration-test projects. That's why this version still remains.

Please see if resolving #124 in this PR is easy for you. If not, let's merge this PR and open a new one.

tools/plugin-base/src/main/java/io/spine/tools/gradle/ConfigurationName.java
```diff
@@ -0,0 +1,64 @@
...
+ * The names of Gradle configurations used by the Spine model compiler plugin.
+ *
+ * @author Dmytro Dashenkov
+ */
+public enum ConfigurationName {
+
+    /**
+     * The {@code runtime} configuration.
+     *
+     * <p>Contains the runtime classpath of the {@code main} scope of the project.
```
armiol: I'd have a reference to the Gradle documentation in the class-level Javadoc instead explaining it here. 

Please remove the second paragraph here and in the definition below and update the class-level documentation.

tools/protoc-plugin/build.gradle
```diff
@@ -42,7 +42,8 @@ jar {
     }
     // Assemble "Fat-JAR" artifact containing all the dependencies.
     from {
-        configurations.compile.collect {
+        configurations.runtimeClasspath.collect {
+            println it
```
armiol: Not sure why we need such a poor console output, it seems to be a debugging artifact. 

Either append it with more details to make it helpful, or remove.

tools/smoke-tests/build.gradle
```diff
@@ -68,9 +68,12 @@ subprojects {
     apply plugin: 'idea'
 
     dependencies {
+        // TODO:2018-06-23:dmytro.dashenkov: Migrate to the `implementation` config when resolved:
+        // todo https://github.com/google/protobuf-gradle-plugin/issues/242
+        // todo Spine issue:
```
armiol: Please create an issue.

tools/javadoc-prettifier/build.gradle
```diff
@@ -24,6 +24,12 @@ apply plugin: 'java-gradle-plugin'
 dependencies {
     implementation deps.build.guava
 
-    implementation project(':plugin-base')
     implementation project(':base')
+    implementation project(':plugin-base')
+
+    testImplementation project(':plugin-base').sourceSets.test.output // Test classes of `plugin-base`
```
armiol: Let's try to move the required classes to some test utility module instead of reusing the compile output of the test sources.

PR #129, opened 2018-06-27
Assemble `base` Validating Builders

dmdashenkov:
# Validating Builders

Since #125, `base` is not built with the model compiler. Thus, the validating builders for the `base` types are built separately.

In this PR we introduce a separate module, which builds the validating builders for `base`. After this, the builders are added to the published archive of `base`.

# Also In This Pull Request

 - The log about missing descriptor set is downgraded from `warn` to `debug`, since the descriptor set generation is configured automatically.
 - Validating builders do not use deprecated `get<MapField>()` methods anymore (but use `get<MapField>Map()` methods).


dmdashenkov:
@armiol, PTAL.

dmdashenkov:
@armiol, PTAL again.

dmdashenkov:
@armiol, PTAL again.

PR #131, opened 2018-07-03
Change `GradleProject` to accept `File` as project directory

dmitrykuzmin:
Previously, the `GradleProject` test tool accepted only JUnit 4 [`TemporaryFolder`](https://junit.org/junit4/javadoc/4.12/org/junit/rules/TemporaryFolder.html) as the project directory. This caused problems for the tests which were already migrated to the JUnit 5 API.

Now, the interface of the `GradleProject` is changed to accept more generic `File` as the project directory, thus enabling it to be used in both JUnit 4 and JUnit 5 tests.

The Spine version advances to `0.10.45-SNAPSHOT`.

### DIFF
--- a/ext.gradle
+++ b/ext.gradle
@@ -25,7 +25,7 @@
  * as we want to manage the versions in a single source.
  */
 
-final def SPINE_VERSION = '0.10.44-SNAPSHOT'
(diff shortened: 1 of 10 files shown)


dmitrykuzmin:
@armiol PTAL.

PR #132, opened 2018-07-10
Add `TempDirectory` extension to the `testlib`

dmitrykuzmin:
This PR adds a `TempDirectory` extension to the `spine-testlib` module.

Previously, we imported the file from the [JUnit Pioneer project](https://github.com/junit-pioneer/junit-pioneer).

Now, we will use the same file copied to the `spine-testlib` to know for sure that its API won't change without our intention.

When the JUnit Pioneer project reaches its release, we should switch back to using dependency instead of the copy-pasted file.

Spine version advances to `0.10.46-SNAPSHOT`.


dmitrykuzmin:
@armiol PTAL. My intention was to take the file "as-is", or should I modify and test JUnit code to bring the PR to the Spine standards?

testlib/src/main/java/io/spine/test/TempDirectory.java
```diff
@@ -0,0 +1,300 @@
...
+                        throw new IOException(
+                                "Failed to delete temp directory " + dir.toAbsolutePath() + " at: " + path.toAbsolutePath(),
+                                ex);
+                    }
+                    return CONTINUE;
+                }
+            });
+        }
+    }
+}
```
armiol: The EOF on this file is broken. See the GitHub complaints.

There is an `core.autocrlf` setting for Git clients to avoid this issue. Please have a read of the Git docs on this.

dmitrykuzmin:
@armiol PTAL.

PR #138, opened 2018-07-16
Add `@CanIgnoreReturnValue` to generated Validating Builders

dmitrykuzmin:
Previously, the generated `ValidatingBuilder` descendants fell under the `@CheckReturnValue` rule and the return values of their methods could not be safely ignored. This caused inconveniences with their usage.

Now, `ValidatingBuilder` setter methods of various forms will be annotated with `@CanIgnoreReturnValue`.

List of methods which are affected by `@CanIgnoreReturnValue`:
1. `set...`
2. `setRaw...`
3. `clear...`
4. `addAll...`
5. `remove...`
6. `addRaw...`
7. `addAllRaw`
8. `put...`
9. `putAll...`
10. `putRaw...`
11. `putAllRaw...`

Basically, all methods have `@CanIgnoreReturnValue` except static `newBuilder()` and getters.

Spine version advances to `0.10.48-SNAPSHOT`.


dmitrykuzmin:
@armiol PTAL.

tools/model-compiler/src/main/java/io/spine/tools/compiler/validation/ValidatingBuilderWriter.java
```diff
@@ -87,6 +88,7 @@ void write(VBType type) {
                                    messageBuilderClassName,
                                    methodsAssembler.createMethods())
                         .addAnnotation(generatedBySpineModelCompiler())
+                        .addAnnotation(canIgnoreReturnValue())
```
armiol: There is also a static `newBuilder` method, which return value must NOT be ignored.

Therefore your approach is way too broad. Let's discuss in person if needed.

base/src/main/java/io/spine/code/java/Annotations.java
```diff
@@ -51,10 +56,20 @@ private Annotations() {
...
-     * @return an {@link AnnotationSpec} describing the {@link Generated javax.annotation.Generated}
+     * @return an {@code AnnotationSpec} describing the {@link Generated javax.annotation.Generated}
      *         annotation
      */
     public static AnnotationSpec generatedBySpineModelCompiler() {
         return GENERATED;
     }
+
+    /**
+     * Generates {@code \@CanIgnoreReturnValue} annotation spec.
```
armiol: I know another  approach is to use `{@literal @}`
```
Generates {@code {@literal @}CanIgnoreReturnValue} annotation spec.
```
Please investigate which one is better.

dmitrykuzmin: It seems that just `{@code @CanIgnoreReturnValue}` is displayed okay in both Idea doc reference and the doc HTML generated by the Gradle.

When using any of the approaches for escaping the literal though, the doc is not displayed properly, so I just removed the escape mechanism here and in the similar case.

PR #145, opened 2018-07-16
Avoid mutable map

alexander-yevsyukov:
This PR migrates from deprecated Protobuf map mutation API.


alexander-yevsyukov:
@armiol, PTAL.

PR #146, opened 2018-07-17
Remove the unused `final` keyword on local variables and repackaging for the `testlib` module.

illia-shepilov:
This PR provides the following changes for the `testlib` module:

- The `final` keyword on local variables was removed.
- The package was renamed from`io.spine.test` to `io.spine.testing`.


illia-shepilov:
@armiol PTAL.

testlib/src/main/java/io/spine/testing/TempDirectory.java
```diff
@@ -1,14 +1,24 @@
 /*
- * Copyright 2015-2018 the original author or authors.
+ * Copyright 2018, TeamDev. All rights reserved.
```
armiol: Please revert the changes to this header.

PR #149, opened 2018-07-17
Replace Guava's `Predicate` and `Function` with JDK8 `Predicate` and `Function` accordingly

illia-shepilov:
This PR provides the following changes for the `base` module:

- Guava's `Predicate` was replaced with JDK8 `Predicate`;
- Guava's `Function` was replaced with JDK8 `Function`.


base/src/main/java/io/spine/validate/ValidationException.java
```diff
@@ -66,7 +66,10 @@ public String toString() {
 
         String violationContent = constraintViolations.isEmpty()
```
armiol: Please move the expression to the next line and then fix the formatting, so that everything fits. Maybe use a static import. `Collectors.toList()` should be properly aligned.

PR #155, opened 2018-07-17
Remove deprecated proto option

alexander-yevsyukov:
This PR removes the `java_generate_equals_and_hash` from proto files. The option does not have effect in Protobuf v3 and was deprecated.

The version advances to `0.10.50-SNAPSHOT`.


alexander-yevsyukov:
@dmdashenkov, PTAL.

alexander-yevsyukov:
@armiol, PTAL, if Dima is away.

PR #157, opened 2018-07-20
Don't apply validation to outer message fields that are not set

dmitrykuzmin:
This PR fixes issue https://github.com/SpineEventEngine/core-java/issues/646.

Previously, we applied validation to all fields marked with `(valid) = true` even if they were not set.

Now, we don't apply validation to such fields' inner values, so if we have message `Outer` which contains field `A` which contains required field `B`, and the `A` is not set in the `Outer` builder, the error is not thrown.

It is worth noting, that the validation is still applied to the field itself, so in the previous example if the `A` is `required` and not set, the error will be thrown.

It is also worth noting that `Message` fields set to default values are considered same as unset in Protobuf, so if we have `OuterVBuilder.setA(A.getDefaultInstance()).build()`, the validation is still not applied to `A`'s fields.

The Spine version advances to `0.10.54-SNAPSHOT`.


dmitrykuzmin:
@armiol PTAL.

PR #159, opened 2018-07-20
Improve tests

alexander-yevsyukov:
This PR:
  * Suppresses a PMD warning in `UtilityClassTest`.
   The suppression is needed because PMD does not recognize 
   `NullPointerTester.testAllPublicStaticMethods()` call as a valid assertion method.
  * Makes the logging supplier memoizing.
  * Migrates tests for `Logging` and `Tests`  to JUnit5.

Version advances to `0.10.55-SNAPSHOT`.


alexander-yevsyukov:
@illia-shepilov, PTAL.

PR #162, opened 2018-07-23
Remove `Supplier` conversion

alexander-yevsyukov:
This PR removes `Suppliers2` utility class that is not necessary, the Guava -> Java conversion can be provided via using method references. See [Guava `Supplier` Javadoc](https://google.github.io/guava/releases/21.0/api/docs/com/google/common/base/Supplier.html)for details.

Also the PR migrates `GenericTypeIndex` interface to use default interface instead of a nested static class.


alexander-yevsyukov:
@armiol, PTAL.

base/src/main/java/io/spine/reflect/Types.java
```diff
@@ -74,4 +76,32 @@ private Types() {
                                                  .getType();
```
armiol: Something wrong with the formatting here and two lines above. Please adjust, as long as you are modifying the codebase right next to this spot.

PR #167, opened 2018-07-30
Tests and JavaDoc for the `assertMatchesMask` method

illia-shepilov:
This PR provides the following changes:

1. Added the missed JavaDoc for the `assertMatchesMask` utility method;
2. Created tests for the `assertMatchesMask` utility method.


testlib/src/main/java/io/spine/testing/Tests.java
```diff
@@ -155,6 +155,9 @@ private static void callConstructor(Constructor constructor) {
     /**
      * Asserts that the passed message has a field that matches the passed field mask.
      *
+     * @param message the message to assert
+     * @param fieldMask the field mask which will be matched with a field of the message
```
armiol: `... which is matched agains the message field`

testlib/src/test/java/io/spine/testing/TestsShould.java
```diff
@@ -136,6 +139,65 @@ void whenFalse() {
...
+                                  .getFullName();
+            FieldMask.Builder builder = FieldMask.newBuilder();
+            builder.addPaths(fieldPath);
+            FieldMask fieldMask = builder.build();
+
+            assertThrows(AssertionError.class, () -> assertMatchesMask(timestampMsg, fieldMask));
+        }
+
+        @Test
+        @DisplayName("throws the exception when field is not set")
```
armiol: `AssertionError` is not an exception. Same above.

Also, I think `...when the field value is not set`.

PR #172, opened 2018-07-31
Type improvements

alexander-yevsyukov:
This PR:
  * Introduces abstract base for value holding objects (`ValueHolder`) and makes
    `StringTypeValue` and `ClassTypeValue` descent from this class.
  * Value holder classes are now `Serializable`.
  * Marks `StringTypeValue` and `ClassTypeValue` `@Immutable`.
  * Unifies predicates for filtering proto message files.


alexander-yevsyukov:
@armiol, PTAL.

alexander-yevsyukov:
@armiol, now PTAL.

PR #174, opened 2018-08-01
Improvements to type-related classes

alexander-yevsyukov:
This PR introduces additional improvements to working with type-related classes. In particular:
  * `KnownTypes` transformed into `enum`.
  * `Type`, `TypeSet` were made `@Immutable`

### DIFF
--- a/base/src/main/java/io/spine/type/KnownTypes.java
+++ b/base/src/main/java/io/spine/type/KnownTypes.java
@@ -51,7 +51,10 @@ public enum KnownTypes {
 
     INSTANCE;
 
-    @SuppressWarnings("TransientFieldNotInitialized") // We substitute the instance in readResolve()
+    @SuppressWarnings({
+            "TransientFieldNotInitialized", // We substitute the instance in readResolve()
+            "ImmutableEnumChecker" // The field is effectively immutable.
+    })
     private final transient Supplier<TypeSet> typeSet = memoize(KnownTypes::load);
 
     /**
--- a/base/src/main/java/io/spine/type/KnownTypes.java
+++ b/base/src/main/java/io/spine/type/KnownTypes.java
@@ -52,7 +52,7 @@ public enum KnownTypes {
     INSTANCE;
 
     @SuppressWarnings({
-            "TransientFieldNotInitialized", // We substitute the instance in readResolve()
+            "TransientFieldNotInitialized", // We substitute the instance in readResolve().
             "ImmutableEnumChecker" // The field is effectively immutable.
     })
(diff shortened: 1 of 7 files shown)


alexander-yevsyukov:
@armiol, PTAL.

alexander-yevsyukov:
@armiol, PTAL.

base/src/main/java/io/spine/type/KnownTypes.java
```diff
@@ -45,15 +47,34 @@
...
  */
 @Internal
-public class KnownTypes {
+public enum KnownTypes {
 
-    private final TypeSet types;
+    INSTANCE;
+
+    @SuppressWarnings({
+            "TransientFieldNotInitialized", // We substitute the instance in readResolve()
```
armiol: As long as you have period at the end of the following sentence, let's have a period here as well.

testlib/src/test/java/io/spine/testing/VerifyShould.java
```diff
@@ -1236,18 +1181,16 @@ public void run() {
...
+            throw runtimeException;
         };
 
         assertThrowsWithCause(runtimeException.getClass(), cause.getClass(),
                                      throwsRuntimeException);
     }
 
-    @SuppressWarnings("EqualsAndHashcode")
+    @SuppressWarnings({"EqualsAndHashcode", "EqualsHashCode"})
+        // Not having hashCode() is a part of this test. environment class.
```
armiol: `test. environment class.` seem to have an extra period.

testlib/src/test/java/io/spine/testing/VerifyShould.java
```diff
@@ -1236,18 +1181,16 @@ public void run() {
...
+        Runnable throwsRuntimeException = () -> {
+            throw runtimeException;
         };
 
         assertThrowsWithCause(runtimeException.getClass(), cause.getClass(),
                                      throwsRuntimeException);
     }
 
-    @SuppressWarnings("EqualsAndHashcode")
+    @SuppressWarnings({"EqualsAndHashcode", "EqualsHashCode"})
```
armiol: I have looked at the issue, and it seems that there is no good way to eliminate this duplication. For some reason JetBrains and ErrorProne have different notion, and it seems non-customizable: in ErrorProne the warning name is [a part of their source code](https://github.com/google/error-prone/blob/master/core/src/main/java/com/google/errorprone/bugpatterns/EqualsHashCode.java#L60), and in IDEA I have found no option to override warning names as well.

So the only way is to have `hashCode()` :)

PR #175, opened 2018-08-02
Inherit `@Internal`

alexander-yevsyukov:
This PR makes `@Internal` annotation `@Inherited`.

### DIFF
--- a/base/src/main/java/io/spine/annotation/Internal.java
+++ b/base/src/main/java/io/spine/annotation/Internal.java
@@ -22,6 +22,7 @@ package io.spine.annotation;
 
 import java.lang.annotation.Documented;
 import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
 import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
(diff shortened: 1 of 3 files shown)


alexander-yevsyukov:
@dmdashenkov, PTAL.

alexander-yevsyukov:
@dmdashenkov, PTAL.

dmdashenkov:
@alexander-yevsyukov 
There is a counterexample. `io.spine.server.delivery.Delivery` is marked `@Internal`, but `io.spine.server.aggregate.AggregateDelivery` is marked as `@SPI`.

I believe, there _may_ be cases when an internal API is implemented in a public API class.

alexander-yevsyukov:
Think of it as if `@Internal` makes a class invisible. We cannot make its descendant visible without making the parent class visible.

dmdashenkov:
@alexander-yevsyukov 
Actually, we can. This way, Java allows making subtypes of a package-local class `public`.

alexander-yevsyukov:
It's not the same thing. Making a class `public` is a deliberate action. Here we want to avoid exposing classes or methods that are already `public` to the end user API by mistake.

PR #180, opened 2018-08-08
Add Error Prone check to advice `...VBuilder` usage

dmitrykuzmin:
This PR resolves issue https://github.com/SpineEventEngine/core-java/issues/769.

The new `spine-errorprone-checks` module contains custom Error Prone check which detects usages of `Message.Builder` and advices to use `...VBuilder` instead.

The `ErrorProneChecksPlugin`, which is a Model Compiler component, configures a project to use the `spine-errorprone-checks` as the `annotationProcessor` for all `JavaCompile` tasks.

Thus, the check is automatically applied to all projects using the `spine-model-compiler`.

Its default severity is `WARN`, but it may be configured per-Gradle module as follows:

1) For all checks:
```
modelCompiler {
    spineCheckSeverity = "OFF"       // or "ERROR", or "WARN"
}
```
2) For the specific checks (though for now there is only one):
```
spineErrorProneChecks {
    useValidatingBuilder = "OFF"       // or "ERROR", or "WARN"
}
```

The latter overrides the former.

Spine version advances to `0.10.64-SNAPSHOT`.


settings.gradle
```diff
@@ -32,6 +32,7 @@ final def module = { final String name, final String path ->
     project(":$name").projectDir = new File("$rootDir/$path")
 }
 
+module 'checker',             'tools/checker'
```
armiol: As long as we also have `model-verifier` (and thus "checker" and "verifier" may be confused with each other), this probably should be `errorprone-checks`.

tools/model-compiler/src/main/java/io/spine/tools/gradle/compiler/Extension.java
```diff
@@ -305,6 +314,13 @@ public void setIndent(int indent) {
         return ImmutableList.copyOf(dirsToClean);
     }
 
+    public static Severity getSpineCheckSeverity(Project project) {
+        Severity result = spineProtobuf(project).spineCheckSeverity;
+        log().debug("The severity of Spine checks is {}",
```
armiol: `The severity of Spine-custom Error Prone checks is {}`

tools/model-compiler/src/main/java/io/spine/tools/gradle/compiler/SpineCheckerPlugin.java
```diff
@@ -0,0 +1,350 @@
...
+
+    /**
+     * Checks if the project has Error Prone plugin applied.
+     */
+    @VisibleForTesting
+    protected boolean hasErrorPronePlugin(Project project) {
+        PluginContainer appliedPlugins = project.getPlugins();
+        boolean hasPlugin = appliedPlugins.hasPlugin(ERROR_PRONE_PLUGIN_ID);
+        return hasPlugin;
+    }
```
armiol: Can we please try to keep the plugin class as small as possible by moving the code into other helper-filter-etc classes. See how it's done for `ModelCompilerPlugin`.

tools/model-compiler/src/main/java/io/spine/tools/compiler/check/ProjectUtils.java
```diff
@@ -0,0 +1,64 @@
...
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * Utilities related to the classes working with Spine-custom Error Prone checks and the
+ * {@link io.spine.tools.gradle.compiler.ErrorProneChecksPlugin}.
+ *
+ * @author Dmytro Kuzmin
+ */
+class ProjectUtils {
```
armiol: Please think of another name. I can see that now it looks like `ProjectArguments`.

Whichever new name you pick, please change the class description accordingly.

tools/model-compiler/src/test/java/io/spine/tools/compiler/check/given/ProjectTaskUtil.java
```diff
@@ -0,0 +1,96 @@
...
+import static com.google.common.collect.testing.Helpers.assertEmpty;
+import static io.spine.testing.Verify.assertContains;
+
+/**
+ * The utility with several methods related to project tasks and required to properly test the
+ * {@link io.spine.tools.gradle.compiler.ErrorProneChecksPlugin} functionality.
+ *
+ * @author Dmytro Kuzmin
+ */
+public class ProjectTaskUtil {
```
armiol: As long as this is a test helper, let's call it `ProjectConfigurations`.

dmitrykuzmin:
@armiol PTAL.

PR #181, opened 2018-08-10
Adopt Google Truth

alexander-yevsyukov:
This PR adds Google Truth into dependencies, and migrates `TimeTest` to using this assertion framework.


alexander-yevsyukov:
@armiol, PTAL.

PR #182, opened 2018-08-11
Better logging

alexander-yevsyukov:
This PR transforms `Logging` into an interface with shortcut methods for logging.

It also adds `Truth`-based testing utilities for testing logging output.

### Deprecated API
 * `Logging.supplyFor(Class)` — use `Logger.get(Class)` for static context.


alexander-yevsyukov:
@armiol, PTAL.

I've suppressed Codacy on the package level for `logging` under `main`. There may be a more narrow approach, if you know it, please let me know. Even though, Codacy does shown an error here in GitHub, it does not display it in its own UI, which is strange.

base/src/main/java/io/spine/logging/Util.java
```diff
@@ -0,0 +1,49 @@
...
+
+import java.util.function.BiConsumer;
+
+/**
+ * Utilities internal to this package that are used by {@link Logging} interface default methods
+ * and should not be exposed in the public API.
+ *
+ * @author Alexander Yevsyukov
+ */
+class Util {
```
armiol: Such a naming may encourage the team to have package-private `Util` classes it their packages. Can we please rename it to something like `LogMessages` (more generic) or `LogThrowables` (more specific)?

base/src/test/java/io/spine/logging/QuadriConsumer.java
```diff
@@ -0,0 +1,37 @@
...
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.spine.logging;
+
+/**
+ * Describes a method with three parameters.
```
armiol: I think, it's `with four parameters`.

alexander-yevsyukov:
@armiol, fixed. PTAL.

PR #187, opened 2018-09-03
Remove redundant logging

dmdashenkov:
In this PR we remove overly large and redundant logging of `.desc` files lookup.

This problem is a blocker since Travis cannot build `core-java` with `--debug` option in Gradle—the log size is over 4 MB.


dmdashenkov:
@armiol, PTAL.

PR #189, opened 2018-09-13
Built-In Marker Interfaces

dmdashenkov:
Since now, the generated events, commands, and rejections are marked with `EventMessage`, `CommandMessage`, and `RejectionMessage` marker interfaces respectively.


dmdashenkov:
@armiol, PTAL.

base/src/test/proto/spine/test/code/inherit_all_source_file_test.proto
```diff
@@ -0,0 +1,30 @@
...
+
+syntax = "proto3";
+
+package spine.test.code;
+
+import "spine/options.proto";
+
+option (type_url_prefix) = "type.spine.io";
+
+message InheritAllMessage {
```
armiol: Can we please document its purpose? The same applies to the rest of the test messages in this test case. It's going to make it easier to support in future.

tools/protoc-plugin/src/main/java/io/spine/tools/protoc/CompilerOutput.java
```diff
@@ -0,0 +1,38 @@
...
+ * @author Dmytro Dashenkov
+ */
+interface CompilerOutput {
+
+    /**
+     * Obtains the {@link File CodeGeneratorResponse.File} representing this output item.
+     *
+     * @return compiler output as a {@link File}
+     */
+    File toFile();
```
armiol: `asFile()` or `getDestination()`. Because `toFile` seems an action, which is most likely not.

tools/protoc-plugin/src/main/java/io/spine/tools/protoc/InsertionPoint.java
```diff
@@ -0,0 +1,79 @@
...
+import static io.spine.code.java.PackageName.delimiter;
+import static java.lang.String.format;
+
+/**
+ * A {@link io.spine.tools.protoc.CompilerOutput CompilerOutput} item, which alters a generated
+ * message class to implement a given marker interface.
+ *
+ * @author Dmytro Dashenkov
+ */
+final class InsertionPoint extends AbstractCompilerOutput {
```
armiol: Isn't it `ExtensionPoint` as per Protobuf compiler dictionary?

dmdashenkov: It's insertion point as in `CodeGeneratorResponse.File.insertion_point`.

Issue #80, opened 2018-02-09
Simplify the `URL` type


alexander-yevsyukov:
We can just do fine with the `value` or `raw` field.

armiol:
Implemented.

PR #191, opened 2018-09-18
Clean up types

alexander-yevsyukov:
This PR:
  * Removes support of structured fields in the `Url` type (#80).
  * Adds stringifiers for `Url`, `InternetDomains`, and `EmailAddress`.
  * Adds `SerializableFunction` interface.
  * Adds `SerializableStringifier` and `FnStringifier` classes.
  * Adds utility class called `Registrar` for registering several serializers.


alexander-yevsyukov:
@armiol, PTAL.

alexander-yevsyukov:
@dmdashenkov, PTAL.

base/src/main/java/io/spine/net/EmailAddresses.java
```diff
@@ -44,10 +47,23 @@ public static Pattern pattern() {
...
     }
 
-    private static void checkArgumentIsEmailAddress(CharSequence value) {
-        checkNotNull(value);
-        checkArgument(pattern().matcher(value)
-                               .matches());
+    /**
+     * Verifies if the passed sequence is a valid email address.
+     */
+    public static boolean isValid(CharSequence value) {
```
dmdashenkov: Let's use `String` instead.

base/src/main/java/io/spine/string/FnStringifier.java
```diff
@@ -0,0 +1,59 @@
...
+ */
+
+package io.spine.string;
+
+import io.spine.util.SerializableFunction;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * Abstract base for stringifiers that convert values using method references.
```
dmdashenkov: `using function objects.`, I think. A method reference is a concept from Java syntax, however `FnStringifier` feels much more than that.

base/src/main/java/io/spine/string/SerializableStringifier.java
```diff
@@ -0,0 +1,55 @@
...
+
+import java.io.Serializable;
+
+/**
+ * Abstract base serializable stringifiers.
+ *
+ * @param <T> the type to stringify
+ * @author Alexander Yevsyukov
+ */
+public abstract class SerializableStringifier<T> extends Stringifier<T> implements Serializable {
```
dmdashenkov: Why can't all `Stringifier`s be serializable?

base/src/main/java/io/spine/string/SerializableStringifier.java
```diff
@@ -0,0 +1,55 @@
...
+
+    /**
+     * Returns the identity of the stringifier.
+     */
+    @Override
+    public final String toString() {
+        return identity;
+    }
+}
+
```
dmdashenkov: Please remove this line.

base/src/main/proto/spine/net/url.proto
```diff
@@ -9,87 +9,99 @@ option java_multiple_files = true;
...
+    //
+    // Validity of the value should be provided by the code creating instances of this message.
     //
-    // This type represents URL according to RFC 3986.
-    // Using this type makes it easier to operate concrete URL attributes.
-    // Please see https://tools.ietf.org/html/rfc3986.
-    message Record {
-        // Application layer protocol
-        Protocol protocol = 1;
+    string spec = 3 [(required) = true];
```
dmdashenkov: Why `spec` and not `value`?

base/src/main/proto/spine/net/url.proto
```diff
@@ -9,87 +9,99 @@ option java_multiple_files = true;
...
+// A URL in a structured form.
+//
+// This type represents URL according to RFC 3986.
+// Using this type makes it easier to operate concrete URL attributes.
+// Please see https://tools.ietf.org/html/rfc3986.
+//
+// This type should be used only by parsing or validating utilities.
+// For general purpose usage, please use `Url`.
+//
+message UrlRecord {
```
dmdashenkov: Maybe, `Uri`?

ext.gradle
```diff
@@ -25,7 +25,7 @@
  * as we want to manage the versions in a single source.
  */
 
-final def SPINE_VERSION = '0.10.70-SNAPSHOT'
+final def SPINE_VERSION = '0.10.72-SNAPSHOT'
```
dmdashenkov: Why by 2 at once?

base/src/main/java/io/spine/string/SerializableStringifier.java
alexander-yevsyukov: Because some of them may use non-serializable guts.

base/src/main/proto/spine/net/url.proto
alexander-yevsyukov: `value` used to be a `oneof` switch. I'll add it to reserved too.

PR #193, opened 2018-09-19
Remove Redundant Code Generation

dmdashenkov:
In this PR we remove the redundant generated classes. 

Before, all the `google.protobuf` and `spine.base` definitions were converted into Java in some modules tests. This might be leftovers from some kind of a debug process. Now, this code generation is not done anymore.

Also, since now all the `generated` directories are deleted on `:clean`.


dmdashenkov:
@alexander-yevsyukov, PTAL.

PR #196, opened 2018-09-21
Integrate Marker Interfaces

dmdashenkov:
In this PR we integrate standard marked interfaces into the framework API. The larger version of this change is coming up in `core-java`. Here, we make `ThrowableMessage` accept and return instances of `RejectionMessage` instead of `GeneratedMessage3`.

To make this change possible, all the standard marker interfaces now extend `java.io.Serializable`. 

Also, in this PR we add a generic type parameter to `MessageClass<M extends Message>`. This is required since different message kinds now have different interfaces.


dmdashenkov:
@armiol, PTAL.

PR #202, opened 2018-09-24
Fix `MessageClass` Interface

dmdashenkov:
Since recently, `MessageClass` is parametrized with a type of the stored message type.

In this PR we change the signature of the constructor from `MessageClass(Class<M> value)` to `MessageClass(Class<? extends M> value)`, so that the extended classes have some wiggle room.

Also, the signature of `ExceptionFactory` was changed to allow any kinds of `MessageClass`-es, regardless of the wrapper type.


dmdashenkov:
@armiol, PTAL.

dmdashenkov:
@armiol, let's keep the version number. This is rather a fix than an API change.

PR #205, opened 2018-09-27
Fix field masks

alexander-yevsyukov:
This PR fixes the way field masks are created and checked. Previously, we assumed that field masks contain full field names. The standard `FieldMaskUtil` creates masks with short names.

This PR also addresses Error Prone and Javadoc warnings.

### DIFF
--- a/tools/reflections-plugin/src/main/java/io/spine/tools/reflections/ReflectionsPlugin.java
+++ b/tools/reflections-plugin/src/main/java/io/spine/tools/reflections/ReflectionsPlugin.java
@@ -105,9 +105,9 @@ public class ReflectionsPlugin extends SpinePlugin {
         reflections.save(reflectionsOutputFilePath);
     }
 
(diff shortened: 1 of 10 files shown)


alexander-yevsyukov:
@dmitrykuzmin, PTAL.

tools/reflections-plugin/src/main/java/io/spine/tools/reflections/ReflectionsPlugin.java
```diff
@@ -105,9 +105,9 @@ private void scanClassPath(Project project) {
         reflections.save(reflectionsOutputFilePath);
     }
 
-    private static Set<URL> toUrls(File outputDir) {
+    @SuppressWarnings({"CollectionContainsUrl", "URLEqualsHashCode"})
         // because they are file URIs, they will not cause any network-related issues.
```
dmitrykuzmin: I think the indentation of the comment is not correct.

PR #208, opened 2018-10-01
Fix remaining warnings

alexander-yevsyukov:
This PR:
  * Addresses remaining Error Prone warnings.
  * Adds `Durations2`, `DurationStringifier`, `TimestampStringifier`, and `EnumStringifier`. These classes were copied and adjusted from Time, and will be removed there when migrating to newer dependency of `base`.

Other notable changes:
  * Abstract base for `FieldValidator` tests was introduced.


alexander-yevsyukov:
@armiol, PTAL.

base/src/test/java/io/spine/base/IdentifierTest.java
```diff
@@ -0,0 +1,386 @@
...
+        }
+    }
+
+    @Nested
+    @DisplayName("convert to String an ID of type")
+    class ConvertToString {
+
+        @Test
+        @DisplayName("Integer")
+        @SuppressWarnings("UnnecessaryBoxing") // OK as we want to show types clearly.
```
armiol: I would say `explicitly` instead of `clearly`. 

It's here and below.

base/src/test/java/io/spine/base/IdentifierTest.java
```diff
@@ -0,0 +1,386 @@
...
+        assertEquals(value, Type.STRING.fromMessage(toMessage(value)));
+    }
+
+    @Nested
+    @DisplayName("reject unsupported")
+    class RejectUnsupported {
+
+        @Test
+        @DisplayName("value")
+        @SuppressWarnings("UnnecessaryBoxing") // We want to make the unsupported type obvious.
```
armiol: It's interesting that the suppression is the same, but the comment is different from the comments above.

PR #212, opened 2018-10-04
Use non-final fields for `@Rule`

alexander-yevsyukov:
Also:
  * Bump Gradle -> `4.10.2`
  * Fix dependency on Spine Testlib


alexander-yevsyukov:
@dmdashenkov, PTAL.

PR #218, opened 2018-10-11
Avoid package dependency between `value` and `type`

alexander-yevsyukov:
This package removes `ClassTypeValue.getClassName()` method that returned `ClassName`, thus making `value` package depend on `type`.

Two test were also migrated to JUnit5.


alexander-yevsyukov:
@armiol, PTAL.

PR #223, opened 2018-10-16
Create rejections using builders

### LINKED ISSUE #86: `Oneof` fields are rewritten in the constructor of `ThrowableMessage` generated for a `Rejection`
At the moment we generate a class, derived from a `ThrowableMessage` for each of rejections. It stores the original rejection `Message` instance and wraps it into a `Throwable` item.

To allow an initialization of the wrapped message through the generated class constructor, we require to feed all the values for the message fields. 

However, for a `Rejection` message, containing the `oneof` field, such an approach fails. I.e. a rejection message such as

```
message CannotCreateProject {
     oneof reason {
          boolean duplicate = 1;
          boolean not_enough_credit = 2;
     }
} 
```
will result into generating a `ThrowableMessage` descendant class:

```java 
@Generated("by Spine compiler")
public class CannotCreateProject extends ThrowableMessage {

    public CannotCreateProject(boolean duplicate, boolean notEnoughCredit) { 
        super(Rejections.CannotCreateProject.newBuilder()
                                            .setDuplicate(duplicate)
                                            .setNotEnoughCredit(notEnoughCredit)
                                            .build());
}
```
According to the `oneof` specification, any setter called for one of its members resets the values for other members to their defaults. In this particular example in all cases the only meaningful value for `CannotCreateProject.getReason()` will be `notEnoughCredit()`, as its setter is called the last.

To address the issue we need to take into account a `oneof` specifics and allow to submit only one of `oneof` values through the constructor parameters.

dmytro-grankin:
This PR fixes #86.

### The Rejections API changes

Previously, throwable rejections were created using constructors. Now, builders are used.
Also, it is worth to note, that during the object construction, a `RejectionMessage` is validated using `Validate.checkValid(Message)`.

Creation of a rejection now looks like:

```java
ThrowableMessage rejection = CannotUpdateUsername.newBuilder()
                                                 .setUsername("name")
                                                 .build()
```

### The motivation

The previous approach was not flexible enough, i.e. it was not possible to work with `oneof` fields.

### Implementation details

When generating a builder for a throwable rejection, a setter is generated for every field of a message. A generated builder uses a Protobuf builder for a message under the hood.

Some abstractions were extracted into new classes:
- `MessageDocumentation` — provides access to comments of a Protobuf message declaration.
- `FieldDeclaration` — provides detailed data about a field declaration in a Protobuf file.

To properly test the generation of rejections, a new `rejection-generation` module was created in `smoke-tests`.

Also, Doclet-based `RootDocReceiver` (for Javadocs extracting from sources) was replaced by `Roaster` library. The reason for the change is that previously Javadocs were regenerated using a Doclet and it was quite inconvenient. Now the comments just extracted from sources.

The version was bumped to `0.11.6-SNAPSHOT`.


dmytro-grankin:
@armiol PTAL.

dmytro-grankin:
@armiol PTAL again.

base/src/main/java/io/spine/code/proto/AbstractMessageDeclaration.java
```diff
@@ -69,6 +72,22 @@ public FileDescriptorProto getFile() {
...
+    public MessageDocumentation documentation() {
+        return new MessageDocumentation(this);
+    }
+
+    /**
+     * Obtains declarations of the message fields.
+     *
+     * @return fields in the order of their declaration
+     */
+    public List<FieldDeclaration> fields() {
```
armiol: I think it should be `ImmutableList`. We aim for clarity. There is a related [reading](http://errorprone.info/bugpattern/MutableConstantField).

base/src/main/java/io/spine/code/proto/FieldDeclaration.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    public FieldDescriptorProto descriptor() {
+        return descriptor;
+    }
+
+    /**
+     * Obtains comments going before the field.
+     *
+     * @return the leading field comments
```
armiol: It returns `Optional`, so let's say so and describe the cases.

tools/model-compiler/src/main/java/io/spine/tools/compiler/rejection/GeneratedRejectionDeclaration.java
```diff
@@ -0,0 +1,74 @@
...
+
+package io.spine.tools.compiler.rejection;
+
+import com.squareup.javapoet.ClassName;
+import io.spine.code.java.PackageName;
+import io.spine.code.java.SimpleClassName;
+import io.spine.code.proto.RejectionDeclaration;
+
+/**
+ * A declaration of a generated rejection.
```
armiol: I don't understand how this is different from `RejectionDeclaration`. It's either the naming (e.g. `GeneratedRejectionDeclaration` should be renamed) or the purpose. Besides, with this naming it's unclear why `RejectionDeclaration` is in `base`, and this one is in `model-compiler`.

tools/model-compiler/src/main/java/io/spine/tools/compiler/rejection/RejectionBuilder.java
```diff
@@ -0,0 +1,240 @@
...
+
+import static com.google.common.collect.Lists.newArrayList;
+import static com.squareup.javapoet.MethodSpec.constructorBuilder;
+import static javax.lang.model.element.Modifier.FINAL;
+import static javax.lang.model.element.Modifier.PRIVATE;
+import static javax.lang.model.element.Modifier.PUBLIC;
+import static javax.lang.model.element.Modifier.STATIC;
+
+/**
+ * Specification of a rejection builder.
```
armiol: It does much more than just describing the specification. It generates code. 

I think we should either split into two classes (one for specification, one for generation) or rename.

tools/model-compiler/src/main/java/io/spine/tools/compiler/rejection/RejectionWriter.java
```diff
@@ -121,62 +124,77 @@ public void write() {
...
+                         .addStatement("return ($T) super.$L", returnType, methodSignature)
                          .build();
     }
 
+    /**
+     * Generates a Javadoc content for the rejection.
+     *
+     * @return the class-level Javadoc content
+     */
+    private CodeBlock classJavadoc() {
```
armiol: Now two classes generate code blocks for the rejection: this one and the `RejectionBuilder`. It's weird.

dmytro-grankin: How would you separate responsibilities? For me, it looks ok, that the `RejectionWriter` delegates the builder generation to the `RejectionBuilder`.

dmytro-grankin:
@armiol PTAL.

PR #225, opened 2018-10-17
Don't attempt to generate validating builders for rejections

### LINKED ISSUE #220: Do not attempt to generate `...VBuilder` for rejection events
Currently we are still trying to generate validating builders for rejection events — while they are created not via builder-based approach, but through calling a generated constructor. 

As long as rejection events are generated as nested classes by the Protobuf compiler, the console gets the warnings such as this one:

```
Error: java.lang.IllegalArgumentException: The io.spine.system.server.CannotDispatchCommandTwice class is not found.
Cannot generate the validating builder for VBType{javaClass=CannotModifyArchivedEntityVBuilder, javaPackage=io.spine.server.entity.rejection, sourceProtoFile=spine/server/entity/standard_rejections.proto}. 
```

In fact we shouldn't be even trying to generate the builders from the rejection event messages.

dmytro-grankin:
This PR fixes #220.

Now, during the lookup of validating builder to generate, rejections are filtered.

Also, `VBTypeLookup` now uses `FileDescriptors.parseSkipStandard(...)` to skip standard Protobuf definitions instead of a custom code.

The version was bumped to `0.11.7-SNAPSHOT`.


dmytro-grankin:
@armiol PTAL.

PR #226, opened 2018-10-17
Remove the log buffer size limitation

dmdashenkov:
Before now, tests marked with `@MuteTests` could log more than 1 MiB of data. In this PR we remove this limitation by migrating from `ByteBuffer` to `List<Byte>`.

Also, in this PR we allow such tests to be executed in parallel without a risk of losing logs.


dmdashenkov:
@armiol, PTAL.

PR #227, opened 2018-10-18
Generate a UUID message using `Identifier`

### LINKED ISSUE #206: Generate UUID-based values in `Identifier`
We can generate identifiers that have as single string field called `uuid` without extending Model Compiler.

It would require a factory method:

`public static I generate(Class<I> idClass) { ... }`

It would then obtain a descriptor by the message class, then field descriptor.

We have already `newUuid()` method returning `String`.

dmytro-grankin:
This PR fixes #206.

Now, it is possible to generate a so-called `UuidMessage` using `Identifier`. The message is a Protobuf `Message` with a single string field named `uuid`. During the generation, the `uuid` field is filled with a value obtained from `Identifier.newUuid()`.

So, if you have a message like:

```proto
message SessionId {
    string uuid = 1;
}
```

You can generate it in this way:

```java
SessionId id = Identifier.generate(SessionId.class);
```

If a message doesn't match the requirements, an exception is thrown.

The framework version was bumped to `0.11.9-SNAPSHOT`.

### DIFF
--- a/base/src/main/java/io/spine/base/Identifier.java
+++ b/base/src/main/java/io/spine/base/Identifier.java
@@ -22,6 +22,8 @@ package io.spine.base;
 
 import com.google.common.reflect.TypeToken;
 import com.google.protobuf.Any;
+import com.google.protobuf.Descriptors;
+import com.google.protobuf.Descriptors.FieldDescriptor;
 import com.google.protobuf.Int32Value;
 import com.google.protobuf.Int64Value;
 import com.google.protobuf.Message;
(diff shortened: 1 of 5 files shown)


dmytro-grankin:
@armiol PTAL.

base/src/main/java/io/spine/base/Identifier.java
```diff
@@ -510,4 +532,47 @@ Message fromMessage(Message message) {
             return result;
         }
     }
+
+    /**
+     * A message containing a single string field with the {@linkplain #FIELD_NAME name}.
+     */
+    private static class UuidMessage<I extends Message> {
```
armiol: Please convert to a package-private class at the top level. The reason is that `Identifier` class is long enough already.

base/src/test/java/io/spine/base/IdentifierTest.java
```diff
@@ -368,6 +369,29 @@ void clazz() {
...
+
+        @Test
+        @DisplayName("the field name is not `uuid`")
+        void nameIsInvalid() {
+            assertThrows(IllegalStateException.class, () -> Identifier.generate(StringValue.class));
+        }
+
+        @Test
+        @DisplayName("it contains more than one field")
+        void moreThanTwoFields() {
```
armiol: The name of the method does not correspond to its display name.

PR #228, opened 2018-10-19
Assume entity IDs required by default

### LINKED ISSUE #179: Assume ID fields required
We need to make it easier to declare identifier fields. Let's support the following convention:

A field which is named `id`, or which name ends on `_id` is required, unless it is annotated with `[(required)=false]`.

dmytro-grankin:
This PR fixes #179.

Now, the framework assumes, that an entity ID is required by default.

The only way to say that an entity ID is not required is to explicitly mark it as `[(required)=false]`.

A field is considered an entity ID if:
- The name is `id` or ends with `_id`.
- The field if the first in a containing message.
- A message containing the field has `(entity).kind` option set.

The version was bumped to `0.11.10-SNAPSHOT`.

### DIFF
--- a/base/src/main/java/io/spine/code/proto/Option.java
+++ b/base/src/main/java/io/spine/code/proto/Option.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright 2018, TeamDev. All rights reserved.
+ *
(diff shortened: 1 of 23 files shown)


dmytro-grankin:
@armiol PTAL.

Issue #201, opened 2018-09-22
Test `AbstractValidatingBuilder` conversion


alexander-yevsyukov:
`convertToMap()` and `convertToList()` appear as not used in the project, which tells me they are not tested. Please look into this.

dmytro-grankin:
Closing the issue since it was fixed in #232.

PR #232, opened 2018-10-24
Get rid of unused type parameters in formal parameters

### LINKED ISSUE #200: Address the warnings `TypeParameterUnusedInFormals`
In most of these places we do a dis-service. See http://errorprone.info/bugpattern/TypeParameterUnusedInFormals for details.

The most sensitive part is `AbstractValidatingBuilder.convert()` method.

dmytro-grankin:
This PR fixes #200, #201.

The API Changes:
- `Stringifiers.fromString(...)` now takes `Class` instead of `Type`.
- `Identifier.unpack(Any)` now returns `Object`. If a result type is important, use `Identifier.unpack(Any, Class)`.
- `TypeConverter.toMessage(Object)` now returns `Message`. If a result type is important , use `TypeConverter.toMessage(Object, Class)`.
- `AbstractValidatingBuilder.convert(...)` now takes `Class` instead of `Type`.

The version was bumped to `0.11.13-SNAPSHOT`.

### DIFF
--- a/base/src/main/java/io/spine/base/Identifier.java
+++ b/base/src/main/java/io/spine/base/Identifier.java
@@ -41,7 +41,6 @@ import java.util.UUID;
 import java.util.regex.Pattern;
 
 import static com.google.common.base.Preconditions.checkNotNull;
-import static com.google.common.base.Preconditions.checkState;
 import static com.google.protobuf.TextFormat.shortDebugString;
 import static io.spine.util.Exceptions.newIllegalArgumentException;
 import static io.spine.util.Exceptions.newIllegalStateException;
@@ -142,9 +141,12 @@ public final class Identifier<I> {
      *   <li>Email address as a couple of local-part and domain
      * </ul>
      *
-     * @param <I>     the type of the ID
-     * @param idClass the class of IDs
-     * @throws IllegalArgumentException if the class of IDs is not of supported type
+     * @param <I>
+     *         the type of the ID
+     * @param idClass
+     *         the class of IDs
+     * @throws IllegalArgumentException
+     *         if the class of IDs is not of supported type
      */
     public static <I> void checkSupported(Class<I> idClass) {
         checkNotNull(idClass);
@@ -166,10 +168,13 @@ public final class Identifier<I> {
      *   <li>For {@code Integer} — {@link Int32Value}
      * </ul>
      *
-     * @param id  the value to wrap
-     * @param <I> the type of the value
+     * @param id
+     *         the value to wrap
+     * @param <I>
+     *         the type of the value
      * @return instance of {@link Any} with the passed value
-     * @throws IllegalArgumentException if the passed value is not of the supported type
+     * @throws IllegalArgumentException
+     *         if the passed value is not of the supported type
      */
     public static <I> Any pack(I id) {
         checkNotNull(id);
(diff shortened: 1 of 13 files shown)


base/src/main/java/io/spine/base/Identifier.java
```diff
@@ -246,17 +263,20 @@ public static String newUuid() {
...
      * @return <ul>
-     * <li>for classes implementing {@link Message} &mdash; a Json form;
-     * <li>for {@code String}, {@code Long}, {@code Integer} &mdash;
-     * the result of {@link Object#toString()};
-     * <li>for {@code null} ID &mdash; the {@link #NULL_ID};
-     * <li>if the result is empty or blank string &mdash; the {@link #EMPTY_ID}.
-     * </ul>
-     * @throws IllegalArgumentException if the passed type isn't one of the above or
-     *                                  the passed {@link Message} instance has no fields
+     *         <li>for classes implementing {@link Message} &mdash; a Json form;
```
dmdashenkov: Let's get rid of `&mdash;` and simply use `-` (minus) instead. This would improve readability of the raw Javadoc. 

Given that the compiled Javadoc is not read as frequently as its raw form in the real world, this is a good thing.

base/src/main/java/io/spine/base/Identifier.java
```diff
@@ -246,17 +263,20 @@ public static String newUuid() {
...
+     *         <li>for {@code String}, {@code Long}, {@code Integer} &mdash;
```
dmdashenkov: Please align the list items properly.

base/src/main/java/io/spine/protobuf/TypeConverter.java
```diff
@@ -106,21 +106,33 @@ private TypeConverter() {
...
-    // Parameters are handled by MessageCaster, that we wrap by this method.
-    public static <T, M extends Message> M toMessage(T value) {
-        @SuppressWarnings("unchecked") // Must be checked at runtime
-        Class<T> srcClass = (Class<T>) value.getClass();
-        MessageCaster<M, T> caster = MessageCaster.forType(srcClass);
-        M message = caster.reverse()
-                          .convert(value);
+    public static <T> Message toMessage(T value) {
+        // Must be checked at runtime
+        @SuppressWarnings("unchecked") Class<T> srcClass = (Class<T>) value.getClass();
```
dmdashenkov: Please move the statement itself to the next line without any indentation.

base/src/test/java/io/spine/protobuf/TypeConverterTest.java
```diff
@@ -56,123 +58,123 @@ protected void configure(NullPointerTester tester) {
...
+        @DisplayName("FloatValue to float")
+        void map_FloatValue_to_float() {
+            float rawValue = 42.0f;
+            Message value = FloatValue.of(rawValue);
+            checkMapping(rawValue, value);
+        }
+
+        @Test
+        @DisplayName("DoubleValue to double")
+        void map_DoubleValue_to_double() {
```
dmdashenkov: Please either fix the naming or create an issue.

dmytro-grankin: See #234.

PR #233, opened 2018-10-26
Handle numbers in Protobuf field names when converting to CamelCase

### LINKED ISSUE #211: Fix model compilation errors for field names with numbers
The problem appears on fields with names like `logo_link_40dp` in `.proto` files (like can be seen [here](https://github.com/TeamDev-Ltd/License-Server/blob/ca188288a25462eba8dac57659bc5d82a5cc3e91/model/src/main/proto/teamdev/licensing/c/product/values.proto#L64)). The capitalization of getter/setter names for such fields is different in different files generated during the build.

Model compilation results in a generated code like this:
```java
    ...
    getMessageBuilder().setLogoLink40dp(logoLink40dp);
```
Whereas the actual setter method that was generated in the builder class used here has a name of `setLogoLink40Dp` (note the `40dp` vs `40Dp` difference):
```java
    public Builder setLogoLink40Dp(io.spine.net.Url value) {
     ...
```

This leads to compilation errors like this:
```
 C:\Projects\TeamDev\TeamDev-Ltd\License-Server\model\generated\main\spine\com\teamdev\licensing\c\product\ProductLogoVBuilder.java:34: error: cannot find symbol
        getMessageBuilder().setLogoLink40dp(convertedValue);
                           ^
  symbol:   method setLogoLink40dp(Url)
  location: class Builder
```

dmytro-grankin:
This PR fixes #211.

Now, for a Proto field name, the next letter after a digit is capitalized in the way the Protoc compiler [does](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/compiler/java/java_helpers.cc#L161). The change was made in `FieldName.words()` and now the words are split not only by an underscore but also by a digit.

Other changes:
- `FieldTypes` was renamed to `FieldTypesProto` and moved to `io.spine.code.proto` with minor implementation improvements.
- `Fields` was renamed to `FieldTypes` and moved `io.spine.code.proto`. Now, it reuses some code from `FieldTypesProto` and is reused in `FieldValidatorFactory`.

The difference between `FieldTypesProto` and `FieldTypes` is that one works with `FieldDescriptorProto` and one with `FieldDescriptor`. Because of that, it was decided to keep two classes even though they are very similar.

The version was bumped to `0.11.14-SNAPSHOT`.


dmytro-grankin:
@dmdashenkov PTAL.

base/src/main/java/io/spine/code/proto/FieldTypesProto.java
```diff
@@ -0,0 +1,130 @@
...
+     * @return the type name without leading dot
+     */
+    public static String trimTypeName(FieldDescriptorProto fieldDescriptor) {
+        String typeName = fieldDescriptor.getTypeName();
+        checkNotNull(typeName);
+
+        if (typeName.isEmpty()) {
+            return typeName;
+        }
+        // it has a redundant dot in the beginning
```
dmdashenkov: Please remove the comment. If the code is not readable enough, extract some variables/functions in order to make it sound better.

base/src/main/java/io/spine/code/proto/CamelCase.java
```diff
@@ -52,4 +52,35 @@ public static String convert(UnderscoredName name) {
...
+                                        .iterator();
+        StringBuilder builder = new StringBuilder(name.value()
+                                                      .length());
+        while (iterator.hasNext()) {
+            String word = iterator.next();
+            boolean capitalizeNext = true;
+            for (char sourceChar : word.toCharArray()) {
+                char charToAppend = capitalizeNext
+                                    ? Character.toUpperCase(sourceChar)
+                                    : sourceChar;
```
dmdashenkov: Please add a test on what happens if a field has a capital letter in the middle of a word. For example, `type_URL`. 

This is a rare case, since it does not adhere to the Protobuf naming convention, but still, we better check it.

base/src/main/java/io/spine/code/proto/CamelCase.java
```diff
@@ -52,4 +52,35 @@ public static String convert(UnderscoredName name) {
...
+     * @see <a href="https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/compiler/java/java_helpers.cc#L161">
+     *         Protoc Camel Case</a>
+     */
+    @SuppressWarnings("MethodWithMultipleLoops" /* Keep specific logic inside one method. */ )
+    public static String protocStyled(UnderscoredName name) {
+        Iterator<String> iterator = name.words()
+                                        .iterator();
+        StringBuilder builder = new StringBuilder(name.value()
+                                                      .length());
+        while (iterator.hasNext()) {
```
dmdashenkov: I don't see why do we use an iterator, not a `for` loop.

dmytro-grankin:
@dmdashenkov PTAL again.

base/src/main/java/io/spine/code/proto/FieldName.java
```diff
@@ -65,10 +71,25 @@ public static FieldName of(FieldDescriptorProto field) {
...
     /**
      * Obtains immutable list of words used in the name of the field.
+     *
+     * <p>A word is a part of the name, the first letter of which should be capitalized
+     * when converting {@linkplain #toCamelCase() to CamelCase}.
+     *
+     * <p>So, the name is split by:
+     * <ul>
+     *   <li>an underscore excluding it from a word;</li>
+     *   <li>a digit leaving it in a word.</li>
```
dmdashenkov: Seems that there are 2 spaces instead of 4 before the `<li>`.

PR #238, opened 2018-11-01
Improve build performance

dmdashenkov:
# WIP

### DIFF
--- a/base/src/main/java/io/spine/code/proto/FileDescriptors.java
+++ b/base/src/main/java/io/spine/code/proto/FileDescriptors.java
@@ -26,15 +26,11 @@ import io.spine.io.ResourceFiles;
 import io.spine.logging.Logging;
 import org.slf4j.Logger;
 
-import java.io.BufferedOutputStream;
 import java.io.File;
 import java.io.FileInputStream;
-import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.OutputStream;
 import java.net.URL;
-import java.util.Collection;
 import java.util.Iterator;
 import java.util.List;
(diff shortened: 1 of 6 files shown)


dmdashenkov:
@alexander-yevsyukov, PTAL.

dmdashenkov:

tools/model-compiler/src/main/java/io/spine/tools/compiler/descriptor/FileDescriptorSuperset.java
```diff
@@ -0,0 +1,133 @@
...
+import static io.spine.util.Exceptions.illegalStateWithCauseOf;
+import static io.spine.util.Exceptions.newIllegalStateException;
+import static java.util.stream.Collectors.toSet;
+
+/**
+ * A set of {@code FileDescriptorSet}s.
+ */
+public final class FileDescriptorSuperset implements Logging {
+
+    private final Collection<FileDescriptorSet> descriptors;
```
alexander-yevsyukov: It would be better to have the field as `LinkedList` and document it why we have it this way. Having it as `Collection` is misleading.

alexander-yevsyukov: Or, have it just as `List` with `ArrayList` implementation, if we don't need specifically `LinkedList`.

alexander-yevsyukov: But we operate with `Set`s in methods that collection `FileDescriptorProto`. Why `List` here?

PR #239, opened 2018-11-01
Get rid of Soy templates

### LINKED ISSUE #79: Consider migrating from Soy for code generation
We depend on Java Poet already. `MethodConstructors` uses Soy.

It would be good to standardize on one method.

dmytro-grankin:
This PR fixes #79.

All Soy templates and related dependencies were removed. Now, only `JavaPoet` is used for the code generation.

Also, the `AbstractMethodConstructor` was introduced to remove some duplications and make the code more object-oriented.


tools/model-compiler/src/main/java/io/spine/tools/compiler/validation/AbstractMethodConstructor.java
```diff
@@ -0,0 +1,114 @@
...
+        String javaClass = checkNotNull(builder.getJavaClass());
+        this.builderClass = getClassName(javaPackage, javaClass);
+    }
+
+    /**
+     * Obtains the the statement, which declares the descriptor
+     * for the {@linkplain #fieldIndex field}.
+     *
+     * @return the statement like
+     *         {@code FieldDescriptor fieldDescriptor = Msg.getDescriptor().getFields().get(0)}
```
dmdashenkov: Please replace `0` with `fieldIndex`.

PR #240, opened 2018-11-02
Properly validate `Oneof` fields

### LINKED ISSUE #154: Unable to make `oneof` options required
It's not possible to require each of the option of `oneof`. When one of the options is set, the exception is thrown for not having the other.

dmytro-grankin:
This PR fixes #154.

Now `Oneof` fields are considered in a special way and only the actually set field is validated. For example, if a `Oneof` contains 3 fields, the only one is validated. The work is done by a newly introduced `OneofValidator`.

The important note is that if none of a `Oneof` fields are set, a declaration is considered invalid.

So, after the change, it is possible to create a message where every `Oneof` field is `required` (see the `EveryRequired` below). And the message is considered valid if the `first` or `second` field is set.

```proto
message EveryRequired {
    oneof fields {
        string first = 1 [(required) = true];
        string second = 2 [(required) = true];
    }
}
```

Also, the `config` was updated.


dmytro-grankin:
@armiol PTAL.

base/src/main/java/io/spine/validate/OneOfValidator.java
```diff
@@ -0,0 +1,81 @@
...
+        return violations.build();
+    }
+
+    private ConstraintViolation noneFieldIsSet() {
+        FieldPath oneOfPath = message.context()
+                                     .getFieldPath()
+                                     .toBuilder()
+                                     .addFieldName(oneOf.getName())
+                                     .build();
+        ConstraintViolation requiredFieldNotFound =
```
alexander-yevsyukov: Can we have the type on the previous line (preparing for Java 11 `var` syntax)?
```suggestion
        ConstraintViolation requiredFieldNotFound = ConstraintViolation
                                   .newBuilder()
```

base/src/test/java/io/spine/validate/MessageValidatorTest.java
```diff
@@ -971,25 +974,64 @@ void notRequiredIfOptionIsFalse() {
...
     }
 
+    @Nested
+    @DisplayName("consider OneOf")
+    class OneOf {
+
+        @Test
+        @DisplayName("valid if a required field is set to a non-default value")
+        void validIfRequireFieldIsNotDefault() {
+            EveryRequired requiredIsNotDefault = EveryRequired.newBuilder()
```
alexander-yevsyukov: Please have `newBuilder()` here and below in new lines.

base/src/main/java/io/spine/validate/OneOfValidator.java
```diff
@@ -0,0 +1,81 @@
...
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * Validates a {@linkplain OneofDescriptor OneOf}.
+ *
+ * <p>Only a single field from a {@code OneOf} is validated — the field that is actually set.
+ * If none of fields is set, a constraint violation is created.
+ *
+ * @see <a href="https://developers.google.com/protocol-buffers/docs/proto3#oneof">One Of documentation</a>
```
armiol: As long as Protobuf documentation references the language element as `Oneof`, let's do the same in our code, docs and PR description. I.e. not `OneOf` or `One Of`.

base/src/test/proto/spine/test/validate/oneof.proto
```diff
@@ -0,0 +1,53 @@
...
+    oneof fields {
+        string first = 1 [(required) = true];
+        string second = 2 [(required) = true];
+    }
+}
+
+message OneRequired {
+    oneof fields {
+        string required_val = 1 [(required) = true];
+        string non_required = 2 [(required) = false];
```
armiol: It should probably be `optional`.

base/src/test/proto/spine/test/validate/oneof.proto
```diff
@@ -0,0 +1,53 @@
...
+    }
+}
+
+message EveryNotRequired {
```
armiol: And this one `EveryOptional`.

base/src/main/java/io/spine/validate/MessageValue.java
```diff
@@ -70,10 +72,17 @@ static MessageValue atTopLevel(Message message) {
         return new MessageValue(message, FieldContext.empty());
     }
 
-    /** Returns field values of the message. */
-    ImmutableList<FieldValue> fields() {
+    /**
+     * Obtains field values of the message.
+     *
+     * <p>Values of {@code OneOf} fields are filtered and not returned.
```
armiol: `...are filtered out...`

base/src/main/java/io/spine/validate/MessageValue.java
```diff
@@ -89,11 +98,20 @@ static MessageValue atTopLevel(Message message) {
...
-        return Optional.of(fieldValue);
+        return valueOfNullable(field);
+    }
+
+    /**
+     * Obtains the value of a populated {@code OneOf} field.
+     *
+     * @param oneOf
+     *         the {@code OneOf} descriptor
+     * @return a value of the populated field or {@code Optional.empty()} if the message
```
armiol: It seems to me that if the message does not contain the field, which description was passed to the method, we should throw an invalid argument exception.

dmytro-grankin:
@alexander-yevsyukov PTAL.

dmytro-grankin:
@armiol PTAL.

PR #245, opened 2018-12-07
Improve type caching and code gen.

alexander-yevsyukov:
This PR:
  * Pulls the `ExceptionFactory` of Validation to the package level.
  * Puts validation errors into one proto file.
  * Improves work with type caching in Spine Model Compiler.
  * Extracts common code in annotators.
  * Simplifies naming in validating builder generators.

Notable API changes:
   * `MessageTypeCache` was renamed into `TypeCache` since we cache enum types too.
   * Loading of types was extracted into small type hierarchy to simplify the logic.
   * `Annotator` got new method `annotate(SourceFile)`.
   * `TypeCache` is used for passing type information instead of `Map`.


alexander-yevsyukov:
@dmytro-grankin, @armiol, PTAL.

PR #248, opened 2018-12-11
Extensive `(is)` option

dmdashenkov:
This PR introduces an improvement to the message marker interface declaration. This PR fixes [core-java/#911](https://github.com/SpineEventEngine/core-java/issues/911).

### Summary

Previously, the only possibility to mark a message with a marker interface was to generate the interface with the help of `protoc` itself. 

From now on, it is possible for marker interfaces to be handcrafted, instead of generated.

### Handcrafted Interfaces

In order to declare that a message Java class should implement an interface, the user should annotate it with `(is)` option as follows:
```proto
message Student {
    option (is).java_type = "org.example.customer.FreemiumCustomer";

    string institution_name = 1;
    // ...
}

message OpenSourceCustomer {
    option (is).java_type = "org.example.customer.FreemiumCustomer";

    string company_name = 1;
    // ...
}
```

Now, in Java, declare `org.example.customer.FreemiumCustomer` interface:
```java
package org.example.customer;

public interface FreemiumCustomer {

    default String getInstitutionOrCompanyName() {
        // Use Java or Protobuf reflection to obtain the value.
    }
}
```
 The interface may have any methods and/or extend any other interfaces as long as it does not introduce new abstract methods to the message class. If the method is already present in the end class or has a `default` implementation, it is allowed.

### Generated Interfaces

In order to generate the marker interface, the `(is).generate` option field should be set to `true` (the default value is `false`):

```proto
message Rgb {
    option (is) = {java_type: "Color", generate: true};

   // ...
}

message Cmyk {
    option (is).java_type = "Color";

   // ...
}
```
Note that it is sufficient to specify `generate = true` only once. This detail may be omitted on further message declarations. However, it is OK to specify to generate the interface on all the messages.

### `(every_is)` Option

The `(every_is)` option is a "brother" of the `(is)` option targeting an entiere file instead of a single message type. The API of `(every_is)` is the same as the API of `(is)`.


dmdashenkov:
@armiol, PTAL.

base/src/main/proto/spine/options.proto
```diff
@@ -848,3 +842,37 @@ message EntityOption {
...
+    // May be an FQN or a simple name. In the latter case, the interface should belong to the same
+    // Java package as the message class which implements this interface.
+    //
+    // The framework does not ensure the referenced type exists.
+    // If the generation is disabled, the Java type is used as-is. Otherwise, a corresponding Java
+    // interface is generated.
+    //
+    // A generated interface has no declared methods and extends `com.google.protobuf.Message`.
+    // The `.java` file is placed alongside with the code generated by the proto-to-java compiler.
+    // If specified with an FQN, the package of the generated type matches the FQN.
```
armiol: This sentence seems to be a new paragraph, as it's about how to reference a type.

PR #249, opened 2018-12-12
Fix a ClassCastException in FieldValue

dmdashenkov:
A `FieldValue` determines the Java type of its value by the field descriptor: if the field is `repeated`, the value is assumed to be a `List`.

However, when a `ValidatingBuilder` validates a value received in an `add<Field>` method (an item added to a repeated field), the runtime de-facto value is not of type `List` but of the field type, e.g. a message.

From now, `FieldValue` checks the actual Java type of the value instead of the descriptor.


dmdashenkov:
@armiol, PTAL.

PR #251, opened 2018-12-12
Skip imports when generating marker interfaces

dmdashenkov:
Before, we tried to generate marker interfaces and insertion points for Protobuf sources which are _included_ for a Proto compilation but are not requested, i.e. Protobuf imports. This led to failures in multi-project Gradle builds when several projects depending on each other had some marker interfaces to generate.

From now, the markers are generated for the requested types only.


dmdashenkov:
@armiol, PTAL.

PR #255, opened 2018-12-17
Generate a method providing `TypeUrl` for Protobuf types compiled to JS

dmytro-grankin:
A possibility to obtain a `TypeUrl` for a message is convenient, but Protobuf definitions compiled to JS don't provide such a possibility.

So, to allow convenient obtaining of `TypeUrl`, this PR enhances `ProtoJsPlugin`. The plugin now generates the static `typeUrl()` method for every message and enum type.

Other changes:
- The naming of classes related to JS code-generation was improved.
- The package structure for `ProtoJsPlugin` was changed to keep related classes in a single package.
- A code `Snippet` was introduced as a replacement of `JsCodeGenerator`. The `Snippet` has no side effects as the generator does.
- `CodeLine` was made abstract and now have different implementations.
- The configuration of test projects via `GradleProject` was changed. Now, instead of copying `config` submodule and `version.gradle`, the path to the root of `spine-base` is supplied. So, test configuration can reference e.g. the `version.gradle` via `${enclosingRootDir}/version.gradle`.

The further improvements is a matter of #258.


dmytro-grankin:
@dmdashenkov PTAL.

base/src/main/java/io/spine/code/proto/EnumType.java
```diff
@@ -46,7 +47,8 @@ private EnumType(EnumDescriptor descriptor,
         super(descriptor, descriptorProto, className, typeUrl);
     }
 
-    private static EnumType create(EnumDescriptor descriptor) {
+    @VisibleForTesting
+    public static EnumType create(EnumDescriptor descriptor) {
```
dmdashenkov: Please find a way to make this method package-private. `public` is way too broad.

base/src/main/java/io/spine/code/proto/MessageType.java
```diff
@@ -45,7 +46,11 @@ protected MessageType(Descriptor descriptor,
         super(descriptor, descriptorProto, className, typeUrl);
     }
 
-    private static MessageType create(Descriptor descriptor) {
+    /**
+     * Creates a message type for the provided descriptor.
+     */
+    @VisibleForTesting
+    public static MessageType create(Descriptor descriptor) {
```
dmdashenkov: Same as above.

tools/proto-js-plugin/src/main/java/io/spine/js/generate/FileSetEnhancement.java
```diff
@@ -0,0 +1,84 @@
...
+        this.fileSet = checkNotNull(fileSet);
+    }
+
+    /**
+     * {@linkplain #processSources() Enhances} generated sources.
+     *
+     * <p>Does nothing if there are no Protobuf files to process.
+     */
+    public final void perform() {
+        if (!hasFilesToProcess()) {
```
dmdashenkov: Please (un-)invert the condition. This way the code becomes a bit more descriptive.

tools/proto-js-plugin/src/main/java/io/spine/js/generate/JsType.java
```diff
@@ -0,0 +1,87 @@
...
+
+    public static JsType generatedFrom(Type type) {
+        return new JsType(type);
+    }
+
+    /**
+     * Obtains the name of the type in the generated code.
+     */
+    public TypeName name() {
+        return TypeName.from(type.descriptor());
```
dmdashenkov: There is the `Type.name()` method for this.

tools/proto-js-plugin/src/main/java/io/spine/js/generate/parse/TypesParsingExtension.java
```diff
@@ -37,13 +36,13 @@
  * <p>The class generates the {@code fromJson} and {@code fromObject} methods for each message
  * declared in the {@link FileDescriptor}.
  */
-public final class FileGenerator extends JsCodeGenerator {
+public final class TypesParsingExtension extends JsCodeGenerator {
```
dmdashenkov:
```suggestion
public final class TypeParsingExtension extends JsCodeGenerator {
```

tools/proto-js-plugin/src/main/java/io/spine/js/generate/typeurl/TypeUrlMethod.java
```diff
@@ -0,0 +1,70 @@
...
+        this.type = JsType.generatedFrom(type);
+    }
+
+    /**
+     * Generates the static {@code typeUrl()} method for the type.
+     */
+    @Override
+    public void generate() {
+        generateStaticMethod();
+        jsOutput().addEmptyLine();
```
dmdashenkov: The `TypeUrlMethod` class, apparently, as well as all `JsCodeGenerator`s use an output parameter `jsOutput` instead of returning values. Such a pattern adds a side effect to the `JsCodeGenerator.generate()` method.

Instead, we should make the method [pure](https://en.wikipedia.org/wiki/Pure_function) and write the result to the output when necessary.

tools/proto-js-plugin/src/main/java/io/spine/js/generate/FileSetEnhancement.java
```diff
@@ -0,0 +1,84 @@
...
+
+import io.spine.code.js.Directory;
+import io.spine.code.proto.FileSet;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * An enhancement of generated Protobuf sources.
+ */
+public abstract class FileSetEnhancement {
```
dmdashenkov: Please elaborate on the class description. Maybe, you'll come up with a better name while doing so. Currently, I do not entirely understand the concept.

tools/proto-js-plugin/src/main/java/io/spine/js/generate/parse/ParsingOfObjects.java
```diff
@@ -60,57 +58,34 @@
  *         message stored in a file.
  * </ol>
  */
-public final class JsonParsersWriter {
+public final class ParsingOfObjects extends FileSetEnhancement {
```
dmdashenkov: This class name is strange. From the name, it seems like the object describes the process of parsing.
Please think of a better name.

tools/proto-js-plugin/src/main/java/io/spine/js/generate/JsType.java
dmytro-grankin: It's a different `TypeName`, from `io.spine.code.js`.

dmytro-grankin:
@dmdashenkov PTAL. I have tried to get rid of `JsCodeGenerator` as much as possible, but they are still present.

tools/proto-js-plugin/src/main/java/io/spine/js/generate/Method.java
```diff
@@ -0,0 +1,114 @@
...
+         */
+        public Builder withArguments(String... arguments) {
+            this.arguments = newArrayList(arguments);
+            return this;
+        }
+
+        /**
+         * Adds a line to the body of the method.
+         */
+        public Builder addBodyLine(String line) {
```
dmdashenkov: `appendBody`.

tools/proto-js-plugin/src/main/java/io/spine/js/generate/Statements.java
```diff
@@ -0,0 +1,60 @@
...
+        checkNotNull(value);
+        String statement = "return " + value + ';';
+        return statement;
+    }
+
+    /**
+     * Composes a map entry with the string literal key.
+     */
+    public static String mapEntry(String key, Object value) {
+        return "['" + key + "', " + value + ']';
```
dmdashenkov: Shouldn't there be a `:` symbol between the key and the value?

tools/proto-js-plugin/src/main/java/io/spine/js/generate/CodeLines.java
```diff
@@ -42,7 +43,7 @@
 @SuppressWarnings({"DuplicateStringLiteralInspection"
         /* The generated code duplicates the code used in test that checks it. */,
         "ClassWithTooManyMethods" /* A lot of simple utility methods for code generation. */})
-public final class JsOutput {
+public final class CodeLines {
```
dmdashenkov: I think, what we have to do here now is to separate code construction from printing. I'd suggest that the `CodeLines` can produce an immutable object which is able to print the code into a string, file, stream, etc., depending on the plugin needs.

tools/proto-js-plugin/src/main/java/io/spine/js/generate/CodeLines.java
```diff
@@ -160,14 +198,14 @@ public void enterMethod(String methodName, String... methodArgs) {
...
+        increaseDepth();
     }
 
     /**
      * Exits method declaration.
      */
-    public void exitFunction() {
-        currentDepth--;
+    public void exitMethod() {
+        decreaseDepth();
```
dmdashenkov: There might be an improvement done to the depth processing. For example, we may define the depth concept as a class and incorporate it with `Indent` class instead of manually computing the number of spaces to print.

Also, is it necessary that `increaseDepth()` and `decreaseDepth()` method are `public`?

tools/proto-js-plugin/src/main/java/io/spine/js/generate/CodeLine.java
```diff
@@ -30,11 +34,21 @@
  * <p>For example, the code inside the {@code if} block is one unit deeper than the {@code if}
  * declaration itself.
  */
-final class CodeLine {
+public final class CodeLine {
```
dmdashenkov: Let's make this class abstract and create descendants for specific cases. Currently, these cases are processed in `CodeLined` class using `String + String` operator. What I'm suggesting are classes, such as `Comment` and `VariableDeclaration`.

tools/proto-js-plugin/src/main/java/io/spine/js/generate/CodeSnippet.java
```diff
@@ -0,0 +1,34 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.spine.js.generate;
+
+/**
+ * A snippet of the code.
+ */
+public interface CodeSnippet {
```
dmdashenkov:
```suggestion
public interface Snippet {
```

tools/proto-js-plugin/src/main/java/io/spine/js/generate/CodeSnippet.java
```diff
@@ -0,0 +1,34 @@
...
+
+    /**
+     * Obtains code lines representing this snippet.
+     *
+     * @return always returns a new {@link CodeLines}
+     */
+    CodeLines value();
```
dmdashenkov: After `CodeLines` is divided into read and write side, this method should return the read side, i.e. the immutable object capable of printing the code.

tools/proto-js-plugin/src/main/java/io/spine/js/generate/Statements.java
dmytro-grankin: It's fine, the same in the [documentation](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Map#%D0%92%D0%B7%D0%B0%D0%B8%D0%BC%D0%BE%D0%BE%D1%82%D0%BD%D0%BE%D1%88%D0%B5%D0%BD%D0%B8%D1%8F_%D1%81_%D0%BE%D0%B1%D1%8A%D0%B5%D0%BA%D1%82%D0%BE%D0%BC_Array).

dmytro-grankin:
@dmdashenkov PTAL.

tools/proto-js-plugin/src/main/java/io/spine/js/generate/CodeLines.java
```diff
@@ -86,65 +82,73 @@ public JsOutput() {
...
-    public void addLine(String codeLine) {
-        checkNotNull(codeLine);
-        CodeLine line = new CodeLine(codeLine, currentDepth);
-        codeLines.add(line);
+    public void append(CodeLines appended) {
+        checkArgument(indentation.equals(appended.indentation),
+                      "Cannot merge code parts with different indentation.");
+        int depthDifference = currentDepth.value() - appended.currentDepth.value();
+        for (IndentedLine appendedLine : appended.codeLines) {
+            IndentedLine adjusted = appendedLine.withIncreasedDepth(depthDifference);
```
dmdashenkov: What if the depth of `appended` is greater than the depth of `this`?

tools/proto-js-plugin/src/main/java/io/spine/js/generate/CodeLines.java
```diff
@@ -266,49 +270,47 @@ public void ifNotNullOrUndefined(String value) {
...
      */
     public void decreaseDepth() {
-        currentDepth--;
+        currentDepth = currentDepth.decremented();
+    }
+
+    /**
+     * Merges lines by addition of a comma to each line except the last one.
+     */
+    public static CodeLines commaSeparated(List<CodeLine> lines) {
```
dmdashenkov: Please use a `Joiner` instead.

tools/proto-js-plugin/src/main/java/io/spine/js/generate/RawLine.java
```diff
@@ -0,0 +1,72 @@
...
+import static java.lang.String.format;
+
+/**
+ * A line of a JavaScript code represented by a string.
+ */
+public class RawLine extends CodeLine {
+
+    private final String content;
+
+    RawLine(String content) {
```
dmdashenkov: This constructor must be private.

tools/proto-js-plugin/src/main/java/io/spine/js/generate/CodeLines.java
dmytro-grankin: As discussed, a `Joiner` doesn't suite here since the method works with lines, which are appended to `CodeLines`.

dmytro-grankin:
@armiol PTAL at the updated package structure.

base/src/main/java/io/spine/code/js/JsType.java
```diff
@@ -0,0 +1,86 @@
...
+        return type.url();
+    }
+
+    /**
+     * Obtains the reference to a static method of the type.
+     *
+     * @param methodName
+     *         the methodName of the method
+     */
+    public String staticMethod(String methodName) {
```
armiol: I would also create a concept for a method to connect it to its position and "depth" (or how we call it).

tools/proto-js-plugin/src/main/java/io/spine/js/generate/FileSetEnhancement.java
```diff
@@ -0,0 +1,86 @@
...
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * An enhancement of generated Protobuf sources.
+ *
+ * <p>The class operates on the provided set of Protobuf files
+ * and enhances their JavaScript counterparts by generating additional code.
+ */
+public abstract class FileSetEnhancement {
```
armiol: Let's discuss this name. I too don't understand the idea.

Judging on its descendants this one looks a "generation task"  that we perform, taking the file set as an argument.

tools/proto-js-plugin/src/main/java/io/spine/js/generate/IndentedLine.java
```diff
@@ -0,0 +1,125 @@
...
+
+/**
+ * The JS code line.
+ *
+ * <p>Consists of the code itself and the depth on which the code is located.
+ *
+ * <p>For example, the code inside the {@code if} block is one unit deeper than the {@code if}
+ * declaration itself.
+ */
+final class IndentedLine extends CodeLine {
```
armiol: There is now `CodeLine`, `RawLine` and `IndentedLine`. To me it's too much. Let's simplify.

It all looks to me as an `format` package, including the abstractions over indentation and depth.

tools/proto-js-plugin/src/main/java/io/spine/js/generate/FileSetEnhancement.java
armiol: See https://gist.github.com/armiol/f059b3c997f9053ed94bb27428f14ad5.

dmytro-grankin:
@armiol PTAL.

base/src/main/java/io/spine/code/js/MethodReference.java
```diff
@@ -0,0 +1,70 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.spine.code.js;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * A reference to a method.
```
armiol: To a method of what? Please update the description so that it's clear without analyzing the fields of this class.

tools/proto-js-plugin/src/main/java/io/spine/js/generate/GenerationTask.java
```diff
@@ -0,0 +1,79 @@
...
+        this.generatedRoot = checkNotNull(generatedRoot);
+    }
+
+    /**
+     * Generates code for the specified file set.
+     *
+     * @param fileSet
+     *         the Protobuf files to generate code for
+     */
+    public final void perform(FileSet fileSet) {
```
armiol: `perform` hardly reads with `fileSet`. It's either `performAt` or `performFor`.

tools/proto-js-plugin/src/main/java/io/spine/js/generate/GenerationTask.java
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * Checks if the task has any files to process.
+     *
+     * <p>Will return {@code false} either if there are no known types to process or the generated
+     * files for them cannot be found.
+     *
+     * @return {@code true} if there are files to process and {@code false} otherwise
+     */
+    private boolean shouldPerform(FileSet fileSet) {
```
armiol: Let's call it by what it really does, e.g. `hasFiles`.

tools/proto-js-plugin/src/main/java/io/spine/js/generate/output/snippet/Method.java
```diff
@@ -0,0 +1,132 @@
...
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.collect.Lists.newArrayList;
+import static io.spine.validate.Validate.checkNotEmptyOrBlank;
+import static java.lang.String.join;
+
+/**
+ * The declaration of a method in Javascript code.
+ */
+public class Method implements Snippet {
```
armiol: It's weird to me that `Method` has no relation to `MethodReference`. They seem to describe the same thing twice.

tools/proto-js-plugin/src/test/resources/build.gradle
```diff
@@ -76,5 +76,6 @@ protobuf {
 build.dependsOn compileProtoToJs
 
 dependencies {
+    protobuf "io.spine:spine-base:${spineVersion}"
```
armiol: This seems wrong, as it references the "previously" published artifact, not the current one.

You can refer to `base` protos via filesystem if you need those to be taken into account during their generation. If you are not completely sure, let's discuss.

tools/proto-js-plugin/src/main/java/io/spine/js/generate/parse/ExportStandardParsersMap.java
```diff
@@ -58,7 +62,7 @@
  * <p>The parsers may be used to parse JSON via their {@code parse(value)} method.
  */
 @SuppressWarnings("OverlyCoupledClass") // Dependencies for the listed Protobuf types.
-public final class ProtoParsersGenerator extends JsCodeGenerator {
+public final class ExportStandardParsersMap implements Snippet {
```
armiol: I don't understand calling this snipped ending with `Map`. Will we need to rename the class if it stops generating a `Map` but generates a JS class instead?

tools/proto-js-plugin/src/main/java/io/spine/js/generate/output/snippet/Method.java
```diff
@@ -0,0 +1,132 @@
...
+import static com.google.common.collect.Lists.newArrayList;
+import static java.lang.String.join;
+
+/**
+ * The declaration of a method in Javascript code.
+ */
+public class Method implements Snippet {
+
+    private final MethodReference reference;
+    private final List<String> arguments;
```
armiol: "Arguments" are real values, "parameters" are declarations. I.e.
```
// `a` and `b` are the names of parameters.
int sum(int a, int b) {...};

// `5` and `10` are arguments.
int result = sum(5, 10);
```

I am not sure what you meant, so please adjust the code with the account of parameter/argument difference.

base/src/main/java/io/spine/code/generate/Generation.java
```diff
@@ -18,7 +18,7 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-package io.spine.code;
+package io.spine.code.generate;
 
 /**
  * Constants for code generation.
```
armiol: As long as we repackage. 

Let's move `ModelCompilerAnnotation` to the top level, kill `Generation` (as it isn't what it claims to be) and rename `ModelCompilerAnnotation` to `GeneratedBySpine`. Ideally we need to relate this new class to the `Generated` annotation itself, e.g. return it from this class.

Issue #243, opened 2018-11-29
Allow `set_once` option for a field


alexander-yevsyukov:
We encourage having ID fields as a first field in a state of an entity. Presumably, such a field should never change after being set for the first time during creation of an entity.

It would be convenient to say something like:

```proto
   TaskId id = 1 [(required) = true, (set_once) = true];
```

Or, even assume such `require` and `set_once` attributes as automatically set to `true` if this is a first field of an entity state.

PR #257, opened 2018-12-19
`set_once` field option

serhii-lekariev:
This PR introduces a new custom Protobuf field validation option - `set_once`.

## Set once
Set once is an option that, when applied to a field, disallows changing it's state more than once.

Any attempt to do so results in a `ValidationException`.

Use cases of such options include identifiers, as once set, they need to stay the same in order to identify their respective entity.

So for a 
```proto
message User {
    UserId id = 1 [(set_once) = true];
}
```
the ID is unchangeable.

This PR partly addresses https://github.com/SpineEventEngine/base/issues/243, as the described option is added, however, it's not implicitly applied to a first field of every entity.


serhii-lekariev:
@dmdashenkov, PTAL.

base/src/main/java/io/spine/validate/AbstractValidatingBuilder.java
```diff
@@ -157,12 +157,21 @@ public void clear() {
...
+        FieldValue currentValue = currentFieldValue(descriptor, fieldContext);
         FieldValue valueToValidate = FieldValue.of(fieldValue, fieldContext);
-        FieldValidator<?> validator = create(valueToValidate);
+        FieldValueChange change = FieldValueChange.of(currentValue, valueToValidate);
+        FieldValidator<?> validator = create(change);
         List<ConstraintViolation> violations = validator.validate();
         checkViolations(violations);
     }
 
+    private FieldValue currentFieldValue(FieldDescriptor descriptor,
```
dmdashenkov: As discussed, please annotate the return type as `Nullable`.

base/src/main/java/io/spine/validate/FieldValidator.java
```diff
@@ -46,7 +47,14 @@
     private static final String ENTITY_ID_REPEATED_FIELD_MSG =
             "Entity ID must not be a repeated field.";
 
-    private final FieldValue value;
+    /**
+     * A value that the field that is being validated has had prior to validation.
+     *
+     * <p>Is {@code Nullable} because when a field is being validated for the first time,
+     * it can't have an existing value, and therefore considered {@code null}.
```
dmdashenkov:
```suggestion
     * it can't have an existing value, and, therefore, considered {@code null}.
```

base/src/main/java/io/spine/validate/DoubleFieldValidator.java
```diff
@@ -37,6 +37,16 @@
         super(fieldValue);
     }
 
+    /**
+     * Creates a new validator instance.
+     *
+     * @param valueChange
+     *         the change of the value of the field to validate
+     */
+    DoubleFieldValidator(FieldValueChange valueChange) {
```
dmdashenkov: Why some validators now have 2 constructors and others—just one?

base/src/main/java/io/spine/validate/FieldValidator.java
```diff
@@ -137,6 +170,24 @@ boolean fieldValueNotSet() {
...
+        if (setOnce && this.oldValue != null) {
+            String fieldName = this.oldValue.declaration()
+                                            .name()
+                                            .javaCase();
+            ConstraintViolation setOnceViolation = ConstraintViolation
+                    .newBuilder()
+                    .setMsgFormat("%s has (set_once) = true, can't change its value from %s to %s")
+                    .addParam(fieldName)
+                    .addParam(this.oldValue.asList()
+                                           .toString())
```
dmdashenkov: Please add a `FieldValue.toString()` method instead of always printing a list.

base/src/main/java/io/spine/validate/FieldValueChange.java
```diff
@@ -0,0 +1,88 @@
...
+ */
+public final class FieldValueChange {
+
+    /**
+     * A value that used to be the value of the field.
+     *
+     * <p>Nullable because when a field is set for the first time, old value has never existed
+     * and can be considered {@code null}.
+     */
+    private final @Nullable FieldValue oldValue;
```
dmdashenkov:
```suggestion
    private final @Nullable FieldValue previousValue;
```

base/src/main/proto/spine/options.proto
```diff
@@ -118,7 +118,22 @@ extend google.protobuf.FieldOptions {
 
     GoesOption goes = 73823;
 
-    // reserved 73824 to 73849 for future validation options.
+    // Indicates that a field can only be set once.
+    //
+    // A typical use-case would include a value of an ID, which doesn't change over the course of
+    // entities lifecycle.
```
dmdashenkov:
```
    // A typical use case would include a value of an ID, which doesn't change over the course of
    // the life of an entity.
```

base/src/main/proto/spine/options.proto
```diff
@@ -118,7 +118,22 @@ extend google.protobuf.FieldOptions {
...
+    //
+    // So for:
+    // message User {
+    //
+    //     UserId id = 1 [(set_once) = true];
+    // }
+    //
+    // The ID `id` doesn't change.
```
dmdashenkov:
```suggestion
    // Once set, the `id` field cannot be changed.
```

base/src/main/java/io/spine/validate/FieldValueChange.java
```diff
@@ -0,0 +1,88 @@
...
+        return new FieldValueChange(null, singleValue);
+    }
+
+    /** Returns a value that the field is going to have as a result of this change. */
+    FieldValue newValue() {
+        return newValue;
+    }
+
+    /** Returns a value that the field used to have before this change took place.*/
+    FieldValue oldValue() {
```
dmdashenkov: Please mark the return type as `Nullable` and document on the nullability.

base/src/main/java/io/spine/validate/FieldValueChange.java
```diff
@@ -0,0 +1,88 @@
...
+ */
+
+package io.spine.validate;
+
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+/**
+ * A change in a {@linkplain io.spine.validate.FieldValue value of a field}.
+ */
+public final class FieldValueChange {
```
dmdashenkov: Please check if we can make this type package-private. If not, it should be marked as `Internal`.

serhii-lekariev:
@dmdashenkov, PTAL again.

base/src/main/java/io/spine/validate/FieldValueChange.java
```diff
@@ -0,0 +1,89 @@
...
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.spine.validate;
+
+
```
dmdashenkov: This line is redundant.

base/src/main/proto/spine/options.proto
```diff
@@ -118,7 +118,22 @@ extend google.protobuf.FieldOptions {
 
     GoesOption goes = 73823;
 
-    // reserved 73824 to 73849 for future validation options.
+    // Indicates that a field can only be set once.
+    //
+    // A typical use-case would include a value of an ID, which doesn't change over the course of
+    // the life of an entity.
+    //
+    // So for:
```
dmdashenkov:
```suggestion
    // Example:
```

serhii-lekariev:
@dmdashenkov, PTAL again.

base/src/main/java/io/spine/validate/AbstractValidatingBuilder.java
```diff
@@ -224,6 +226,28 @@ private void validateResult(T message) throws ValidationException {
         checkViolations(violations);
     }
 
+    @SuppressWarnings("unused")
+        // Called by all actual validating builder subclasses.
+    protected final void validateSetOnce(FieldDescriptor descriptor) throws ValidationException {
+        boolean setOnce = descriptor.getOptions()
+                                    .getExtension(OptionsProto.setOnce);
```
dmdashenkov: Please use `Options` utility instead.

base/src/main/java/io/spine/validate/AbstractValidatingBuilder.java
```diff
@@ -224,6 +226,28 @@ private void validateResult(T message) throws ValidationException {
...
+        boolean valueAlreadySet = getMessageBuilder().hasField(descriptor);
```
dmdashenkov: Calling `hasField(...)` is not allowed for `repeated` and `map` fields. If this is intentional, please add a `checkArgument` statement.

base/src/main/java/io/spine/validate/AbstractValidatingBuilder.java
```diff
@@ -224,6 +226,28 @@ private void validateResult(T message) throws ValidationException {
...
+        if (setOnce && valueAlreadySet) {
+            throw setOnceViolation(descriptor);
+        }
+    }
+
+    private static ValidationException setOnceViolation(FieldDescriptor descriptor) {
```
dmdashenkov:
```suggestion
    private static ValidationException violatedSetOnce(FieldDescriptor descriptor) {
```

Otherwise, this method looks like a setter.

base/src/main/java/io/spine/validate/AbstractValidatingBuilder.java
```diff
@@ -224,6 +226,28 @@ private void validateResult(T message) throws ValidationException {
...
+        String fieldName = descriptor.getName();
+        ConstraintViolation setOnceViolation = ConstraintViolation
+                .newBuilder()
+                .setMsgFormat("Attempted to change a value of the field %s that has " +
```
dmdashenkov:
```suggestion
                .setMsgFormat("Attempted to change the value of the field %s, which has " +
```

base/src/main/java/io/spine/validate/AbstractValidatingBuilder.java
```diff
@@ -224,6 +226,28 @@ private void validateResult(T message) throws ValidationException {
...
+                              "(set_once) = true and is already set")
```
dmdashenkov:
```suggestion
                              "`(set_once) = true` and is already set.")
```

tools/model-compiler/src/main/java/io/spine/tools/compiler/validation/AbstractMethodGroup.java
```diff
@@ -119,4 +119,10 @@ static String validateStatement(String fieldValue, String fieldName) {
                                            FIELD_DESCRIPTOR_NAME, fieldValue, fieldName);
         return codeBlock.toString();
     }
+
+    /** Creates a statement that calls the {@code validateSetOnce} method. */
+    static String validateImmutable() {
```
dmdashenkov:
```suggestion
    static String validateSetOnce() {
```

Strictly speaking, this field is not immutable. It is not `MonotonicNonNull` either. It is a partial combination of two.

tools/smoke-tests/validation-tests/src/test/java/io.spine/validate/ValidatingBuilderTest.java
```diff
@@ -161,6 +164,66 @@ void dispense_with_unchecked_map_fields() {
         builder.build();
     }
 
+    @Test
+    @DisplayName("not allow to mutate a (set_once) marked string field via the setter method")
+    void testSetDoesNotAllowMutate() {
+        TaskVBuilder builder = TaskVBuilder
+                .newBuilder()
+                .setId(Identifier.newUuid());
+        assertThrows(ValidationException.class, () -> builder.setId(Identifier.newUuid()));
```
dmdashenkov: Please add a check and respective tests for the `mergeFrom(...)`  method.

serhii-lekariev:
@dmdashenkov, PTAL.

serhii-lekariev:
@dmdashenkov, PTAL again.

base/src/main/java/io/spine/validate/AbstractValidatingBuilder.java
```diff
@@ -233,21 +233,22 @@ private void validateResult(T message) throws ValidationException {
...
+            boolean setOnceNotRecommended = descriptor.isRepeated() || descriptor.isMapField();
+            if (setOnceNotRecommended) {
+                String containingTypeName = descriptor.getContainingType()
+                                                      .getName();
+                String fieldName = descriptor.getName();
+                log().warn("Error found in %s.%s. " +
+                                   "Repeated and map fields can't be marked as `(set_once) = true`",
+                           containingTypeName,
+                           fieldName);
+                return;
```
dmdashenkov: Please use `else` instead. That would make the flow more obvious.

tools/smoke-tests/validation-tests/src/test/proto/spine/test/validate/validating_builder_test.proto
```diff
@@ -120,7 +120,25 @@ message Attachment {
     spine.net.Url url = 3 [(required) = true, (valid) = true];
 }
 
+message TaskContents {
```
dmdashenkov: This is a `Book`, an `Article`, or a `Paper`, but not a `Task`, judging on the fields.

base/src/main/java/io/spine/validate/AbstractValidatingBuilder.java
```diff
@@ -42,7 +46,7 @@
  * Serves as an abstract base for all {@linkplain ValidatingBuilder validating builders}.
  */
 public abstract class AbstractValidatingBuilder<T extends Message, B extends Message.Builder>
-        implements ValidatingBuilder<T, B> {
+        implements ValidatingBuilder<T, B>, Logging {
```
dmdashenkov: Let's replace with a static method call.

A validating builder is a public Java API, which hardly needs to expose its `log` to the users.

Issue #197, opened 2018-09-21
Invalid `required` validation for number fields


mdrachuk:
Right now the required validation errors for number fields results in unclear `Value must be set.` error if the integer part of the number is `0`. This conflicts with cases when a `0` value is set. This behavior is not documented for the protobuf `required` option. 

Furthermore, only an integer part of double is considered when checking `double` and `float` values. Thus a required double field with `0.9` value will result in `Value must be set.` error.

Possibly, the `required` validation is semantically invalid for number fields. Can we consider using `min` and `max` validation instead of `required`, making `required` throw an error for `int32`, `int64`, `uint32`, `uint64`, `float` and `double` message fields?

dmdashenkov:
The `0` for integers pars seems OK. [Here](https://developers.google.com/protocol-buffers/docs/proto3#default) is the spec describing the default values of the message fields. However, `min`-`max` validation does sound more reasonable.

PR #266, opened 2018-12-21
Encourage to avoid using `(required)` field option for numeric and boolean types

serhii-lekariev:
This PR closes https://github.com/SpineEventEngine/base/issues/197.

Before, attempting to use `(required) = true` on all numeric types and the boolean type resulted in false-positive violations of `ValidatingException` - the numeric fields were compared against  `0` and boolean fields were compared against `false`, and if they were equal to those default values, the fields were considered unset.

This is semantically incorrect, because both `0` and `false`, while being a default, are valid values nonetheless. 

Now, when trying to validate `required` fields of the following numeric types:
- `int32`
- `int64`
- `float`
- `double`

or a `required` field of type `bool`, a warning is produced, telling that doing so is incorrect.


base/src/main/java/io/spine/validate/FieldValidator.java
```diff
@@ -169,14 +170,34 @@ boolean fieldValueNotSet() {
...
+    }
+
+    /**
+     * Checks whether a field can be {@code (required) = true}.
+     *
+     * Validators can define that for their respective types
+     *
+     * @return whether a field that is being validated can be {@code required}
+     */
+    protected abstract boolean requiredAllowed();
```
dmdashenkov: This method name is complicated. It uses 2 past form verbs. Please think of a better name.

base/src/main/java/io/spine/validate/FieldValidator.java
```diff
@@ -169,14 +170,34 @@ boolean fieldValueNotSet() {
...
+            log().warn("A field %s of type that can't be required was declared as" +
+                               "(required) = true. See required option documentation in " +
+                               "the options.proto file.", field().name());
+        }
+    }
+
+    /**
+     * Checks whether a field can be {@code (required) = true}.
+     *
+     * Validators can define that for their respective types
```
dmdashenkov: `<p>` is missing.

base/src/main/java/io/spine/validate/FieldValidator.java
```diff
@@ -169,14 +170,34 @@ boolean fieldValueNotSet() {
         return result;
     }
 
+    private void canBeRequired() {
```
dmdashenkov: The name does not reflect the action. I'd expect a method starting with `can` to return a `boolean`. Please rename.

base/src/main/java/io/spine/validate/FieldValidator.java
```diff
@@ -169,14 +170,34 @@ boolean fieldValueNotSet() {
...
+
     /**
      * Validates whether a logic enforced by the {@code (set_once)} option.
      *
      * <p>If a field that is being validated already has any value, and there's an attempt to change
      * it, a {@code ConstraintViolation} is generated.
      */
     @SuppressWarnings("ConstantConditions")
-        // `isFirstTimeSet` controls for `null`s.
+    // `isFirstTimeSet` controls for `null`s.
```
dmdashenkov: Please undo this change.

base/src/main/java/io/spine/validate/FieldValidator.java
```diff
@@ -252,6 +273,8 @@ private boolean hasCustomMissingMessage() {
      * <p>If the field is repeated, it must have at least one value set, and all its values
      * must be valid.
      *
+     * <p>If the field is of a numeric type, it's always considered set.
```
dmdashenkov: What about `bool`s?

base/src/main/java/io/spine/validate/StringFieldValidator.java
```diff
@@ -101,4 +101,9 @@ protected boolean isNotSet(String value) {
         boolean result = value.isEmpty();
         return result;
     }
+
+    @Override
+    protected boolean requiredAllowed() {
+        return true;
```
dmdashenkov: All the overloads of this method return a constant. Please consider passing the value in a `super(...)` ctor call.

base/src/main/java/io/spine/validate/NumberFieldValidator.java
```diff
@@ -112,11 +112,21 @@ protected void validateOwnRules() {
         }
     }
 
+    /**
+     * Returns {@code false}.
+     *
+     * @implNote Attempting to define whether a numeric field is set or not is semantically
+     * incorrect, because the way to define whether a value is set or not is to compare it against
+     * its default value, which for numeric fields is {@code 0}, which is sometimes a desired and
+     * valid value for a field.
```
dmdashenkov: This _one_ sentence is too long. Please split it up and simplify. Also, I'm not sure if `<p>` paragraphs are supported in `@implNote`.

base/src/main/java/io/spine/validate/FieldValidator.java
```diff
@@ -138,6 +143,18 @@ boolean fieldValueNotSet() {
         return result;
     }
 
+    private void checkCanBeRequired() {
+        boolean fieldIsRequired = isRequiredField();
+        if (!canBeRequired && fieldIsRequired) {
+            String messageFormat = "Fields of type %s can't be declared as `(required)`.";
```
dmdashenkov:
```suggestion
            String messageFormat = "Fields of type %s shouldn't be declared as `(required)`.";
```

base/src/test/java/io/spine/validate/FloatFieldValidatorTest.java
```diff
@@ -28,6 +30,17 @@
     private static final float HALF = 0.5F;
 
     FloatFieldValidatorTest() {
-        super(HALF, -HALF, new FloatFieldValidator(FieldValue.of(HALF, fieldContext)));
+        super(HALF, -HALF, new FloatFieldValidator(FieldValue.of(HALF, fieldContext)),
+              requiredFieldValidator());
```
dmdashenkov: Please allocate each next param on a new line.

base/src/main/java/io/spine/validate/FieldValidator.java
```diff
@@ -138,6 +143,18 @@ boolean fieldValueNotSet() {
...
+            String typeName = this.field()
+                                  .descriptor()
+                                  .getType()
+                                  .name();
+            log().warn(messageFormat, typeName);
```
dmdashenkov: Please use `Logging._warn()` and similar methods instead.

base/src/main/java/io/spine/validate/FieldValidator.java
```diff
@@ -138,6 +143,18 @@ boolean fieldValueNotSet() {
...
     }
 
+    private void checkCanBeRequired() {
+        boolean fieldIsRequired = isRequiredField();
+        if (!canBeRequired && fieldIsRequired) {
+            String messageFormat = "Fields of type %s can't be declared as `(required)`.";
+            String typeName = this.field()
+                                  .descriptor()
+                                  .getType()
+                                  .name();
```
dmdashenkov: This is ugly. Please use `FieldDeclaration` methods. If there is no such method, please add one.

serhii-lekariev:
@dmdashenkov, PTAL again.

base/src/main/java/io/spine/validate/BooleanFieldValidator.java
```diff
@@ -47,8 +47,6 @@ protected boolean isNotSet(Boolean value) {
 
     @Override
     protected void validateOwnRules() {
-        if (isRequiredField()) {
-            log().warn("'required' option not allowed for boolean field");
-        }
+        // NOP
```
dmdashenkov: Please document the reason.

PR #267, opened 2018-12-21
Factory methods for UUID-based `Message` identifiers

dmitrykuzmin:
This PR resolves issue https://github.com/SpineEventEngine/core-java/issues/778.

The `MarkerInterfaceGenerator` will search for messages like:

```
message ProjectId {
    string uuid = 1;
}
```
i.e. with a single `string` value named "uuid" and mark them as `UniqueId`.

The `UniqueId` interface provides convenient `default` methods for messages creation: `generate` and `of(String)` to generate an ID from random string or given value respectively.

Also the `MarkerInterfaceGenerator` was extracted into a separate package `marker` under the `io.spine.tools.protoc`.


dmitrykuzmin:
@armiol PTAL.

tools/protoc-plugin/src/main/java/io/spine/tools/protoc/MarkerInterfaceParameter.java
```diff
@@ -0,0 +1,40 @@
...
+ * The generic parameter of the {@link MarkerInterface}.
+ *
+ * <p>Parameter value is presented as a {@code String} for usage in the generated code.
+ */
+interface MarkerInterfaceParameter {
+
+    /**
+     * Obtains a parameter value based on who is the marker interface implementor.
+     *
+     * @param implementor
```
armiol: Let's call it `descendant` or `target` (i.e. a target to mark).

tools/protoc-plugin/src/main/java/io/spine/tools/protoc/MessageItself.java
```diff
@@ -0,0 +1,37 @@
...
+
+import com.google.protobuf.DescriptorProtos.DescriptorProto;
+
+/**
+ * The marker interface parameter whose value is the implementing {@code Message} itself.
+ *
+ * <p>For example, for the {@code ProjectId} class, the value of the parameter will be:
+ * {@code ProjectId}.
+ */
+final class MessageItself implements MarkerInterfaceParameter {
```
armiol: Let's discuss. 

To me a class such as `io.spine.tools.protoc.MessageItself` is questionable.

Let's consider splitting marker interface-specific routines into their own package.

tools/protoc-plugin/src/test/proto/spine/tools/protoc/incorrect_ids.proto
```diff
@@ -0,0 +1,44 @@
...
+import "spine/options.proto";
+
+option (type_url_prefix) = "type.spine.io";
+option java_package = "io.spine.tools.protoc";
+option java_outer_classname = "IncorrectIds";
+option java_multiple_files = true;
+
+import "google/protobuf/timestamp.proto";
+
+message NotId {
```
armiol: Same as above. It _is_ an ID. Or at least there can be an identifier like this.

tools/protoc-plugin/src/test/proto/spine/tools/protoc/incorrect_ids.proto
```diff
@@ -0,0 +1,44 @@
...
+//
+syntax = "proto3";
+
+package spine.tools.protoc;
+
+import "spine/options.proto";
+
+option (type_url_prefix) = "type.spine.io";
+option java_package = "io.spine.tools.protoc";
+option java_outer_classname = "IncorrectIds";
```
armiol: Can we have the naming of the outer class and the file somehow related to `UniqueId` marker interface?

base/src/main/java/io/spine/base/UniqueId.java
```diff
@@ -0,0 +1,84 @@
...
+     * Generates a new identifier instance using a random {@code String}.
+     */
+    default I generate() {
+        Class<I> thisClass = (Class<I>) this.getClass();
+        UuidMessage<I> uuidMessage = UuidMessage.of(thisClass);
+        return uuidMessage.generate();
+    }
+
+    /**
+     * Generates a new identifier instance from the passed value.
```
alexander-yevsyukov: It does not generate. It creates an instance by value.

base/src/main/java/io/spine/base/UniqueId.java
```diff
@@ -0,0 +1,84 @@
...
+ *     }
+ * </pre>
+ *
+ * @param <I>
+ *         the type of the message
+ */
+@SuppressWarnings({"unchecked" /* Class cast correctness guaranteed by the interface contract. */,
+        "InterfaceNeverImplemented" /* Used by the Protobuf Compiler plugin. */})
+@Immutable
+public interface UniqueId<I extends Message> extends SerializableMessage {
```
alexander-yevsyukov: Why do we want `Unique` in the name? What's so unique about this name?

base/src/main/java/io/spine/base/UniqueId.java
```diff
@@ -0,0 +1,84 @@
...
+        return uuidMessage.generate();
+    }
+
+    /**
+     * Generates a new identifier instance from the passed value.
+     */
+    default I of(String value) {
+        Class<I> thisClass = (Class<I>) this.getClass();
+        UuidMessage<I> uuidMessage = UuidMessage.of(thisClass);
+        return uuidMessage.create(value);
```
alexander-yevsyukov: What happens if the passed value is not a UUID-based string?

alexander-yevsyukov:
How about renaming `UuidMessage` into `UuidFactory` and having `UuidValue` as an interface (which is now named `UniqueId`)?

dmitrykuzmin:
@armiol PTAL.

dmitrykuzmin:
@armiol PTAL.

PR #268, opened 2018-12-24
Filter out rejections when generating validating builders

alexander-yevsyukov:
This PR prevents generation of Validating Builders for rejection messages.

Validating Builders are not required for rejection messages because they are produced by objects derived from `ThrowableMessage`.


alexander-yevsyukov:
@mdrachuk, PTAL.

PR #269, opened 2018-12-26
Fix annotator field lookup

dmdashenkov:
Previously, if a message had two fields with similar names, there was a risk that the annotator plugin would annotate both fields (with `@Internal`, `@SPI`, etc.) instead of only one field.

For example:
```proto
message Container {
    string spi_field = 1 [(SPI) = true];
    string not_spi_field = 2;
}
```
When annotating methods of `Container` class, both `spi_field` and `not_spi_field` accessors would be annotated with `@SPI` (since `"not_spi_field".contains("spi_field")`).

In this PR we fix this issue by specifying the method names explicitly.


dmdashenkov:
@armiol, PTAL.

dmdashenkov:
@armiol, PTAL again.

tools/model-compiler/src/main/java/io/spine/tools/compiler/field/type/RepeatedFieldType.java
```diff
@@ -17,46 +17,72 @@
...
+import static io.spine.tools.compiler.field.AccessorTemplates.listGetter;
+import static io.spine.tools.compiler.field.AccessorTemplates.setter;
+
 /**
  * Represents repeated {@linkplain FieldType field type}.
  */
 public final class RepeatedFieldType implements FieldType {
 
-    private static final String SETTER_PREFIX = "addAll";
+    private static final ImmutableSet<AccessorTemplate> GENERATED_ACCESSORS = ImmutableSet.of(
```
armiol: Let's move `ImmutableSet.of(` to the next line (it's here and in some other files in this changeset).

tools/model-compiler/src/main/java/io/spine/tools/compiler/field/type/RepeatedFieldType.java
```diff
@@ -17,46 +17,72 @@
...
     }
 
+    @Override
+    public ImmutableSet<AccessorTemplate> generatedAccessorTemplates() {
+        return GENERATED_ACCESSORS;
+    }
+
     /**
      * Returns "addAll" setter prefix,
      * used to initialize a repeated field using with a call to Protobuf message builder.
```
armiol: Let's format this comment evenly in two lines.

PR #270, opened 2018-12-26
Simplify dependency for generation of vbuilders for `base`

alexander-yevsyukov:
This PR migrates generation of Validating Builders of types defined in the `base`  from dependency on locally deployed JAR to dependencies to interim artifacts. 

The primary purpose of this change is to avoid double dependency on both source code of the `base` module, and classes from the locally deployed artifact.

The notable fix in this PR is restored filtering of proto types by the module which declares them.


alexander-yevsyukov:
@armiol, PTAL.

PR #272, opened 2018-12-27
Validating Builders for nested nessages

alexander-yevsyukov:
This PR enables generation of validating builders for nested message types.

A validating builder class would be named after the pattern: `<EnclosingType><NestedType>VBuilder`
So, for example for `Uri.Protocol`, the Validating Builder class would be `UriProtocolVBuilder`. 

If `java_multiple_files` is set to false, the outer class name would be used in the prefix of the name:
`<OuterType><TopLevel><Nested><YetMoreNested>VBuilder`.

Notable API changes:
  * `MessageDocumentation` was eliminated in favor of having comment-related methods under `MessageType`.
  * `LocationPath` now consolidates the logic of working with paths.
  * Aspects of a `GradleProject` were extracted as separate classes.


alexander-yevsyukov:
@armiol, @dmdashenkov, PTAL.

alexander-yevsyukov:
@armiol, once this finished, [this PR](https://github.com/SpineEventEngine/core-java/pull/931) in `core-java` should fix the broken build because of absence of `version.gradle`.

PR #274, opened 2019-01-02
Completely rework parsing of generated Protobufs and improve the API of Proto JS Plugin

dmytro-grankin:
This PR changes the way how generated Protobufs are parsed and improves the API of the Proto JS Plugin.

### Changes in the parsing API

The code generated for the parsing of non-standard messages was changed. Previously, a `fromObject` and a `fromJson` method was generated for messages. Now, the methods are not generated. Instead `ProtoJsPlugin` generates a parser per message type.

The parsers are still not generated for standard types (`Any`, `Timestamp` etc) since they are parsed differently. The parsers for these types are going to be introduced in the Spine Web.

### Sharing responsibilities with Spine Web

Some responsibilities for working with Protobuf types were moved to the Spine Web. So now the generated code depends on `spine-web` module. In particular, an abstract base for parsers (`ObjectParser`) is provided by the Spine Web.

The parsers may use other parsers under the hood. But, these parsers aren't accessed directly. They are accessed via `TypeParsers` — another class, which is going to be introduced in the Spine Web. The `TypeParsers` contains all the parsers and provides the API to access them.

There were two options on how to register the parsers in `TypeParsers`. Either by doing by it in generated code or by the Spine Web directly. It was decided to expose all the parsers for a module and then register them in the Spine Web, not in the generated code. This option is more reliable.

The parsers are exposed in the generated `index.js` file. The index file also contains the types to be registered in `KnownTypes` in the same manner as the parsers.

The index file is going to be used by the Spine Web as well as by its users.

### Resolving imports of Spine Web

There was one issue with the dependency on Spine Web in generated code. It appeared in the Spine generated files since Javascript cannot resolve `spine-web` module. 

To deal with the issue it was decided to post-process imports and replace them with relative paths when possible. The work is done by a newly introduced `ResolveImports` task. The task checks if imported files are present on the file system (assuming particular module structure) and, if they are present, replaces the imports by relative paths to these files.

### The Javascript plugin configuration (#285)

Previously, the plugin required to place generated Protobufs and the descriptor set file under specific directories. Now you can specify the paths via the plugin extension:

```gradle
protoJs {
    mainGenProtoDir = "$projectDir/customProtoFolder/main/js"
    mainDescriptorSetPath = "$projectDir/custom/main/known_types.desc"
    // The same for test files.
}
```

The one point to remember is that Protobuf compiler should be configured respectively.

Also, the default directories for generated Protobufs are changed (from `proto/main/js` to `generated/main/js`). So, you need to update the directory path either in the plugin extension or in Protobuf compiler settings.


dmytro-grankin:
@armiol PTAL.

tools/proto-js-plugin/src/main/java/io/spine/js/generate/parse/FromJsonMethod.java
```diff
@@ -104,42 +89,43 @@ public static FromJsonMethod createFor(Descriptor message) {
...
+                .withParameters(parameterName)
+                .appendToBody(newParser)
+                .appendToBody(Return.value(callParser))
+                .build();
     }
 
     /**
      * Obtains {@code fromJson} method for the specified message.
+     *
+     * The {@code fromJson} parses the JSON string via the {@code JSON.parse}
```
armiol: `<p>`

base/src/main/java/io/spine/code/js/TypeName.java
```diff
@@ -45,10 +41,23 @@ private TypeName(String value) {
...
         checkNotNull(descriptor);
         String typeName = descriptor.getFullName();
         String nameWithPrefix = PREFIX + typeName;
         return new TypeName(nameWithPrefix);
     }
+
+    /**
+     * Obtains the type name with the specified value.
+     */
+    public static TypeName of(String value) {
```
armiol: This looks harmful.

tools/proto-js-plugin/src/main/java/io/spine/js/generate/parse/Parser.java
```diff
@@ -0,0 +1,177 @@
...
+@SuppressWarnings("DuplicateStringLiteralInspection" /* Used in a different context. */)
+public class Parser implements Snippet {
+
+    /**
+     * The name of the {@code fromObject} method return value.
+     *
+     * <p>This value represents the generated JS message whose fields are parsed and set from the
+     * JS object.
+     */
+    public static final String MESSAGE = "msg";
```
armiol: Not sure why the constants are `public` here. Let's discuss.

tools/proto-js-plugin/src/main/resources/io/spine/tools/protojs/knowntypes/object_parser
```diff
@@ -0,0 +1,51 @@
...
+     * @abstract
+     * @param {!Object} object the object representing a Protobuf message
+     * @return {!Message} the parsed Protobuf message
+     */
+    fromObject(object) {
+        throw new Error('The method is abstract and should be implemented by a subclass');
+    }
+}
+
+module.exports = ObjectParser;
```
armiol: EOL is broken, as reported by GitHub.

Also, having a file like this with no extension is suspicious. If this is a template, let's name it so.

tools/proto-js-plugin/src/main/java/io/spine/js/generate/parse/Parser.java
```diff
@@ -0,0 +1,177 @@
...
+ * standard types are manually created and require no code generation.
+ *
+ * <p>Code provided by the class is in {@code ES5} standard
+ * since Protobuf compiler generates Javascript in {@code ES5}.
+ *
+ * <p>The class is effectively {@code final} and is left non-{@code final} only for testing
+ * purposes.
+ */
+@SuppressWarnings("DuplicateStringLiteralInspection" /* Used in a different context. */)
+public class Parser implements Snippet {
```
armiol: I am not sure `Parser` is a `Snippet`.

dmytro-grankin:
@armiol PTAL.

tools/proto-js-plugin/src/main/java/io/spine/js/generate/field/FieldToParse.java
```diff
@@ -0,0 +1,68 @@
...
+package io.spine.js.generate.field;
+
+import com.google.protobuf.Descriptors.FieldDescriptor;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * A Protobuf field, which is parsed from a plain Javascript object to a Protobuf-specific type.
+ */
+public final class FieldToParse {
```
armiol: Let's discuss this one.

kamaek:
@armiol PTAL.

dmytro-grankin:
@armiol PTAL.

tools/proto-js-plugin/src/test/java/io/spine/js/generate/output/snippet/ImportTest.java
```diff
@@ -0,0 +1,74 @@
...
+package io.spine.js.generate.output.snippet;
+
+import com.google.protobuf.Any;
+import io.spine.code.js.FileName;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+
+import static com.google.common.truth.Truth.assertThat;
+
+@DisplayName("Import")
```
armiol: It's something with the code generation. So to me a sentence composed from a class-level `@DisplayName` and a test-level `@DisplayName` should read as follows:

`<Somebody? such as code generator> should generate import for`
       `a file relative to a current directory`.

dmytro-grankin:
@armiol PTAL.

dmytro-grankin:
@armiol PTAL.

PR #275, opened 2019-01-02
Set once by default on first field of all entity states

serhii-lekariev:
This PR closes https://github.com/SpineEventEngine/base/issues/243, while building on top of https://github.com/SpineEventEngine/base/pull/257 PR.

From now on, the first field of every entity state message is `(set_once) = true` by default, meaning that its value cannot be changed.

If you still want to be able to change the value of the first field, you have to declare it as `(set_once = false`) explicitly.


serhii-lekariev:
@dmdashenkov, PTAL.

base/src/main/java/io/spine/validate/AbstractValidatingBuilder.java
```diff
@@ -229,29 +231,55 @@ private void validateResult(T message) throws ValidationException {
...
-                        fieldName);
-        } else {
-            boolean valueAlreadySet = getMessageBuilder().hasField(field);
-            if (valueAlreadySet) {
-                throw violatedSetOnce(field);
+        Optional<Boolean> setOnceDeclaration = Options.option(field, OptionsProto.setOnce);
+        FieldDeclaration fieldDeclaration = new FieldDeclaration(field);
+        boolean setOnceValue = setOnceDeclaration.orElse(false);
+        boolean requiredByDefault = fieldDeclaration.isEntityId() && !setOnceDeclaration.isPresent();
+        if (setOnceValue || requiredByDefault) {
```
dmdashenkov: Please split this method up. It takes some time to understand what's going on.

tools/smoke-tests/validation-tests/src/test/java/io.spine/validate/ValidatingBuilderTest.java
```diff
@@ -175,91 +180,55 @@ void dispense_with_unchecked_map_fields() {
...
+        );
+        Stream<Function<TaskVBuilder, ?>> messageFieldMutations = Stream.of(
+                taskVBuilder -> taskVBuilder.setAssignee(Member.getDefaultInstance()),
+                taskVBuilder -> taskVBuilder.mergeFrom(sampleTask()),
+                TaskVBuilder::clearAssignee
+        );
+
+        testFields(taskVBuilder -> taskVBuilder.setId(newUuid()), stringFieldMutations);
+        testFields(taskVBuilder -> taskVBuilder.setLabel(IMPORTANT), enumFieldMutations);
+        testFields(taskVBuilder -> taskVBuilder.setAssignee(member()), messageFieldMutations);
```
dmdashenkov: Please split this test case into many. If we perform multiple similar checks, then there are multiple reasons for the test to fail.

serhii-lekariev:
@dmdashenkov, PTAL again.

base/src/main/java/io/spine/validate/AbstractValidatingBuilder.java
```diff
@@ -229,33 +231,71 @@ private void validateResult(T message) throws ValidationException {
...
+                boolean valueAlreadySet = getMessageBuilder().hasField(field);
+                if (valueAlreadySet) {
+                    throw violatedSetOnce(field);
+                }
             }
         }
     }
 
+    private static boolean isSetOnce(FieldDescriptor field) {
+        Optional<Boolean> setOnceDeclaration = Options.option(field, OptionsProto.setOnce);
```
dmdashenkov: Please use static imports.

PR #279, opened 2019-01-04
Better validation diagnostics

alexander-yevsyukov:
This PR simplifies code related to validation constraints violations.


alexander-yevsyukov:
@armiol, PTAL.

base/src/main/java/io/spine/validate/diags/ViolationText.java
```diff
@@ -0,0 +1,89 @@
...
+        checkNotNull(violation);
+        return new ViolationText(violation);
+    }
+
+    /**
+     * Creates text with diagnostics for the passed violations, starting each of them from
+     * a new line.
+     */
+    public static String ofAll(Iterable<ConstraintViolation> violations) {
+        StringBuilder builder = new StringBuilder(100);
```
armiol: If we use a Stream API, it's going to be shorter:

```java

String result = StreamSupport.stream(violations.spliterator(), false)
                             .map(ViolationText::of)
                             .map(ViolationText::toString)
                             .collect(Collectors.joining(lineSeparator()));
return result;
```

armiol: @alexander-yevsyukov FYI, I have updated the codebase as suggested.

PR #280, opened 2019-01-04
Customize generated code annotations

dmdashenkov:
### Code generation annotations

In this PR we enable customization for annotations used for code generation.

If a Protobuf definition uses one of the API options (see `options.proto`), the resulting Java code is post-processed by adding respective API annotations.
Before this PR, the annotations were hard-coded: 
 - `io.spine.annotation.Beta`;
 - `io.spine.annotation.Experimental`;
 - `io.spine.annotation.Internal`;
 - `io.spine.annotation.SPI`.

Starting from now, the annotation types may be specified with a Gradle extension:
```gradle
modelCompiler {
    generateAnnotations {
        internal = 'foo.bar.MyInternal'
        beta = 'com.google.common.annotations.Beta'
        // ...
    }
}
```

### Other significant changes:
 - `(SPI)` field option is deleted, since there are no obvious cases for such an option;
 - `EnumAnnotator` class is deleted, since it was never used;
 - new smoke tests module `annotator-tests` is added in order to verify the annotator plugin behaviour in cases when custom annotations are specified.


dmdashenkov:
@armiol, PTAL.

tools/smoke-tests/annotator-tests/src/test/java/io/spine/gradle/compiler/package-info.java
```diff
@@ -0,0 +1,27 @@
...
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+@CheckReturnValue
+@ParametersAreNonnullByDefault
+package io.spine.gradle.compiler;
```
armiol: Please document the package purpose. I understand this is a test-source package, but still, let's have at least a couple of words.

PR #282, opened 2019-01-08
Bump gradle wrapper version to `5.1.1`.

serhii-lekariev:
This PR updates the version of Gradle wrapper to `5.1`.

Before, `PreprocessorConfigurer` used to define annotation processor lookup logic by adding a `-processorpath` compiler argument. This way of setting up annotation processors is no longer supported, and the annotation processors are picked up by Gradle without any explicit logic.

It also adds a `settings.gradle` file to `base-validating-builders`, since nested builds without `settings.gradle` are no longer supported.


tools/smoke-tests/build.gradle
```diff
@@ -119,36 +119,4 @@ subprojects {
...
-                        .flatMap { it.stream() }
-                        .filter { it.name.endsWith('.jar') }
-                        .flatMap { zipTree(it).files.stream() }
-                        .filter { it.path.contains('io/spine') }
-                        .toArray()
-        )
-    }
-}
-
-check.dependsOn jacocoRootReport
```
serhii-lekariev: This script was replaced with the `config/gradle/jacoco.gradle`, but it doesn't do the same thing. 
I suspect that the drop in coverage is a result of this change, and, perhaps, a separate `smoke-test-jacoco.gradle` file should be created containing exactly the removed script.

serhii-lekariev:
@dmdashenkov, PTAL.

PR #284, opened 2019-01-09
Internal by pattern

dmdashenkov:
In this PR we introduce a new feature of the Model Compiler plugin—pattern-wise annotation.

For now, it is only possible to mark classes matching a specific pattern or set of patterns with the `@Internal` annotation (or the annotation specified via `modelCompiler.generateAnnotations.internal`).

We're also planning to add a similar syntax for annotating methods named in a specific way in future releases.

### Example

In the Gradle config, specify the patterns defining the internal classes:
```gradle
modelCompiler {
    internalClassPatterns += ['.*One.*', '.*Three.*']
}
```
Protobuf sources:
```proto
message One {
}

message Two {
}

message FiftyThree {
}
```
As the result, the message classes `One` and `FiftyThree` will be `@Internal`, as well as their `Builder` and `MessageOrBuilder` helper types.

#### Nesting

It is an important feature of class name pattern matching that the fully qualified canonical name of the class is always used for matching. 

Keep this in mind when designing class name patterns for messages with nested types. For example, if the message `One` from the previous example had a nested message `Half`
```proto
message One {

    message Half {
    }
}
```
then the classes generated for `Half` message would also be `@Internal`.

#### Suggested applications

One of the most useful applications for this feature is hiding the supporting generated types, such as
`MessageOrBuilder` subtype interfaces and outer classes (ending with the `...Proto` postfix):
```gradle
modelCompiler {
    internalClassPatterns += ['.+OrBuilder', '.+Proto']
}
```
This way, these types may be excluded from Javadoc, etc.


dmdashenkov:
@armiol, PTAL.

dmdashenkov:
@alexander-yevsyukov, PTAL.

PR #287, opened 2019-01-11
Mark selected methods internal

dmdashenkov:
In this PR we introduce yet another feature of the Protobuf annotator plugin—internal method by-name annotation.

In order to use this feature, configure the method names via Gradle:
```gradle
modelCompiler {
    internalMethodNames += ['parseFrom', 'mergeFrom', 'toString']
}
```

In the example above, all the methods with names `parseFrom`, `mergeFrom`, or `toString` in generated code will be annotated with the `@Internal` annotation (or the annotation configured in `modelCompiler.generateAnnotations.internal`).

Note that, unlike `internalClassPatterns`, `internalMethodNames` expects the values to be full valid names of Java methods. A value like `set.*` will result in no annotation added to any of the generated sources.

### Performance considerations

In order to annotate methods by names, the plugin parses __all the generated code__. Please consider using selective annotation (based on Protobuf options) where possible.

### Recommended use cases

When it is required to mark part of the generated class API which does not depend on the Protobuf sources directly, the users should use this feature. For example, `parseFrom`, `parseDelimitedFrom`, and similar methods are generated regardless the Protobuf sources, thus may only be marked as `@Internal` via the naming convention lookup.


dmdashenkov:
@armiol, PTAL.

base/src/main/java/io/spine/code/java/NestedClassName.java
```diff
@@ -43,4 +51,25 @@ static NestedClassName create(ClassName className) {
...
+     *
+     * <p>For example, if this class name is {@code "Container.Job.Builder"}, then the resulting
+     * list would be {@code ["Container", "Job", "Builder"]}.
+     *
+     * <p>If this class name is just a {@link SimpleClassName}, then the only entry of the resulting
+     * list is that simple name.
+     *
+     * @return this name split into simple class names
+     */
+    public ImmutableList<SimpleClassName> asPath() {
```
armiol: To me this is more like `split`. And `toPath()`should return the `Path` to declaring file.

Issue #127, opened 2018-06-25
Split `Options` into separate classes


dmdashenkov:
Currently, `Options` utility is a dump of all the custom option extractor methods. It should be, at least, split into separate classes. Another possible approach is to wrap the Protobuf descriptors and extract custom options with instance methods.

alexander-yevsyukov:
@armiol, it would be good if @serhii-lekariev finish assignments from @dmdashenkov, _before_ getting new ones.

PR #291, opened 2019-01-14
Refactor option validation

serhii-lekariev:
Before, when using Spine Protobuf options, e.g. for validation purposes, all of the option values were extracted by-hand in their respective `FieldValidator` implementation, e.g. to validate the `Pattern` option, one had to extract the value of the `PatternOption` in the `StringFieldValidator`, and then define the pattern-matching logic,

Instead, two new hierarchies were introduced: `Option` and `Constraint` interfaces allow to define validation logic in terms of constraints applied to fields, so now, to validate the `Pattern` field option, `StringFieldValidator` has only to create an instance of the `Pattern` validating option, call the `getConstraint()` method, that every `FieldValidatingOption` (which is an abstract `Option` interface implementation) has, and call the `check()`method.

Adding a new validating option is now as easy as creating a `Constraint` implementation and a `FieldValidatingOption` subclass.

This PR also adds a new repeated field validating option - `on_duplcate`, that imposes a `Set`-like behavior on a repeated field, generating `ConstraintViolation` if a non-distinct value has been found.


base/src/main/java/io/spine/validate/RequiredField.java
```diff
@@ -0,0 +1,179 @@
...
+        for (Descriptors.FieldDescriptor optionDescriptor : options.keySet()) {
+            if (OPTION_REQUIRED_FIELD.equals(optionDescriptor.getName())) {
+                String expression = (String) options.get(optionDescriptor);
+                return expression;
+            }
+        }
+        return "";
+    }
+
+    private static class RequiredFieldOptionData {
```
dmdashenkov: Please get rid of the unnecessary `Data` suffix. If a name not specific enough, re-consider the responsibility of the class.

base/src/main/proto/spine/options.proto
```diff
@@ -754,6 +766,23 @@ message GoesOption {
     string msg_format = 2;
 }
 
+// Defines options of a repeated field that defines a logic
+message OnDuplicate {
```
dmdashenkov: `OnDuplicate` may itself be an enum. Then we won't be forced to write the field name when using the option.

tools/smoke-tests/validation-tests/src/test/proto/spine/test/validate/validating_builder_test.proto
```diff
@@ -143,6 +143,21 @@ message Essay {
     map<string, int32> table_of_contents = 2 [(set_once) = true];
 }
 
+message Blizzard {
+    repeated Snowflake snowflake = 1 [(on_duplicate).duplicatePolicy = ERROR];
+}
+
+message Constitution {
+    string amendments = 1 [(on_duplicate).duplicatePolicy = ERROR];
```
dmdashenkov: I don't see a test for `IGNORE` behaviour. Please add one.

base/src/main/java/io/spine/validate/EnumFieldValidator.java
```diff
@@ -34,7 +37,7 @@
      *         the value to validate
      */
     EnumFieldValidator(FieldValue fieldValue) {
-        super(fieldValue, false, true);
+        super(fieldValue, false, true, ImmutableSet.of());
```
dmdashenkov: @serhii-lekariev, please add a simple `super` constructor, so you don't have to repeat `false, true, ImmutableSet.of()` every time. The existing one should then only be used for customization.

serhii-lekariev:
@dmdashenkov, PTAL at the intermediate result.

serhii-lekariev:
@dmdashenkov. PTAL.

base/src/main/java/io/spine/validate/ByteStringFieldValidator.java
```diff
@@ -44,14 +47,13 @@ protected boolean isNotSet(ByteString value) {
     }
 
     /**
-     * {@inheritDoc}
+     * @inheritDoc
      *
-     * <p>Performs no action since no special options are declared for the {@link ByteString}
-     * validation.
+     * <p>All of the additional validation logic is passed as the third
+     * {@linkplain FieldValidator(FieldValue, boolean, Set )} super constructor} parameter.
```
dmdashenkov: This line is the reason for the "unused" import reported by Codacy. Please do not add imports just the Javadoc.

base/src/main/java/io/spine/validate/DigitsConstraint.java
```diff
@@ -0,0 +1,96 @@
...
+                    (fractionDigitsCount > fractionDigitsMax);
+            if (isInvalid) {
+                ConstraintViolation digits = digits(fieldValue, digitsOption, val);
+                return ImmutableList.of(digits);
+            }
+        }
+        return ImmutableList.of();
+    }
+
+    private ConstraintViolation digits(FieldValue<V> fieldValue, DigitsOption digitsOption,
```
dmdashenkov: This method name is strange.

base/src/main/java/io/spine/validate/DistinctConstraint.java
```diff
@@ -0,0 +1,69 @@
...
+ */
+public class DistinctConstraint<T> implements Constraint<FieldValue<T>> {
+
+    @Override
+    public List<ConstraintViolation> check(FieldValue<T> fieldValue) {
+        ImmutableList<T> values = fieldValue.asList();
+        Set<T> duplicates = findDuplicates(values);
+        List<ConstraintViolation> violations = duplicates.stream()
+                                                         .map(duplicate -> distinctViolated(
+                                                                 fieldValue, duplicate))
```
dmdashenkov: Please fix the alignment.

base/src/main/java/io/spine/validate/DistinctConstraint.java
```diff
@@ -0,0 +1,69 @@
...
+                                                         .collect(toList());
+        return violations;
+    }
+
+    private ConstraintViolation distinctViolated(FieldValue<T> value, T duplicate) {
+        return ConstraintViolation.getDefaultInstance();
+    }
+
+    private static <T> Set<T> findDuplicates(Iterable<T> potentialDuplicates) {
+        Set<T> duplicateLess = new HashSet<>();
```
dmdashenkov:
```suggestion
        Set<T> unique = new HashSet<>();
```

base/src/main/java/io/spine/validate/DistinctFieldOption.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.Optional;
+
+/**
+ * An option that can be applied to {@code repeated} Protobuf fields to specify that values
+ * represented by that {@code repeated} field don't contain duplicates.
+ *
+ * @param <T>
+ *         type fields that can be checked against this option
+ */
+final class DistinctFieldOption<T> extends FieldValidatingOption<OnDuplicate, T> {
```
dmdashenkov: Why not just `DistinctOption`?

base/src/main/java/io/spine/validate/EnumFieldValidator.java
```diff
@@ -44,13 +47,14 @@ protected boolean isNotSet(EnumValueDescriptor value) {
...
     }
 
+
     /**
-     * {@inheritDoc}
+     * @inheritDoc
      *
-     * <p>Performs no action since no special options are declared for the enum values validation.
+     * <p>All of the additional validation logic is passed as the third
+     * {@linkplain FieldValidator(FieldValue, boolean, Set )} super constructor} parameter.
```
dmdashenkov: This line creates an import as well.

base/src/main/java/io/spine/validate/FieldValue.java
```diff
@@ -101,28 +102,33 @@ static FieldValue of(Object rawValue, FieldContext context) {
...
-                return new EnumFieldValidator(this);
+                return new EnumFieldValidator(castThis());
             default:
                 throw fieldTypeIsNotSupported(fieldType);
         }
     }
 
+    @SuppressWarnings("unchecked")
+    private <S> FieldValue<S> castThis(){
+        return ((FieldValue<S>) this);
```
dmdashenkov: Please remove redundant braces and document.

base/src/main/java/io/spine/validate/MaxConstraint.java
```diff
@@ -0,0 +1,71 @@
...
+
+    @Override
+    public List<ConstraintViolation> check(FieldValue<V> fieldValue) {
+        MaxOption option = fieldValue.valueOf(OptionsProto.max);
+        ImmutableList<V> actualValue = fieldValue.asList();
+        double maxValue = parseDouble(option.getValue());
+        boolean exclusive = option.getExclusive();
+        Predicate<V> exceeds = exclusive ?
+                               value -> value.doubleValue() >= maxValue :
+                               value -> value.doubleValue() > maxValue;
```
dmdashenkov: Please fix the alignment (`?` and `:` at the start of the line, not at the end).

base/src/main/java/io/spine/validate/MessageFieldValidatingOption.java
```diff
@@ -0,0 +1,41 @@
...
+
+    MessageFieldValidatingOption(GeneratedExtension<FieldOptions, O> extension) {
+        super(extension);
+    }
+
+    @Override
+    boolean shouldValidate(FieldValue<M> value) {
+        FieldDeclaration declaration = value.declaration();
+        boolean valid = value.valueOf(OptionsProto.valid);
+        return super.shouldValidate(value) && (declaration.isNotCollection() || valid);
```
dmdashenkov: Please spit this expression.

base/src/main/java/io/spine/validate/OptionInapplicableException.java
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Exception thrown whenever an attempt to apply an option to an incompatible type was found.
+ */
+public class OptionInapplicableException extends RuntimeException {
+
+    private static final long serialVersionUID = 0L;
+
+    /** Creates a new instance of this exception with the specified clarifying message. */
+    public OptionInapplicableException(String message) {
```
dmdashenkov: Seems like there should be a constructor with option and type and format the message.

base/src/main/java/io/spine/validate/FieldValidatingOption.java
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * An option that validates a field.
+ *
+ * Validating options impose constraint on fields that they are applied to.
+ *
+ * @param <T>
+ *         type of information that this option holds
+ */
+abstract class FieldValidatingOption<T, F> extends ValidatingOption<T, FieldValue<F>> {
```
dmdashenkov: Are there non-field validating options?

serhii-lekariev: `RequiredField` is an example of a `MessageValidatingOption`.

base/src/main/java/io/spine/validate/DistinctFieldOption.java
serhii-lekariev: I thought that such naming would imply an option that is distinct from other options.

base/src/main/proto/spine/options.proto
```diff
@@ -735,6 +747,16 @@ message GoesOption {
     string msg_format = 2;
 }
 
+// Defines options of a repeated field that defines a logic
+enum OnDuplicate {
```
alexander-yevsyukov: As discussed vocally, please use `distinct` boolean option so that we have true `Set` behavior.

base/src/main/java/io/spine/validate/FieldValidatingOption.java
```diff
@@ -46,6 +50,29 @@ boolean shouldValidate(FieldValue<F> value) {
         return valueFrom(value).isPresent();
     }
 
+    @Override
+    public Optional<T> valueFrom(FieldValue<F> bearer) {
```
alexander-yevsyukov: Please rename the parameter to `value` (similarly to one passed to `optionValue()`).

base/src/main/java/io/spine/validate/FieldValidator.java
```diff
@@ -290,7 +290,7 @@ protected FieldDeclaration field() {
     }
 
     private Set<FieldValidatingOption<?, V>> commonOptions(boolean isStrict) {
```
alexander-yevsyukov: Please rename the parameter to `strict`.

base/src/main/java/io/spine/validate/Required.java
```diff
@@ -53,25 +52,15 @@
      * @param isOptionPresent
      *         a function that defines whether this option is present
      */
-    private Required(Predicate<FieldValue<?>> isOptionPresent) {
+    Required(boolean isStrict) {
```
alexander-yevsyukov: Please rename to `strict`.

base/src/main/java/io/spine/validate/NumericFieldConstraint.java
```diff
@@ -0,0 +1,64 @@
...
+                            .map(unsatisfactory -> constraintViolated(fieldValue, unsatisfactory))
+                            .collect(toList());
+        return violations;
+    }
+
+    /**
+     * Whether the actual value of the field satisfies this constraint.
+     *
+     * @param value actual value that is being validated
+     * @param fieldValue a value of the field. The difference between this and the first parameter
```
alexander-yevsyukov: The need to explain the difference between parameters is a smell. Please think of avoiding it via having more specific API.

base/src/main/java/io/spine/validate/Distinct.java
```diff
@@ -20,37 +20,49 @@
...
-    public Optional<io.spine.option.OnDuplicate> valueFrom(FieldValue<T> fieldValue) {
-        io.spine.option.OnDuplicate option = fieldValue.valueOf(optionExtension());
-        boolean isDefault = option.getNumber() == 0;
-        return isDefault ? Optional.empty() : Optional.of(option);
+    public Optional<Boolean> valueFrom(FieldValue<T> fieldValue) {
+        FieldDescriptor descriptor = fieldValue.declaration()
+                                               .descriptor();
+        Option distinct = Option.from(descriptor, optionExtension());
+
+        return distinct.isExplicitlySet()
```
alexander-yevsyukov: We have a repeated scenario  with this `isExplicitlySet()` thing:
 1. First we get a thing which is called `Option` 
     This is a part of the problem to me: this thing is a value of an option, not an option itself. The `Option` class does not have a default value and works differently to what we have in Protobuf (which has `Option.getValue()`).

 2. Then, we query if the thing `isExplicitlySet()`.
     In the absence of the notion of the default value in an option, the method looks strange.

3. Then, depending on the value returned by `isExplicitlySet()` we have return the value of the field wrapped into `Optional`, or `empty()`.

Why cannot `fieldValue.valueOf(optionExtension())` perform the checking and return corresponding `Optional`?

alexander-yevsyukov: @armiol, FYI.

base/src/main/java/io/spine/validate/FieldValidatingOption.java
alexander-yevsyukov: How about naming this class as `FieldValueConstraint`?

base/src/main/java/io/spine/validate/MessageFieldValidatingOption.java
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * An option that validates a message field.
+ *
+ * @param <O>
+ *         type of option value
+ * @param <M>
+ *         type of the message being validated by this option
+ */
+abstract class MessageFieldValidatingOption<O, M extends Message>
```
alexander-yevsyukov: The name of the class sounds suspicious. Since we define validation constraints in proto, all these fields are in messages. So, every field value constraint would be the one which is for a field of some message type.

base/src/main/java/io/spine/validate/MessageFieldValidator.java
```diff
@@ -55,9 +47,11 @@
...
      *         such constraint is not explicitly set
      */
-    MessageFieldValidator(FieldValue fieldValue, boolean assumeRequired) {
-        super(fieldValue, assumeRequired, true);
-        this.timeConstraint = fieldValue.valueOf(OptionsProto.when);
+    @SuppressWarnings("unchecked"/*`When` option validates `Timestamp`s, which extend `Message`.*/)
+    MessageFieldValidator(FieldValue<V> fieldValue, boolean assumeRequired) {
+        super(fieldValue,
+              assumeRequired,
+              ImmutableSet.of(((FieldValidatingOption<?, V>) When.create())));
```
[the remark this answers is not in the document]
alexander-yevsyukov: > ImmutableSet.of(((FieldValidatingOption<?, V>) When.create())));

Do we speak Lisp here?

base/src/main/java/io/spine/validate/MessageFieldValidator.java
```diff
@@ -67,11 +61,10 @@ protected void validateOwnRules() {
             validateFields();
             BuiltInValidation.ANY.validateIfApplies(this);
         }
-        BuiltInValidation.TIMESTAMP.validateIfApplies(this);
     }
 
     private boolean shouldValidateFields() {
-        return getValidateOption() && !fieldValueNotSet();
+        return fieldValue().valueOf(OptionsProto.valid) && !fieldValueNotSet();
```
alexander-yevsyukov: Please extract variables for each of the component. It would make it clearer what we check. This is especially important since we have double negation at the end of the expression.

alexander-yevsyukov:
@dmdashenkov, @serhii-lekariev, @armiol

Gentlemen, we also have `io.spine.option.Options` class which has the following TODO in Javadoc:

```java
// TODO:2018-06-25:dmytro.dashenkov: Refactor https://github.com/SpineEventEngine/base/issues/127
```
What if we sort the options matters in this PR, including the [mentioned issue](https://github.com/SpineEventEngine/base/issues/127)?

dmdashenkov:
@alexander-yevsyukov, this is a matter of a follow-up pull request.

base/src/main/java/io/spine/code/proto/FieldOption.java
```diff
@@ -0,0 +1,75 @@
...
+     *
+     * @param extension
+     *         extensions that corresponds to this option
+     * @param <O>
+     *         value of the option
+     * @param <T>
+     *         type of field that the returned option can be applied to
+     * @return a new instance
+     */
+    public static <O, T> FieldOption<O, T> someOption(
```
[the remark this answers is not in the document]
alexander-yevsyukov: Why is it `some`?  It reads as if it was test data.

Why isn't it named `from()`, for example?

base/src/main/java/io/spine/code/proto/Option.java
```diff
@@ -20,73 +20,24 @@
...
 
 /**
- * An option of a Protobuf declaration.
+ * A Protobuf option.
  *
  * @param <T>
- *         the type of the option value
+ *         type of information held by this option
+ * @param <K>
+ *         kind of entities that this option can be applied to
```
alexander-yevsyukov: Is it really an entity, or just an object?

base/src/main/java/io/spine/code/proto/Option.java
```diff
@@ -20,73 +20,24 @@
...
      *
-     * @return {@code true} if the option was explicitly set, {@code false} otherwise
+     * @param bearer
+     *         holder of the option
+     * @return value of this option
      */
-    public boolean isExplicitlySet() {
-        return explicitlySet;
-    }
+    Optional<T> valueFrom(K bearer);
```
alexander-yevsyukov: It it's just an object, let's call it `object`.

base/src/main/java/io/spine/code/proto/FieldOption.java
```diff
@@ -0,0 +1,74 @@
...
+            GeneratedExtension<FieldOptions, O> extension) {
+        return new FieldOption<>(extension);
+    }
+
+    @Override
+    public Optional<O> valueFrom(FieldValue<T> bearer) {
```
alexander-yevsyukov: Please consider renaming `bearer`.  `value` would be find.

base/src/main/java/io/spine/validate/FieldValidator.java
```diff
@@ -150,7 +151,9 @@ final IfInvalidOption ifInvalid() {
      *
      * <p>Do not call this method directly. Use {@link #validate() validate()} instead.
      */
-    protected abstract void validateOwnRules();
+    @SuppressWarnings("NoopMethodInAbstractClass"/* Does nothing for all but one subclasses. */)
```
alexander-yevsyukov: This is not fool-proof.  If there can be only one case of calling this method, please have a fried class that does this. Or, check the caller type. We have `InvocationGuard`, but its usage should be seriously weighted out.

base/src/main/java/io/spine/validate/FieldValue.java
```diff
@@ -169,14 +169,14 @@ JavaType javaType() {
      * @param <O>
      *         the type of the option value
      */
-    <O> Option<O> option(GeneratedExtension<FieldOptions, O> option) {
-        Optional<Option<O>> validationRuleOption = getOptionValue(context, option);
+    <O> O option(GeneratedExtension<FieldOptions, O> option) {
```
alexander-yevsyukov: Please have another letter for the generic parameter. `O` looks like zero. Candidates are: `V` (for value), `F` (for field).

base/src/main/java/io/spine/validate/MessageFieldValidator.java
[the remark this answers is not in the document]
serhii-lekariev: I have failed to come up with anything good to fix this one,

serhii-lekariev:
@dmdashenkov, PTAL.

base/src/main/java/io/spine/validate/FieldValidator.java
```diff
@@ -316,4 +286,9 @@ protected FieldPath getFieldPath() {
     protected FieldDeclaration field() {
         return declaration;
     }
+
+    private Set<FieldValidatingOption<?, V>> commonOptions(boolean strict) {
```
dmdashenkov: I think, this can be `static`.

serhii-lekariev:
@dmdashenkov, PTAL.

base/src/main/java/io/spine/validate/ValidatingOption.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * An option that validates an arbitrary Protobuf entity.
+ *
+ * @param <T>
+ *         type of information that this option holds, e. g.
+ *         {@linkplain io.spine.option.OptionsProto.required required option} would hold a {@code
+ *         Boolean}
+ * @param <K>
+ *         kind entity that this option is applied to
```
dmdashenkov: There's a typo in this line.

base/src/main/java/io/spine/validate/ValidatingOption.java
```diff
@@ -0,0 +1,38 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.spine.validate;
+
+import io.spine.code.proto.Option;
+
+/**
+ * An option that validates an arbitrary Protobuf entity.
```
dmdashenkov:
```suggestion
 * An option that validates a Protobuf entity.
```

base/src/main/java/io/spine/code/proto/FieldOption.java
```diff
@@ -0,0 +1,65 @@
...
+ *         value of this option
+ * @param <T>
+ *         type of field that this option is applied to
+ */
+public class FieldOption<O, T> implements Option<O, FieldValue<T>> {
+
+    private final GeneratedExtension<FieldOptions, O> optionExtension;
+
+    /** Specifies the extension that corresponds to this option. */
+    public FieldOption(GeneratedExtension<FieldOptions, O> optionExtension) {
```
dmdashenkov: Please have `Valid` and `IfMissing` options defined and make the constructor `protected`. This class should _only_ be used as an implementation base, not a ready-baked solution.

base/src/main/java/io/spine/code/proto/FieldOption.java
```diff
@@ -0,0 +1,65 @@
...
+        this.optionExtension = optionExtension;
+    }
+
+    @Override
+    public Optional<O> valueFrom(FieldValue<T> field) {
+        FieldContext context = field.context();
+        Optional<O> optionValue = ValidationRuleOptions.getOptionValue(context, optionExtension);
+        if (optionValue.isPresent()) {
+            return optionValue;
```
dmdashenkov: This detail should reside in `ValidationFieldOption`, another abstraction level for such options. Otherwise, other options would also become overridable, which is not what we want (at least for now). Please also add a test case verifying that options such as `(by)` cannot be defined externally.

serhii-lekariev:
@alexander-yevsyukov, PTAL.

base/src/main/java/io/spine/code/proto/FieldOption.java
serhii-lekariev: Such test would make more sense after the `(by)` option starts using the current `Option` type hierarchy.

base/src/main/java/io/spine/validate/MessageFieldValidator.java
```diff
@@ -47,25 +47,26 @@
...
     MessageFieldValidator(FieldValue<V> fieldValue, boolean assumeRequired) {
-        super(fieldValue,
-              assumeRequired,
-              ImmutableSet.of(((FieldValidatingOption<?, V>) When.create())));
+        super(fieldValue, assumeRequired, additionalOptions());
     }
 
     @Override
-    protected void validateOwnRules() {
+    protected List<ConstraintViolation> validate() {
```
alexander-yevsyukov: Please use `ImmutableList` in results when we do so. It would minimize the need for explaining the fact in Javadocs.

serhii-lekariev:

base/src/main/java/io/spine/validate/FieldValue.java
```diff
@@ -89,6 +92,19 @@ private FieldValue(List<T> values, FieldContext context, FieldDeclaration declar
...
+        return result;
+    }
+
+    @SuppressWarnings({
+            "unchecked", /* Since the field value is created by validating builders,
+                           the raw value always corresponds to one of the Protobuf field types,
+                           so the cast to `T` is always safe.
+                         */
+            "ChainOfInstanceofChecks" /* No common ancestor forces chain of `instanceofs`. */})
+    private static <T> FieldValue<T> resolveType(FieldContext context,
```
alexander-yevsyukov: Please change the order of parameters having this:
```
(FieldDeclaration field, FieldContext context, T value)
```

alexander-yevsyukov: Parameters should tell a little story.

base/src/main/java/io/spine/validate/FieldValue.java
```diff
@@ -89,6 +92,19 @@ private FieldValue(List<T> values, FieldContext context, FieldDeclaration declar
         FieldDescriptor fieldDescriptor = context.getTarget();
         FieldDeclaration declaration = new FieldDeclaration(fieldDescriptor);
 
+        FieldValue<T> result = resolveType(context, value, declaration);
+        return result;
+    }
+
+    @SuppressWarnings({
+            "unchecked", /* Since the field value is created by validating builders,
```
alexander-yevsyukov: Please transform this long suppression comment into Javadoc. And then, reference the Javadoc from the warning suppression.

base/src/main/java/io/spine/validate/MessageFieldValidatingOption.java
```diff
@@ -28,15 +28,15 @@
 /**
  * An option that validates a message field.
  *
- * @param <O>
+ * @param <P>
```
alexander-yevsyukov: Why not `T`?

serhii-lekariev:

base/src/main/java/io/spine/code/proto/FieldOption.java
```diff
@@ -0,0 +1,62 @@
...
+ * A Protobuf option that is applied to fields in Protobuf messages.
+ *
+ * @param <V>
+ *         type of value held by this option
+ * @param <T>
+ *         type of field that this option is applied to
+ */
+public class FieldOption<V, T> implements Option<V, FieldValue<T>> {
+
+    private final GeneratedExtension<FieldOptions, V> optionExtension;
```
alexander-yevsyukov: Please rename the field to `extension` (and the accessor method correspondingly).

base/src/main/java/io/spine/code/proto/FieldOption.java
```diff
@@ -0,0 +1,62 @@
...
+
+    /** Specifies the extension that corresponds to this option. */
+    protected FieldOption(GeneratedExtension<FieldOptions, V> optionExtension) {
+        this.optionExtension = optionExtension;
+    }
+
+    protected GeneratedExtension<FieldOptions, V> optionExtension() {
```
alexander-yevsyukov: Please add this Javadoc:

```
    /**
     * Obtains the Protobuf extension associated with the option.
     */
```

base/src/main/java/io/spine/code/proto/FieldOption.java
```diff
@@ -0,0 +1,62 @@
...
+ * @param <V>
+ *         type of value held by this option
+ * @param <T>
+ *         type of field that this option is applied to
+ */
+public class FieldOption<V, T> implements Option<V, FieldValue<T>> {
+
+    private final GeneratedExtension<FieldOptions, V> optionExtension;
+
+    /** Specifies the extension that corresponds to this option. */
```
alexander-yevsyukov: Please have this Javadoc:
```
    /**
     * Creates an instance with the
     * <a href="https://developers.google.com/protocol-buffers/docs/proto3#custom_options">Protobuf extension</a>
     * that corresponds to this option.
     */
```

base/src/main/java/io/spine/code/proto/Option.java
```diff
@@ -20,73 +20,24 @@
...
 
 /**
- * An option of a Protobuf declaration.
+ * A Protobuf option.
  *
  * @param <T>
- *         the type of the option value
+ *         type of value held by this option
+ * @param <K>
+ *         kind of value that this option is applied to
```
alexander-yevsyukov: Please apply this Javadoc:

```
/**
 * A Protobuf option.
 *
 * @param <T>
 *         the type of a value held by this option
 * @param <K>
 *         the type of object which holds the option, such as “field”, “message”, or “file”
 * @see <a href="https://developers.google.com/protocol-buffers/docs/proto3#custom_options">Protobuf
 *         Custom Options</a>
 */
```

base/src/main/java/io/spine/validate/AbstractValidatingBuilder.java
```diff
@@ -286,8 +286,8 @@ private static boolean isSetOnce(FieldDescriptor field) {
     private void logError(FieldDescriptor field) {
         String fieldName = field.getFullName();
         _error("Error found in `%s`. " +
-                    "Repeated and map fields can't be marked as `(set_once) = true`",
-                    fieldName);
+               "Repeated and map fields can't be marked as `(set_once) = true`.",
```
alexander-yevsyukov: Please replace `can't` according to [these recommendations](https://www.grammarly.com/blog/cannot-or-can-not/).

serhii-lekariev:

serhii-lekariev:
@dmdashenkov, PTAL at the recent changes.

base/src/main/java/io/spine/validate/FieldValidator.java
```diff
@@ -278,44 +273,53 @@ private boolean isRequiredId() {
...
+        IfInvalidOption ifInvalid = ifInvalidOption.valueFrom(descriptor)
                                                    .orElse(IfInvalidOption.getDefaultInstance());
         return ifInvalid;
     }
 
     private IfMissingOption ifMissing() {
-        IfMissing<V> ifMissing = new IfMissing<>();
-        return ifMissing.valueFrom(value)
+        IfMissing ifMissing = new IfMissing();
+        return ifMissing.valueFrom(descriptor())
```
dmdashenkov: Please have a `valueOrDefault(...)` method and replace this duplicative `orElse` call.

base/src/main/java/io/spine/validate/OptionValue.java
```diff
@@ -0,0 +1,25 @@
...
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.spine.validate;
+
+public interface OptionValue {
```
dmdashenkov: Please remove this unused type.

base/src/main/java/io/spine/validate/FieldValidatingOption.java
```diff
@@ -37,38 +40,74 @@
...
+            String fieldName = value.declaration()
+                                    .name()
+                                    .value();
+            String containingTypeName = descriptor.getContainingType()
+                                                  .getName();
+            return illegalState(fieldName, containingTypeName);
+        });
+    }
+
+    private IllegalStateException illegalState(String fieldName, String containingTypeName) {
```
alexander-yevsyukov: Please have a specific name for the method. What went wrong?

base/src/main/java/io/spine/validate/FieldValidatingOption.java
```diff
@@ -37,38 +40,74 @@
...
      */
-    @Override
-    public Optional<T> valueFrom(FieldValue<F> field) {
-        FieldContext context = field.context();
-        Optional<T> validationForOption = ValidationRuleOptions.getOptionValue(context,
-                                                                               optionExtension());
-        return validationForOption.isPresent()
-               ? validationForOption
-               : super.valueFrom(field);
+    private Optional<T> valueFromContextualized(FieldDescriptor descriptor, FieldContext context) {
```
alexander-yevsyukov: Please rename `descriptor` to `field`.

alexander-yevsyukov: Are there cases when we have a non-contextualized input?

base/src/main/java/io/spine/validate/FieldValidator.java
```diff
@@ -65,23 +66,24 @@
...
-        this.ifInvalid = ifInvalid(fieldValue);
+                             ImmutableSet<FieldValidatingOption<?, V>> validatingOptions) {
+        this.value = value;
+        this.declaration = value.declaration();
+        this.values = value.asList();
         this.assumeRequired = assumeRequired;
-        this.fieldValidatingOptions = Sets.union(commonOptions(assumeRequired), validatingOptions);
+        this.ifInvalid = ifInvalid(descriptor(value));
+        this.fieldValidatingOptions = ImmutableSet.copyOf(
+                Sets.union(commonOptions(assumeRequired), validatingOptions));
```
alexander-yevsyukov: Pease extract `commonOptions` variable.

base/src/main/java/io/spine/validate/FieldValidator.java
```diff
@@ -169,11 +166,11 @@ final IfInvalidOption ifInvalid() {
     private List<ConstraintViolation> optionViolations() {
         List<ConstraintViolation> violations =
                 fieldValidatingOptions.stream()
-                                      .filter(option -> option.shouldValidate(value))
+                                      .filter(option -> option.shouldValidate(descriptor()))
                                       .map(option -> option.constraintFor(value))
                                       .flatMap(constraint -> constraint.check(value)
                                                                        .stream())
-                                      .collect(toList());
+                                      .collect(Collectors.toList());
```
alexander-yevsyukov: Please use static import for `toList()`.

base/src/main/java/io/spine/validate/FieldValue.java
```diff
@@ -233,7 +234,7 @@ boolean isDefault() {
 
     private boolean isSingleValueDefault() {
         if (this.singleValue() instanceof EnumValueDescriptor) {
-            return ((EnumValueDescriptor) this.singleValue()).getNumber() == 0;
+            return ((Internal.EnumLite) this.singleValue()).getNumber() == 0;
```
alexander-yevsyukov: Please cast to `EnumValueDescriptor` which is a part of public API of Protobuf.

base/src/main/java/io/spine/validate/MaxConstraint.java
```diff
@@ -20,73 +20,28 @@
...
+        super(optionValue, maxRange(optionValue));
     }
 
-    private boolean isExclusive() {
-        return optionValue.getExclusive();
+    @SuppressWarnings({
+            "unchecked", // is safe because Double is a subtype of both Number and Comparable
+            "WrapperTypeMayBePrimitive" // primitive is uncastable
+    })
+
```
alexander-yevsyukov: Please remove the empty line.

base/src/main/java/io/spine/validate/MinConstraint.java
```diff
@@ -20,73 +20,29 @@
...
+        super(optionValue, minRange(optionValue));
     }
 
-    private boolean isExclusive() {
-        return optionValue.getExclusive();
+    @SuppressWarnings({
+            "unchecked",
+            "WrapperTypeMayBePrimitive" // primitive is uncastable
+    })
+    // is safe because Double is a subtype of both Number and Comparable
```
alexander-yevsyukov: We suppress the same thing twice already. Please refactor.

base/src/main/java/io/spine/validate/RangedConstraint.java
```diff
@@ -0,0 +1,113 @@
...
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Range;
+import io.spine.base.FieldPath;
+
+import static com.google.common.collect.BoundType.CLOSED;
+import static io.spine.protobuf.TypeConverter.toAny;
+import static java.lang.String.format;
+
+public abstract class RangedConstraint<V extends Number & Comparable, T>
```
alexander-yevsyukov: Please document.

base/src/main/java/io/spine/validate/RangedConstraint.java
```diff
@@ -0,0 +1,113 @@
...
+        extends NumericFieldConstraint<V, T> {
+
+    static final String OR_EQUAL_TO = "or equal to";
```
alexander-yevsyukov: In such an abstraction this thing must be private.

serhii-lekariev:

PR #293, opened 2019-01-15
Complete time provider description

mdrachuk:
This PR completes the description of the time provider which previously was left hanging at the most exciting part.


armiol:
@mdrachuk as we have already released `pre3`, this should now be `1.0.0-SNAPSHOT` again.

Also, the changeset are outdated.

mdrachuk:
@armiol, PTAL.

PR #294, opened 2019-01-17
Remove `decimal` number options

### LINKED ISSUE #289: Delete `DecimalMin` and `DecimalMax`
Currently, we have 2 kinds of number range validation options: `(decimal_min)` & `(decimal_max)` and `(min)` & `(max)`. This structure reflects similar Java annotations is the Bean validation framework. However, for our needs, this approach is accessive and brings code duplication.

We should delete `(decimal_min)` & `(decimal_max)` options and allow `(min)` & `(max)` have `inclusive` configuration.

dmdashenkov:
Before this PR, we had two sets of number range options: `(min)`/`(max)` and `(decimal_min)`/`(decimal_max)`. The former used to be inclusive only and the latter—exlusive by default and configurable.

This approach is borrowed from the `java.bean.validation` framework. Though, it does not mean that it's good.

Starting from now, we will have only `(min)`/`(max)` options to configure the accepted range of a number.

Both bounds are inclusive by default. In order to make them exclusive, a user should specify `(min).exclusive = true`.

Both options expect a decimal number representation as the `value`.

This PR fixes #289.

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -3,7 +3,6 @@
     <option name="RIGHT_MARGIN" value="100" />
     <JavaCodeStyleSettings>
       <option name="PREFER_LONGER_NAMES" value="false" />
-      <option name="CLASS_NAMES_IN_JAVADOC" value="2" />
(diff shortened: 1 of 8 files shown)


dmdashenkov:
@armiol, PTAL.

PR #296, opened 2019-01-18
Interfaces for nested types

dmdashenkov:
This PR fixes generating interfaces for nested Protobuf messages.

Before this PR, the `(is)` and `(every_is)` options were ignored for nested messages. Starting from now, they are treated the same for top-level and nested messages.

Important notice: `EventMessage`, `CommandMessage`, and `RejectionMessage` interfaces are not applied to nested messages declared in the corresponding files.
However, `UuidValue` interface may be applied to both top-level and nested messages.


dmdashenkov:
@armiol, PTAL.

base/src/test/proto/spine/test/code/proto/test_commands.proto
```diff
@@ -17,21 +17,27 @@
...
- */
-@Immutable
-final class EventMessageClassifier extends MessageClassifier {
+option (type_url_prefix) = "type.spine.io";
+option (SPI_all) = true;
+option java_package = "io.spine.test.code.proto.command";
+option java_multiple_files = true;
+option java_outer_classname = "MttCommandsProto";
+
+// This file define test environment for `MessageTypeTest.java`.
```
armiol:
```suggestion
// This file defines test environment for `MessageTypeTest.java`.
```

base/src/test/proto/spine/test/code/proto/test_events.proto
```diff
@@ -17,21 +17,27 @@
...
- */
-@Immutable
-final class CommandMessageClassifier extends MessageClassifier {
+option (type_url_prefix) = "type.spine.io";
+option (SPI_all) = true;
+option java_package = "io.spine.test.code.proto.event";
+option java_multiple_files = true;
+option java_outer_classname = "MttEventsProto";
+
+// This file define test environment for `MessageTypeTest.java`.
```
armiol: Same as above.

base/src/test/proto/spine/test/code/proto/test_events.proto
```diff
@@ -17,21 +17,27 @@
...
+
+// A sample event to verify telling that a message is an event from `MessageType`.
```
armiol:
```suggestion
// A sample event to verify detecting that a message is an event from `MessageType`.
```

PR #299, opened 2019-01-21
Fix repeated rejection generation

alexander-yevsyukov:
This PR:
  * Fixes the issue of duplicated rejection generations (#300)
  * Removes outdated proto doc for the `by` option


alexander-yevsyukov:
@armiol, PTAL.

PR #302, opened 2019-01-22
Enrichment code improvements

alexander-yevsyukov:
This PR:
  * De-static `EnrichmentLookup`.
  * Simplifies building `EnrichmentMap`.
  * Annotates sub-packages of `code` as `@Internal` (to avoid exposure in Javadocs).
  * Fixes wrong diagnostics on removing a directory which does not exist.


alexander-yevsyukov:
@armiol, it's a small improvement, but I don't want to hang it in-progress for too long. PTAL.

alexander-yevsyukov:
@dmdashenkov, PTAL.

tools/model-compiler/src/main/java/io/spine/tools/compiler/DirectoryCleaner.java
```diff
@@ -43,8 +43,10 @@
     public static void deleteDirs(List<String> dirs) {
         for (String dirPath : dirs) {
             File file = new File(dirPath);
-            if (file.exists() && file.isDirectory()) {
-                deleteRecursively(file.toPath());
+            if (file.isDirectory()) {
+                if (file.exists()) {
```
dmdashenkov: `File.isDirectory()` returns `false` if the file does not exist. Thus, this check is redundant.

tools/model-compiler/src/main/java/io/spine/tools/compiler/enrichment/ByOption.java
```diff
@@ -38,18 +38,13 @@
...
-class ByOption implements Logging {
+final class ByOption implements Logging {
 
-    private final String packagePrefix;
     private final DescriptorProto message;
     private final FieldDescriptorProto field;
 
-    ByOption(String packagePrefix, DescriptorProto message, FieldDescriptorProto field) {
-        this.packagePrefix = packagePrefix;
+    ByOption(DescriptorProto message, FieldDescriptorProto field) {
```
dmdashenkov: Please check if we're able to migrate to `MessageType` instead of two descriptors.

alexander-yevsyukov: The second descriptor is one for a field. Probably `ByOption` would get more tricks and be based on `MessageType` in another PR related to enrichments.

PR #303, opened 2019-01-24
More enrichments

alexander-yevsyukov:
_WIP_

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,8 +4,11 @@
 **/.idea/shelf
 **/.idea/*.xml
 
-# Do not ignore code style settings.
-!/.idea/codeStyleSettings.xml
+# Do not ignore the following IDEA settings
+!.idea/misc.xml
+!.idea/codeStyleSettings.xml
+!.idea/codeStyles/codeStyleConfig.xml
+!.idea/codeStyles/Project.xml
 
 # Gradle interim configs
 **/.gradle/**
@@ -18,6 +21,15 @@
 
 # Build files produced by the IDE
 **/out/**
+
+# Ignore Gradle GUI config
+gradle-app.setting
+
+# Avoid ignoring Gradle wrapper jar file (.jar files are usually ignored)
+!gradle-wrapper.jar
+
+# Cache of project
+.gradletasknamecache
 
 # Spine internal directory for storing intermediate artifacts
 **/.spine/**
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -1,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="JavaScriptSettings">
+    <option name="languageLevel" value="ES6" />
+  </component>
+  <component name="NullableNotNullManager">
+    <option name="myNullables">
+      <value>
+        <list size="9">
+          <item index="0" class="java.lang.String" itemvalue="org.jetbrains.annotations.Nullable" />
+          <item index="1" class="java.lang.String" itemvalue="javax.annotation.Nullable" />
+          <item index="2" class="java.lang.String" itemvalue="javax.annotation.CheckForNull" />
+          <item index="3" class="java.lang.String" itemvalue="edu.umd.cs.findbugs.annotations.Nullable" />
+          <item index="4" class="java.lang.String" itemvalue="android.support.annotation.Nullable" />
+          <item index="5" class="java.lang.String" itemvalue="androidx.annotation.Nullable" />
+          <item index="6" class="java.lang.String" itemvalue="org.checkerframework.checker.nullness.qual.Nullable" />
+          <item index="7" class="java.lang.String" itemvalue="org.checkerframework.checker.nullness.compatqual.NullableDecl" />
+          <item index="8" class="java.lang.String" itemvalue="org.checkerframework.checker.nullness.compatqual.NullableType" />
+        </list>
+      </value>
+    </option>
+    <option name="myNotNulls">
+      <value>
+        <list size="9">
+          <item index="0" class="java.lang.String" itemvalue="org.jetbrains.annotations.NotNull" />
+          <item index="1" class="java.lang.String" itemvalue="javax.annotation.Nonnull" />
+          <item index="2" class="java.lang.String" itemvalue="javax.validation.constraints.NotNull" />
+          <item index="3" class="java.lang.String" itemvalue="edu.umd.cs.findbugs.annotations.NonNull" />
+          <item index="4" class="java.lang.String" itemvalue="android.support.annotation.NonNull" />
+          <item index="5" class="java.lang.String" itemvalue="androidx.annotation.NonNull" />
+          <item index="6" class="java.lang.String" itemvalue="org.checkerframework.checker.nullness.qual.NonNull" />
+          <item index="7" class="java.lang.String" itemvalue="org.checkerframework.checker.nullness.compatqual.NonNullDecl" />
+          <item index="8" class="java.lang.String" itemvalue="org.checkerframework.checker.nullness.compatqual.NonNullType" />
+        </list>
+      </value>
+    </option>
+  </component>
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" default="false" project-jdk-name="1.8" project-jdk-type="JavaSDK" />
+</project>


alexander-yevsyukov:

I did not put efforts into covering `EnrichmentMapBuilder` since this code is expected to be re-worked in one of the following PRs (in the scope of code-generation enhancements for  enrichments).

base/src/main/java/io/spine/code/proto/ref/FieldRef.java
```diff
@@ -0,0 +1,238 @@
...
+    private static ImmutableList<String> split(String value) {
+        List<String> elements = fieldNameSplit.splitToList(value);
+        return ImmutableList.copyOf(elements);
+    }
+
+    /**
+     * Obtains references found in the passed field.
+     */
+    public static ImmutableList<FieldRef> allFrom(FieldDescriptorProto field) {
+        ImmutableList<String> refs = ByOption.allFrom(field);
```
yuri-sergiichuk: Maybe we can use streams here to make it prettier?
```java
ImmutableList<FieldRef> result = ByOption
        .allFrom(field)
        .stream()
        .map(FieldRef::new)
        .collect(ImmutableList.toImmutableList());
return result;
```

base/src/main/java/io/spine/code/proto/ref/InPackage.java
```diff
@@ -0,0 +1,70 @@
...
+     * A suffix for referencing all types in a proto package.
+     */
+    private static final String WILDCARD = ".*";
+
+    /**
+     * The name of the referenced proto package.
+     */
+    private final PackageName packageName;
+
+    static Optional<TypeRef> parse(String value) {
```
yuri-sergiichuk: Shouldn't we have some docs here as well?

base/src/main/java/io/spine/code/proto/ref/InPackage.java
```diff
@@ -0,0 +1,70 @@
...
+        checkNotNull(value);
+        if (value.endsWith(WILDCARD)) {
```
yuri-sergiichuk: Is `.*` should be allowed? Will it mean that we are referencing all message types within all possible Protobuf packages?

base/src/main/java/io/spine/code/proto/ref/Parsing.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.Optional;
+import java.util.function.Function;
+
+/**
+ * Attempts to parse the passed value by sequentially invoking provider functions.
+ */
+final class Parsing {
+
+    private final String value;
+    private final ImmutableList<Provider> provider;
```
yuri-sergiichuk: Maybe `providers`?

base/src/main/java/io/spine/code/proto/ref/Parsing.java
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     * Provides a type reference <em>if</em> the passed string matches the format
+     * requirements of the provider, otherwise returns an empty {@code Optional}.
+     *
+     * @see CompositeTypeRef#parsePart(String)
+     * @see TypeRef#parse(String)
+     */
+    @FunctionalInterface
+    interface Provider extends Function<String, Optional<TypeRef>> {
```
yuri-sergiichuk: Maybe it should be `Parser` or `TypeRefParser`?  Also, is there any particular reason to extend from `Function`? Having a `@FunctionalInterface` itself should be sufficient and we can provide a cleaner method than a general-purpose `apply`.
The following looks cleaner to me:
```java
@FunctionalInterface
interface TypeRefParser {
     Optional<TypeRef> parse(String reference);
}
```

tools/model-compiler/src/main/java/io/spine/tools/compiler/enrichment/Entry.java
```diff
@@ -20,116 +20,117 @@
...
+import java.util.Comparator;
 import java.util.List;
-import java.util.Map;
 
-import static io.spine.option.OptionsProto.by;
 import static io.spine.util.Exceptions.newIllegalStateException;
 
 /**
- * Obtains event names from the {@code "by"} field option of a message.
+ * Provides mapping from a enrichment type name to names of enriched types.
```
yuri-sergiichuk: from `an` enrichment type

base/src/main/java/io/spine/code/proto/PackageName.java
```diff
@@ -0,0 +1,76 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * A name of a Protobuf package.
+ */
+@Immutable
+public final class PackageName extends StringTypeValue {
+
+    private static final long serialVersionUID = 0L;
+    private static final String DELIMITER = String.valueOf('.');
```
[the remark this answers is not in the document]
alexander-yevsyukov: No, it's just an omission when copied a similar delimiter form another class.

base/src/main/java/io/spine/code/proto/ref/BuiltIn.java
```diff
@@ -0,0 +1,170 @@
...
+     * <li>not a wild card type reference in a suffix form
+     *  (such as {@code '*CommonEventNameSuffix.field_name'}, which is not currently supported.
+     * </ol>
+     */
+    @CanIgnoreReturnValue
+    static String checkTypeReference(String typeReference) {
+        checkNotEmptyOrBlank(typeReference);
+        String wildcard = ALL.value();
+        if (typeReference.startsWith(wildcard)) {
+            checkArgument(
```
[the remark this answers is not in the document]
alexander-yevsyukov: It's a part of `FieldRef` logic. Please see tests for this class. It is possible to add such a reference, meaning that only `*` represents `ALL` type reference. Where “all” can mean types referenced in `(enrichment_for)`. If there is no such option defined, it would really mean “all types” with such a field.

base/src/main/java/io/spine/code/proto/ref/Parsing.java
alexander-yevsyukov: Well spotted!  This way it is simpler.

alexander-yevsyukov:
@yuri-sergiichuk, PTAL.

base/src/main/java/io/spine/code/proto/ref/DirectTypeRef.java
```diff
@@ -0,0 +1,152 @@
...
+     * Compose a potentially nested type name from the passed parts.
+     *
+     * <p>Assumes that a type name starts from an upper-case letter.
+     */
+    private static String toNestedName(List<String> parts) {
+        List<String> types =
+                parts.stream()
+                     .filter(p -> Character.isUpperCase(p.charAt(0)))
+                     .collect(toList());
+        return Joiner.on(TypeName.NESTED_TYPE_SEPARATOR)
```
yuri-sergiichuk: Why `TypeName.NESTED_TYPE_SEPARATOR` is a public field while `PackageName.delimiter()` is a function?

base/src/main/java/io/spine/code/proto/ref/DirectTypeRef.java
```diff
@@ -0,0 +1,152 @@
...
+        return Optional.ofNullable(packageName);
+    }
+
+    /**
+     * Obtains simple type name of the direct type reference.
+     *
+     * <p>If a reference is for a nested type, returned value contains the most nested name.
+     */
+    public String simpleTypeName() {
+        return this.simpleTypeName;
```
yuri-sergiichuk: Is there a particular reason to use `this` here and below? As far as I've checked this is not a common style within the project.

base/src/main/java/io/spine/code/proto/ref/Parsing.java
```diff
@@ -0,0 +1,61 @@
...
+    private final String value;
+    private final ImmutableList<Parser> parsers;
+
+    Parsing(String value, Parser... parser) {
+        this.value = value;
+        this.parsers = ImmutableList.copyOf(parser);
+    }
+
+    Optional<TypeRef> parse() {
+        for (Parser supplier : parsers) {
```
yuri-sergiichuk: I guess it should be a `parser` now and not a `supplier`.

tools/model-compiler/src/main/java/io/spine/tools/compiler/enrichment/Entry.java
```diff
@@ -20,116 +20,117 @@
...
         String fieldName = field.getName();
-        Logger log = log();
-        Collection<String> eventGroup = new HashSet<>(events.size());
-        for (String eventName : events) {
-            if (eventName == null || eventName.trim()
-                                              .isEmpty()) {
-                throw invalidByOptionValue(enrichment);
+        List<String> list = new ArrayList<>();
+        for (String srcType : srcTypes) {
+            if (srcType == null || srcType.trim()
```
yuri-sergiichuk: You can use `Strings.isBlank()` here.

yuri-sergiichuk:
@alexander-yevsyukov PTAL at the comments above. I'm basically good to go, just resolve the comments if you think they are inconsiderable.

tools/model-compiler/src/main/java/io/spine/tools/compiler/enrichment/Entry.java
alexander-yevsyukov: I'd rather avoid dependency on `Roaster` just because of a string utility.

base/src/main/java/io/spine/code/proto/ref/DirectTypeRef.java
alexander-yevsyukov: For consistency, because `TypeName` already have another character constant for package separator. Should we decide to unify, both constants should be migrated.

base/src/main/java/io/spine/code/proto/ref/ByOption.java
```diff
@@ -0,0 +1,113 @@
...
+     *
+     * @see #allFrom(FieldDescriptorProto)
+     */
+    @VisibleForTesting
+    static ImmutableList<String> parse(String rawValue) {
+        String byArgument = SPACE.matcher(rawValue)
+                                 .replaceAll("");
+        String[] result;
+        result = byArgument.contains(PIPE_SEPARATOR)
+                 ? PATTERN_PIPE_SEPARATOR.split(byArgument)
```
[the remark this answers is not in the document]
alexander-yevsyukov: This class is a temporary solution until we have `Option` implementations that do the splitting job in one of the base classes. There's no need to spend time polishing this class.

PR #305, opened 2019-01-29
Set `protoc-api` publish group to `io.spine.tools`

dmitrykuzmin:
Previously, the `spine-protoc-api` artifact was published under auto-assigned `spine-base` group.

This caused problems with dependent repositories build.


dmitrykuzmin:
@armiol PTAL.

Issue #292, opened 2019-01-14
`ValidatingBuilder.mergeFrom` methods are generated with some issues


dmdashenkov:
The `validateSetOnce(...)` check is called in a loop for each field in `mergeFrom(...)` overloads.

However, the generated code looks rather strange:
```java
 @CanIgnoreReturnValue
    public CommandContextScheduleVBuilder mergeFrom(CommandContext.Schedule message) {
        java.util.Map<com.google.protobuf.Descriptors.FieldDescriptor, java.lang.Object> fieldsMap = message.getAllFields();
                for (java.util.Map.Entry<com.google.protobuf.Descriptors.FieldDescriptor, java.lang.Object> entry : fieldsMap.entrySet()) {
                  validateSetOnce(entry.getKey());
                }
                ;
        super.mergeFrom(message);
        return this;
    }
```

The strange things about this code are:
 - the alignment of the for loop;
 - the redundant `;` after the loop.

PR #309, opened 2019-02-01
Validating builders `mergeFrom` method formatting fix

serhii-lekariev:
This PR closes https://github.com/SpineEventEngine/base/issues/292.

Before, all generated `mergeFrom` methods were malformed.

This PR fixes it, making the generated code look as expected.
It also extracts a `Splitter` creation from a loop to a `static final` field, since creating multiple `Splitters` is pointless.

### DIFF
--- a/tools/model-compiler/src/main/java/io/spine/tools/compiler/validation/VBuilderCode.java
+++ b/tools/model-compiler/src/main/java/io/spine/tools/compiler/validation/VBuilderCode.java
@@ -52,7 +52,7 @@ import static java.lang.String.format;
  */
 final class VBuilderCode implements Logging {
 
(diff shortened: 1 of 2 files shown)


serhii-lekariev:
@dmdashenkov, PTAL.

tools/model-compiler/src/main/java/io/spine/tools/compiler/validation/VBuilderCode.java
```diff
@@ -52,6 +52,7 @@
  */
 final class VBuilderCode implements Logging {
 
+    private static final Splitter DOT_SPLITTER = Splitter.on('.');
```
dmdashenkov: `PACKAGE_SPLITTER`.

Issue #310, opened 2019-02-04
Drop support of nested enrichments and FQN (by) option


yuri-sergiichuk:
It should be not possible to define enrichments as nested event messages, as such usage of enrichments is misleading and does not stand for the actual enrichment purpose.

Also, in order to make the current enrichments syntax more strait-forward and precise enrichments should not be automatically discovered with FQN in `(by)` option.

alexander-yevsyukov:
No longer applicable since enrichments no longer use proto options.

PR #311, opened 2019-02-04
Remove type reference from `(by)` option

alexander-yevsyukov:
This PR simplifies the way enrichment types are defined:
  1. Now an enriched type can be only referenced from the `(enrichment_for)` option.
  2. Referencing a type for a qualified field reference from the `(by)` option is no longer allowed. (See #310.)

Other notable changes:
  * `FieldContext` moved to the `code.proto` package. 
  * `FieldContext` methods were renamed avoiding the `get` prefix.


alexander-yevsyukov:

PR #312, opened 2019-02-04
Remove (enrichment) option

### LINKED ISSUE #308: Drop support of `(enrichment)` option
It was decided that `(enrichment)` option should be removed. The reason behind such a decision is a potential ability of a developer to define enrichments from different places at the same time: e.g. from an event that should be enriched and from the enrichment itself.

```protobuf
message Enrichment {
    string user_name = 1 [(by) = "user_id];
}

message SecondEnrichment {
    option (enrichment_for) = "Event";

    string user_address = 1 [(by) = "user_id];
}

message Event {
    option (enrichment) = "Enrichment";
    
    int32 user_id = 1;
}
```
The enrichment is smth that adds some additional information on top of a message and it is not a responsibility of the message to define how this information is linked to the message itself.

Enrichments should be defined only with `(enrichment_for)` option.

yuri-sergiichuk:
This PR closes #308.

`(enrichment)` option is not available anymore. The contract of `(enrichment_for)` and `(by)` options are refined and made stricter.

Further changes to the enrichment functionality should be done in the following PRs. (e.g., see #310 and #311).


base/src/main/proto/spine/options.proto
```diff
@@ -180,59 +180,49 @@ extend google.protobuf.FieldOptions {
...
+    // changed without any notices.
+    //
+    // An example:
+    //
+    // message MyEnrichment {
+    //     option (enrichment_for) = "example.events.*"
     //
-    // If using the FQN field names, the `enrichment_for` annotation may be omitted
-    // (the same behavior as with a single-argument `by` annotation).
+    //     string username = 1 [(by) = "rule:com.acme.UsernameEnrichmentRule"
```
alexander-yevsyukov: Let's not give this in documentation until we finally agreed on how it's going to work (if at all).

base/src/main/proto/spine/options.proto
```diff
@@ -180,59 +180,49 @@ extend google.protobuf.FieldOptions {
     //
     // The value of the option is the name of the target event message/context field
     // which is used to set the corresponding enrichment's field.
-    // The value must have event/context FQN prefix if `enrichment_for` option is omitted
-    // ("example.events.MyEvent.user_id" instead of just "user_id").
+    // The value must have event/context plain field name (i.e. FQN names are not supported).
```
alexander-yevsyukov: A general suggestion: let's not use `FQN` abbreviation in the end user documentation. We can always say “a fully-qualified name”. 

Regarding this field documentation, the value of `(by)` option may contain more than one field reference, separated with the pipe character.

base/src/main/proto/spine/options.proto
```diff
@@ -386,42 +376,28 @@ extend google.protobuf.MessageOptions {
     //     ...
     // }
     //
+    // The enrichment can qualify for all messages within a package using suffix wildcard (*)
```
alexander-yevsyukov: Please have the second part as a new sentence.

tools/model-compiler/src/main/java/io/spine/tools/compiler/enrichment/EnrichmentMapBuilder.java
```diff
@@ -51,15 +51,19 @@
...
-    private final TypeRefs enrichmentOption;
 
     /** Multimap for storing intermediate results. */
     private final HashMultimap<String, String> multimap = HashMultimap.create();
 
+    /**
+     * Creates enrichment map builder with a supplied {@code packagePrefix}.
+     *
+     * @param packagePrefix
+     *         a Protobuf package prefix (including the dot)
```
alexander-yevsyukov: I'd say: “... including the trailing dot”.

yuri-sergiichuk:
@alexander-yevsyukov PTAL again.

Issue #297, opened 2019-01-18
Enhance code generation for enrichments


alexander-yevsyukov:
Now Model Compiler gathers information about enrichments into the property file (named `enrichments.properties`) which maps a type of an enrichment to the type it enriches.

Obtaining an enrichment from `EventContext` is currently awkward. Users of the framework need a more straight (and domain-driven!) API for obtaining enrichment message. It is suggested to extend Model Compiler so that generated enrichment messages provide easier API.

Suppose we have an enrichment message, which is declared like this:
```proto
// Enrichment for the `OrderUpdated` event which provides sections of the updated order.
message SectionList {
     option (enrichment_for) = "ackme.ordering.OrderUpdated";
     repeated SectionId section = 1;
}
```
We want to have event handler of the enriched event to look like this:
```java
@Suscribe 
void to(OrderUpdated event, Context ctx) {
   SectionList sections = SectionList.fromOrderUpdated(ctx);
   // ...
}
```
Here the call `SectionList.fromOrderUpdated() ` is a generated static factory method.
Presumably, each enrichment message would implement an interface `EnrichmentMessage`.

When working on this issue please pay attention to providing customization of the type of the interface used by the Model Compiler (via `generateInterfaces` Gradle script closure).

alexander-yevsyukov:
@armiol, @dmdashenkov, @yuri-sergiichuk, this issue summarizes our discussion. Please see if you want to add or change something.

yuri-sergiichuk:
As discussed vocally, the `EnrichmentMessage` interface should not be parameterized, cause it'll not work as soon as we'll have at least two enrichment targets:
```protobuf
// Enrichment for the `OrderUpdated` and `OrderCreated` events which provides sections of the updated order.
message SectionList {
     option (enrichment_for) = "ackme.ordering.OrderUpdated,"
                               "ackme.ordering.OrderCreated";
     repeated SectionId section = 1;
}
```

alexander-yevsyukov:
@yuri-sergiichuk, isn't it already resolved?

yuri-sergiichuk:
Yeah, it is obsolete.

PR #314, opened 2019-02-05
Generate marker interface for enrichments

yuri-sergiichuk:
This PR is a part of the enrichments code generation improvements (#297).

From now all enrichment messages (messages that have `(enrichment_for)` option) will implement a common marked interface.
By default the `io.spine.base.EnrichmentMessage` interface is used.

The default value can be overridden in gradle `modelCompiler.generateInterfaces` configuration using `enrichmentMessage` option.

Here is how we can use `com.example.Enrichment` interface as an enrichment marker interface:

```groovy
modelCompiler {
    generateInterfaces {
        enrichmentMessage().markWith("com.example.Enrichment")
    }
}
```

It is also possible to disable generation of the enrichment message marker interface as follows:

```groovy
modelCompiler {
    generateInterfaces {
        enrichmentMessage().ignore()
    }
}
```

### DIFF
--- a/tools/model-compiler/src/main/java/io/spine/tools/gradle/compiler/GeneratedInterfaces.java
+++ b/tools/model-compiler/src/main/java/io/spine/tools/gradle/compiler/GeneratedInterfaces.java
@@ -24,11 +24,13 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Objects;
 import io.spine.annotation.Internal;
 import io.spine.base.CommandMessage;
+import io.spine.base.EnrichmentMessage;
 import io.spine.base.EventMessage;
 import io.spine.base.MessageFile;
 import io.spine.base.RejectionMessage;
 import io.spine.base.UuidValue;
 import io.spine.code.java.ClassName;
+import io.spine.tools.protoc.EnrichmentInterface;
 import io.spine.tools.protoc.SpineProtocConfig;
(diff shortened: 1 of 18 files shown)


tools/model-compiler/src/main/java/io/spine/tools/gradle/compiler/GeneratedInterfaces.java
```diff
@@ -172,29 +185,70 @@ public GeneratedInterfaceConfig uuidMessage() {
         return uuidInterfaceConfig;
     }
 
+    /**
+     * Configures an interface generation for messages with {@code (enrichment_for)} option.
+     *
+     * <p>This method functions are similar to the {@link #filePattern(FilePattern)} except for
+     * several differences:
+     * <ul>
+     * <li>the file in which the message type is defined does not matter;
```
armiol: Please indent.

tools/model-compiler/src/main/java/io/spine/tools/gradle/compiler/GeneratedInterfaces.java
```diff
@@ -155,6 +164,10 @@ public PostfixPattern endsWith(String postfix) {
         return new PostfixPattern(postfix);
     }
 
+    private PostfixPattern endsWith(MessageFile messageFile) {
```
armiol: Before merging let's discuss why you need this method in scope of this changeset.

PR #317, opened 2019-02-06
Fix the way strings are resolved to number types.

serhii-lekariev:
Currently, `Min` and `Max` validating options have to extract their actual numeric minimum and maximum values from a `String`. 
Since both `MinConstraint` and `MaxConstraint` use typed `Range<T>` range instances, this may lead to `ClassCastExceptions` at run-time, since `T` in this case may be both a `Long` and an `Integer`.

This PR fixes it, by resolving the `String` value to the smallest possible value.

This PR also suppresses some of the unsuppressed warnings.


serhii-lekariev:
@alexander-yevsyukov, PTAL.

Issue #321, opened 2019-02-07
Wrong warnings on boolean fields marked as required


alexander-yevsyukov:
The build console for `core-java` has the following:

```
WARNING: Fields of type BOOL should not be declared as `(required)`.
```
This is reported for `spine.client.Target.include_all` field, which is `boolean`, but does not have the `(required)` option set.

PR #322, opened 2019-02-08
Get rid of false positive `required BOOL` warnings during validation.

serhii-lekariev:
This PR closes https://github.com/SpineEventEngine/base/issues/321.

Before, a check for whether a `required` option was correctly applied to a field (i.e. the type of the field allowed to check for whether the value is present) happened during the evaluation of the `shouldValidate()` method.

This behaviour is incorrect, since such check should happen only when a field has been found to be required, as opposed to when the field is being checked for the presence of the `required` option.

This PR moves the warning-producing method calls into a more appropriate location.


base/src/main/java/io/spine/validate/Required.java
```diff
@@ -111,6 +112,7 @@ private void checkCanBeRequired(FieldDescriptor field) {
 
     @Override
     public Constraint<FieldValue<T>> constraintFor(FieldValue<T> fieldValue) {
+        checkCorrectUsage(fieldValue.descriptor());
```
alexander-yevsyukov: Please rename to `checkUsage()`.

PR #324, opened 2019-02-10
`EnrichmentContainer` interface

alexander-yevsyukov:
This PR introduces `EnrichmentContainer` interface so that generated `EnrichmentMessage`s can obtain instances from the container.

This interface is implemented by message contexts (which are defined in `core-java` sub-project, see [this PR](https://github.com/SpineEventEngine/core-java/pull/966)).


alexander-yevsyukov:
@armiol, PTAL.

base/src/main/java/io/spine/util/NamedProperty.java
```diff
@@ -31,10 +31,9 @@
  * The abstract base for classes obtaining a value of a named property.
  *
  * @param <T> the type of the property
- * @param <O> the type of the object from which we get the value
- * @author Alexander Yevsyukov
+ * @param <H> the type of the object from which we get the value
  */
-public abstract class NamedProperty<T, O> {
+public abstract class NamedProperty<T, H> {
```
armiol: I assume `H` stands for "holder"?

To me it looks more `S` (source) or `O` (object, as previously).

alexander-yevsyukov: We're recently renaming `<O>` to something less looking like zero.

PR #325, opened 2019-02-10
Enable PMD static analysis during the build

armiol:
This PR enables the PMD code analysis for `base` modules.

Some warnings were suppressed and some violations were addressed.

Checkstyle rules aren't enabled with this PR due to the https://github.com/checkstyle/checkstyle/issues/3238, which causes Checkstyle to crash on `LogMessages` with `@Nullable ... params`.


armiol:
@alexander-yevsyukov PTAL.

PR #326, opened 2019-02-10
Obtain enrichments via Known Types

alexander-yevsyukov:
This PR eliminates the generation of enrichment map property resources in favor of obtaining `EnrichmentType`s from `KnownTypes`. As the result the Gradle plug-in no longer generates the property files. Instead, verification of created enrichment message declarations is checked via generated descriptors.

Other notable changes:
  * Improved method names in type-related classes.
  * `MessageContext` is now a part of the `base` module (residing under `io.spine.base` package).
  * `Messages.newInstance()` method (previously deprecated) is removed.
  * `DirectoryCleaner` was de-static.


alexander-yevsyukov:

base/src/main/java/io/spine/code/proto/enrichment/FieldRef.java
```diff
@@ -47,30 +51,35 @@
...
     FieldRef(String value) {
         super(checkValue(value));
-        this.parts = split(value);
-        this.typeRef = TypeRef.parse(typeRef(value));
+        ImmutableList<String> parts = split(value);
+        this.typeRef = CONTEXT.parse(parts.get(0))
+                              .orElse(ANY);
+        // If the first element is context reference, skip it from the path.
+        ImmutableList<String> fieldPath =
+                parts.subList((this.typeRef == CONTEXT ? 1 : 0), parts.size());
```
yuri-sergiichuk: Why not use `isContext` here instead of equality check?

base/src/main/java/io/spine/code/proto/enrichment/FieldRef.java
```diff
@@ -115,68 +124,54 @@ private static void checkThat(boolean b, String value) {
      * Verifies if the reference is to a field from the same type.
      */
     public boolean isInner() {
-        boolean result = typeRef.equals(SELF);
+        boolean result = typeRef.equals(ANY);
```
yuri-sergiichuk: Let's use identity checks here and below: `ANY == typeRef`.

base/src/main/java/io/spine/code/proto/ref/DirectTypeRef.java
```diff
@@ -53,10 +52,9 @@
      */
     static Optional<TypeRef> parse(String value) {
         checkNotNull(value);
-        if (value.contains(InPackage.WILDCARD)) {
-            return Optional.empty();
-        }
-        if (value.equals(SELF.value())) {
+        if (value.isEmpty()
```
yuri-sergiichuk: What about `blank` values? Are they supported?

I'd better use `value.trim().isEmpty()` instead.

base/src/main/java/io/spine/type/KnownTypes.java
```diff
@@ -137,10 +143,23 @@ public ClassName getClassName(TypeUrl typeUrl) throws UnknownTypeException {
...
     }
 
+    /**
+     * Obtains known enrichment types.
+     */
+    public ImmutableSet<EnrichmentType> enrichments() {
+        ImmutableSet<EnrichmentType> result =
+                typeSet.messageTypes()
+                       .stream()
+                       .filter(t -> EnrichmentType.test(t.descriptor()))
```
yuri-sergiichuk: Maybe it worth adding a shortcut `test` method to `EnrichmentType` that accepts `MessageType` and not a descriptor? We'll have a cleaner code here with `filter(EnrichmentType::test)`.
 Also, it's the only usage of the `test` method.

base/src/main/java/io/spine/type/KnownTypes.java
```diff
@@ -157,11 +176,25 @@ public TypeSet getAllTypes() {
...
+
+    /**
+     * Obtains all types matching the passed type reference.
+     */
+    public ImmutableSet<MessageType> allMatching(TypeRef typeRef) {
+        ImmutableSet<MessageType> result =
+                asTypeSet().allTypes()
+                           .stream()
+                           .filter(t -> t instanceof MessageType)
+                           .map(t -> (MessageType) t)
```
yuri-sergiichuk: Maybe it worth replacing lambdas with method references here:
```
.filter(MessageType.class::isInstance)
.map(MessageType.class::cast)
```

tools/model-compiler/src/main/java/io/spine/tools/compiler/validation/VBuilderGenerator.java
```diff
@@ -73,10 +74,12 @@ public VBuilderGenerator(File protoSrcDir, File targetDir, Indent indent) {
     public void process(FileSet files) {
         FileSet fileSet = moduleFiles(files);
         ImmutableCollection<MessageType> messageTypes = TypeSet.onlyMessages(fileSet);
+        Predicate<MessageType> notRejection =
+                ((Predicate<MessageType>) MessageType::isRejection).negate();
         ImmutableList<MessageType> customTypes =
                 messageTypes.stream()
                             .filter(MessageType::isCustom)
-                            .filter(MessageType::isNotRejection)
+                            .filter(notRejection)
```
yuri-sergiichuk: We can use `Predicates.not(MessageType::isRejection)` here, with a static import it looks cleaner to me: `not(MessageType::isRejection)`.

tools/model-compiler/src/main/java/io/spine/tools/compiler/annotation/MethodNameAnnotator.java
```diff
@@ -59,7 +59,7 @@ public void annotate() {
             SourceVisitor<?> visitor = new AnnotateMethods();
             descriptors().stream()
                          .map(TypeSet::from)
-                         .flatMap(typeSet -> typeSet.types().stream())
+                         .flatMap(typeSet -> typeSet.allTypes().stream())
                          .map(type -> type.javaClassName().resolveFile())
```
yuri-sergiichuk: The more Java8 style will be:
```
descriptors().stream()
    .map(TypeSet::from)
    .map(TypeSet::allTypes)
    .flatMap(ImmutableSet::stream)
    .map(Type::javaClassName)
    .map(ClassName::resolveFile)
    .forEach(file -> rewriteSource(file, visitor));
```

And, IMO it is a bit more readable and easy to understand.

base/src/main/java/io/spine/base/MessageContext.java
```diff
@@ -0,0 +1,37 @@
...
+import com.google.protobuf.Message;
+
+/**
+ * Base interface for message contexts.
+ */
+@SuppressWarnings("InterfaceNeverImplemented") // implemented by generated code.
+public interface MessageContext extends Message {
+
+    /**
+     * Obtains type name suffix required for all message context classes.
```
armiol: `MessageContext.typeSuffix()` looks as weird as `User.weatherForecast()` to me.

 Can you please say a few words on the design purpose.

base/src/main/java/io/spine/code/proto/enrichment/EnrichmentType.java
```diff
@@ -0,0 +1,201 @@
...
+import static com.google.common.collect.ImmutableSet.toImmutableSet;
+
+/**
+ * An enrichment type is a message which is added to a context of another message
+ * to extends its information for the benefit of message handlers.
+ *
+ * <p>An enrichment message declaration has:
+ * <ol>
+ *   <li>The {@link io.spine.option.OptionsProto#enrichmentFor
+ *       (enrichment_for)} option with a {@linkplain TypeRef reference} to one or more message types
```
armiol: `(enrichment_for)}` would look better on the previous line.

base/src/main/java/io/spine/code/proto/enrichment/EnrichmentType.java
```diff
@@ -0,0 +1,201 @@
...
+     * Obtains an enrichment type for the passed name.
+     */
+    public static EnrichmentType from(MessageType type) {
+        checkNotNull(type);
+        Descriptor descriptor = type.descriptor();
+        return from(descriptor);
+    }
+
+    /**
+     * Verifies if the passed message type is an enrichment type.
```
armiol: The "valid / invalid" terms aren't used here. However, there is "the enrichment declaration is not valid" above. Is this method somehow related to "valid"? Another thing is that "valid / invalid" is also about message and constraints. 

I would not use "valid" term in this class if it's not related to message constraints.

base/src/main/java/io/spine/code/proto/enrichment/FieldRef.java
alexander-yevsyukov: Let's not escape the constructor here. There's no much sense.

base/src/main/java/io/spine/code/proto/enrichment/EnrichmentType.java
alexander-yevsyukov: I've updated docs. But I'd rather tell more about validity of declaration because we do validate that the declaration is supported by known types and their fields.

alexander-yevsyukov: Not in terms of validation framework constraints, but it's still validation. I understand that the term is too broad, though.

Issue #323, opened 2019-02-08
Range option boundaries are not checked.


serhii-lekariev:
`Range` option constraints a numeric field value with two boundaries.
However, those boundaries are not checked, so it is possible to declare a field that cannot fit into the interval described in the `range` option value.

Example:

```proto
message CatchTwentyTwo {
    
    int32 value = 1 (range = "(23..21)"];
}
````
The declaration above states that the `value` has to be greater than 23 but less than 21, which is an empty set of numbers. An empty `(22..22)` interval can also be declared without any errors.

The declared interval should be checked for emptiness, and an appropriate exception must be thrown.

PR #327, opened 2019-02-11
Empty ranges in `range` option.

serhii-lekariev:
This PR closes https://github.com/SpineEventEngine/base/issues/323.

The goal is to control for `range = "(x..x)"`-esque declarations, to protect the user from defining an empty range which leads to a guaranteed `ValidationException`.


serhii-lekariev:
@alexander-yevsyukov, currently, `com.google.collect.Range` (`Range` by Guava), which one of implementation details of `Constraint` does not allow to create empty ranges, throwing an IAE with `Invalid range:(x..x)`-esque message.

Should we just allow Guavas exception to be thrown?

base/src/test/java/io/spine/validate/RangeConstraintTest.java
```diff
@@ -78,4 +92,13 @@ void throwOnMalformedRanges(String badRange) {
                 Arguments.of("[3...5]")
         );
     }
+
+    private static Stream<Arguments> rangeCombinationsFor(Number left, Number right) {
+        return Stream.of(
+                Arguments.of(format("[%s..%s]", left, right)),
```
alexander-yevsyukov: This block does not stand the modification of the enumeration of possible edges. Please refactor so that you don't have to repeat the code which is largely the same.

base/src/test/java/io/spine/validate/RangeConstraintTest.java
```diff
@@ -59,6 +60,19 @@ void throwOnMalformedRanges(String badRange) {
...
+        assertThrows(IllegalArgumentException.class,
+                     () -> RangeConstraint.rangeFromOption(emptyRange));
+    }
+
+    private static Stream<Arguments> emptyRanges() {
+        return Stream.concat(rangeCombinationsFor(1, 0),
+                             Stream.of(Arguments.of("(0..0)")));
+    }
+
     private static Stream<Arguments> badRanges() {
```
alexander-yevsyukov: Why cannot `List` or `Set` be used instead of `Stream`? We have fixed set of values to check. We don't really stream.

base/src/test/java/io/spine/validate/RangeConstraintTest.java
```diff
@@ -59,6 +60,19 @@ void throwOnMalformedRanges(String badRange) {
...
         return Stream.of(
                 Arguments.of("{3..5]"),
```
alexander-yevsyukov: Please refactor so that you don't have to repeat `Arguments.of()` calls.

serhii-lekariev:
@alexander-yevsyukov, PTAL.

serhii-lekariev:
@alexander-yevsyukov PTAL, this PR got a little bit forgotten.

PR #331, opened 2019-02-12
Mute all Spine logging in the `MuteLogging` extension

dmitrykuzmin:
This PR extends a `MuteLogging` tool with the functionality to mute all the Spine logging.

Previously, the `MuteLogging` tool redirected the `System.out` and `System.err` streams to some temporary location.

But in some cases, when we are running a lot of tests in series, this is not enough as `Logger`s are basically the universal static state and manipulations with their streams are not always predictable.

So, now the `MuteLogging` tool will manually mute all the Spine logging. It was thus extracted to the separate `io.spine.tools:spine-mute-logging` module as it now depends on `base`.


base/src/main/java/io/spine/logging/LoggerClassValue.java
```diff
@@ -73,13 +93,44 @@ protected Logger computeValue(Class<?> type) {
...
+     *
+     * <p>In case of test environment, this logger's output can still be redirected via
+     * {@link SubstituteLogger#setDelegate(Logger)}.
+     */
+    private Logger noopLogger(Class<?> cls) {
+        Logger result = maybeConvertToSubstitute(NOP_LOGGER, cls);
+        return result;
+    }
+
+    private Logger maybeConvertToSubstitute(Logger logger, Class<?> cls) {
```
armiol: `substituteIfAvailable()`

base/src/main/java/io/spine/logging/LoggerClassValue.java
```diff
@@ -47,22 +49,40 @@
      */
     private final @MonotonicNonNull SubstituteLoggerFactory substFactory;
 
+    /**
+     * Is {@code true} if this class value should be resolved to a
```
armiol: Remove `Is `.

dmitrykuzmin:
@armiol PTAL.

PR #334, opened 2019-02-14
Do not generate `typeUrl()` method for services in ProtoJs plugin

### LINKED ISSUE #328: ProtoJs plugin generates `typeUrl()` method for Protobuf services
Currently the `TypeSet` for which a `ProtoJs` appends `typeUrl()` method includes all types gathered from the classpath, including service definitions.

However, the `.js` file generated by the original `protoc` has no JS definition for a service — that's normal. But the method is generated anyway, appending the `typeUrl` function to a non-existent `... .prototype`.

We need to filter out the service definitions. For a quick reference see `AppendTypeUrlGetter.typeUrlMethods()` where _all_ types are currently selected.

As a side-effect, the JS code coverage is failed to build in `web` module — since the JS code in the service JS-definitions is invalid. Therefore in `web` the code coverage is going to be temporarily disabled.

dmitrykuzmin:
This PR fixes #328.

Service definitions are now skipped when generating `typeUrl()` method in ProtoJs plugin.

Spine version is changed to `1.0.0-SNAPSHOT`.


dmitrykuzmin:
@armiol PTAL.

tools/proto-js-plugin/src/main/java/io/spine/js/generate/index/KnownTypesMap.java
```diff
@@ -41,6 +41,9 @@
  *
  * <p>This class generates the map with all the known types written in the form of
  * "{@linkplain io.spine.type.TypeUrl type-url}-to-JS-type".
+ *
+ * <p>Note that services are not included in the known types map as such types are basically
```
armiol: @dmitrykuzmin I think this is going to be a problem. We have intentionally included services into known types previously, as we wanted to make our code generation working for services as well.

Can we resolve the issue and _not_ modify the behavior of `KnownTypes`?

dmitrykuzmin:
@armiol PTAL.

Issue #335, opened 2019-02-17
Allow short package references in `(enrichment_for)`


alexander-yevsyukov:
We now require a full package name in types or packages referenced in the `(enrichment_for)` option.

It's not convenient and may easily result in an error because what we have here is a string.  We don't have an ability to type check here. The longer the path, the more we have chances to make an error.

Consider the following declaration:

```proto
message EttAlternativeFieldNames {
    option (enrichment_for) =
           "spine.test.code.enrichment.type.user.sharing.EttSharingRequestApproved,"
           "spine.test.code.enrichment.type.user.permission.EttPermissionGrantedEvent";

    string user_google_uid = 1 [(by) = "second_user_uid | user_uid"];
}
```

It would be more convenient to write:

```proto
message EttAlternativeFieldNames {
    option (enrichment_for) =
           "sharing.EttSharingRequestApproved,"
           "permission.EttPermissionGrantedEvent";

    string user_google_uid = 1 [(by) = "second_user_uid | user_uid"];
}
```

When referencing a message in a field declaration, Protobuf type resolution allows to find types prefixed like this (provided that a corresponding file is correctly imported). Such types can be under the package of the message which uses them as fields, or _above_ the message.

This flexibility allows to move the types while the domain language and the project structure are still under definition.

The open question is how to treat two types that have the same short name, but reside in packages with the same nested name, but in another branches. 

How does Protobuf treat these cases? If it does, how do we support it? If not, what do we do?

alexander-yevsyukov:
@dmdashenkov, have a look at this issue. I'd like to give it to @serhii-lekariev as the next assignment, after I finish with my current enrichment improvements.

What would you say?

@armiol, FYI.

alexander-yevsyukov:
I have a sample file `spine/test/code/enrichment/type/type_reference_sample.proto` which demonstrates various type referencing scenarios. See the `EttReferenceTest` type, which is available from [this PR](https://github.com/SpineEventEngine/base/pull/326).


It's now in the

alexander-yevsyukov:
This is no longer applicable since enrichments are reworked and no longer use proto options.

PR #338, opened 2019-02-21
Allow to reference messages by names shorter than their FQNs.

serhii-lekariev:
This PR closes https://github.com/SpineEventEngine/base/issues/335.

Before, when referencing types in `(enrichment_for)` option, an FQN of the referenced message or package had to be specified.

Now this rule is relaxed, and reference rules that repeat Protobuf  are in place.

## Referencing messages

Now, if referencing messages in the same package, only a name of the message has to specified.

When referencing messages from packages outside of the package of the enrichment message, 
all common packages may be omitted.

For example: 

```proto

package spine.type.user.permission;

message UserAllowedEditingRights {
    
UserId id = 1;
}
```
```proto
package spine.type.user;

message EditingRightsDetails {
    option (enrichment_for)  = "permission.UserAllowedEditingRights";

    repeated string details = 1;

}
```

As seen above, only `permission` package had to be specified explicitly, since `spine.type.user` packages are common between both of the messages.

## Referencing packages

Referencing packages has also been improved.

For example:

```proto
package spine.type.user;

message User {
   
 UserId id = 1;
   ...
}
```
a message like above can be enriched by:

```proto
package spine.type.user;

message UserDetails {
   option (enrichment_for) = "user.*";

   string details = 1 [(by) = "id"];
}
```

above example shows how to enrich a package that is shared between the enriching message and the target of enrichment.

In order to reference a package that is adjacent to the current one, its short name followed by `.*` can  be used.

For example:

```proto

package spine.type.customers.user;

message User {

    UserId id = 1;
}
```
can be enriched by

```proto 
package spine.type.customers.enrichments;

message UserDetails {
    option (enrichment_for) = "user.*";

    repeated string details = 1 [(by) = "id"];
}
```

Had `UserDetails` been located under `spine.type`, `option (enrichment_for) = "customers.*"` or `customers.user.*` would have worked.


serhii-lekariev:
@alexander-yevsyukov, PTAL.

alexander-yevsyukov:
> Now, if referencing messages in the same package, only a name of the message has to specified.
I think it used to work prior to this PR. Didn't it?

alexander-yevsyukov:
> To reference an entire package, FQN has to be used always...

This conflicts with the referencing types. It should work the same way.

base/src/main/java/io/spine/code/proto/enrichment/EnrichmentForOption.java
```diff
@@ -126,7 +127,13 @@ protected MessageOptions optionsFrom(Descriptor object) {
      * Makes sure that if a passed type reference is direct reference to a type,
      * it is a fully-qualified reference, or becomes one as the result of this method.
      */
+    @SuppressWarnings("ChainOfInstanceofChecks") /* TypeRef interface does not provide
+                                                    the means to check the package. */
     private static TypeRef ensurePackage(PackageName packageName, TypeRef ref) {
+        if (ref instanceof InPackage) {
```
alexander-yevsyukov: Why do we need to cast? Why cannot we have a virtual `expand()` method that would return the same type, but with the fully-qualified reference to the same thing?

It could be a decorator per type, as we probably want to retain the original reference information.

base/src/main/java/io/spine/code/proto/ref/InPackage.java
```diff
@@ -30,7 +30,7 @@
 /**
  * Reference to all message types in a proto package.
  */
-final class InPackage extends AbstractTypeRef {
+public final class InPackage extends AbstractTypeRef {
```
alexander-yevsyukov: You don't have to open this class, if you build in the knowledge about the short reference expansion into the base interface.

serhii-lekariev:
@alexander-yevsyukov, PTAL.

base/src/main/java/io/spine/type/KnownTypes.java
```diff
@@ -225,6 +239,19 @@ private Type get(TypeName name) throws UnknownTypeException {
         return result;
     }
 
+    /**
+     * Gathers package names of all known message types.
+     */
+    private static ImmutableSet<PackageName> packageNames(TypeSet typeSet) {
```
alexander-yevsyukov: Have a look at `io.spine.reflect.PackageGraph`. We may have something similar for proto packages, and then we don't have to create sub-packages for each of the package. We'd have the full picture created once.

base/src/main/java/io/spine/code/proto/PackageName.java
```diff
@@ -84,4 +93,96 @@ public boolean isInnerOf(PackageName parentCandidate) {
...
+     * Obtains all subpackages of this package.
+     *
+     * <p>A subpackage is a package nested in the current one.
+     *
+     * <p>A package is always its own subpackage.
+     *
+     * <p>For example, for a package {@code spine.type.user.customers} subpackages are exactly
+     * {@code [spine, spine.type, spine.type.user, spine.type.user.customers]}.
+     */
+    public ImmutableList<PackageName> subpackages() {
```
alexander-yevsyukov: We had an API which speaks about nesting in terms of `inner`. This is how Protobuf documentation refers to them.

Please also see my comment about having a graph.

base/src/main/java/io/spine/code/proto/enrichment/BuiltIn.java
```diff
@@ -73,6 +83,16 @@ public boolean test(Descriptor message) {
             return message.getName()
                           .endsWith(typeSuffix);
         }
+
+        @Override
+        public Optional<PackageName> packageName() {
```
alexander-yevsyukov: What if a type reference refers to more than one package, like composite type reference? Or, there are two packages in a short form that match different branches having the common infix?

base/src/main/java/io/spine/code/proto/PackageName.java
```diff
@@ -35,6 +42,8 @@
     private static final long serialVersionUID = 0L;
     private static final String DELIMITER = ".";
 
+    private static final Splitter PACKAGE_SPLITTER = Splitter.on(DELIMITER)
+                                                             .omitEmptyStrings();
```
alexander-yevsyukov: So we can have something like `spine..type..foo....bar`?

serhii-lekariev:
This PR is frozen.
See https://github.com/SpineEventEngine/base/issues/345.

alexander-yevsyukov:
This PR is no longer relevant since we've removed enrichment options.

PR #339, opened 2019-02-22
Remove `events` and `rejections` options

dmitrykuzmin:
This PR removes `events` and `rejections` Spine options as Model entries should describe entity state rather than behavior.

The options are used in `users` project but there is no evidence that anyone is parsing them, so the removal shouldn't introduce any errors.


dmitrykuzmin:
@armiol PTAL.

alexander-yevsyukov:
@armiol, until this PR we used to speak about messages as entity states. Now with these options they also describe behavior. It's sort of confusing. Let's discuss.

base/src/main/java/io/spine/code/proto/MessageType.java
```diff
@@ -64,6 +65,15 @@ public MessageType(Descriptor descriptor) {
         super(descriptor, true);
     }
 
+    /**
+     * Creates a new instance from the given {@code Message} class.
+     */
+    public static MessageType of(Class<? extends Message> aClass) {
```
alexander-yevsyukov: Please have a constructor instead. We have types that derive from `MessageType`. Such a static method may be a dis-service.

alexander-yevsyukov:
@dmitrykuzmin, I relation to my comment regarding the factory method of `MessageType`, there are some changes to the type-related classes that I'm currently working on. See #340.

armiol:
@alexander-yevsyukov the reason we speak of `lifecycle` is the same as speaking of which `(events)` are emitted in response to which commands. Let's discuss vocally.

dmitrykuzmin:
@armiol PTAL again.

PR #340, opened 2019-02-24
Consolidate `types` package

alexander-yevsyukov:
This PR moves type-related classes to the `io.spine.type` package.

Notable API changes:
  * Now it's possible to create a `RejectionType`, `EnrichmentType`, and `EnumType` by the descriptor. Previously, this API was package-private.
  * `LocationPath` class becomes public.
  * `MessageClassifiers` utility class is eliminated.


alexander-yevsyukov:
@armiol, PTAL.

PR #343, opened 2019-02-25
Make a copy of "live" view on Gradle tasks in Protoc plugin before iterating through them

armiol:
Currently the version updates of Spine dependencies aren't possible. With the updated versions of Gradle plugins something will change. In particular, the way Protoc plugin attempts to modify a list of Gradle tasks while iterating through the same list. 

So basically instead of inserting an element into a collection being traversed, this PR changes the way we treat a collection. And before iterating through tasks, we create a copy of the Gradle task collection and use it for iteration.


armiol:
@dmdashenkov PTAL.

PR #344, opened 2019-02-26
Allow set same value if `(set_once) = true`

dmdashenkov:
Currently, an enabled `(set_once)` option forbids the user to set any kind of value to the field. This is inconvenient since sometimes users may want to set the same value to the field. The `hasField(...)` checks are the only way to make sure that an error would not be produced in such a case.

This PR allows users to set the same value to the field.

Also, the `(set_once)` checks are being added to `setRaw...(...)` and `clear...()` methods.
`put...(...)` and `add...(...)` methods now have explicit checks that the collection field is not marked with `(set_once)`.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -8,10 +8,15 @@
 *.html text
 *.xml text
 *.css text
+*.scss text
 *.js text
 *.properties text
-*.bat text
 *.rtf text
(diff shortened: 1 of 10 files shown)


dmdashenkov:
@armiol, PTAL.

PR #346, opened 2019-02-26
Validate enrichment type

alexander-yevsyukov:
This PR introduces validation of an enrichment type.

API changes:
 * `EnrichmentType` moved under `type` package.
 * `ref` package moved under `type` (from `code.proto`).
 * `enrichment` package moved under `type`.


alexander-yevsyukov:
@yuri-sergiichuk, PTAL.

This is interim set of changes before reworking the enrichments.

base/src/test/java/io/spine/type/ref/CompositeTypeRefTest.java
```diff
@@ -62,7 +62,7 @@ void emptyRef() {
 
         void assertRejects(String value) {
             assertThrows(IllegalArgumentException.class,
-                         () -> CompositeTypeRef.doParse(value));
+                                    () -> CompositeTypeRef.doParse(value));
```
yuri-sergiichuk: Please re-format this file. Looks like smth strange happened to format here.

PR #350, opened 2019-03-04
Incremental build features for some Model Compiler tasks

dmdashenkov:
This PR optimizes the build time of Spine-based projects by allowing to skip some tasks as `UP-TO-DATE` or `NO-SOURCE`.

The [incremental build](https://docs.gradle.org/current/userguide/more_about_tasks.html#sec:up_to_date_checks) feature allows Gradle to skip some tasks if their input and output files have not been changed since the latest run. 

If a task has no inputs at all (e.g. the input directories are empty), the task is marked as `NO-SOURCE` and skipped. 

If a task has inputs and does not have outputs or the inputs or outputs have been changed since the latest run, the task is executed as usual.

If a task has both inputs and outputs which have not been changed since the latest run, the task is marked as `UP-TO-DATE` and skipped.

Tasks that support incremental build are:
 - `generateValidators` and `generateTestValidators`;
 - `generateRejections` and `generateTestRejections`;
 - `copyPluginJar`.

Also, in this PR we delete the `EnrichemntLookupPlugin`, as it does nothing for quite some time already.


dmdashenkov:
@armiol, PTAL.

tools/model-compiler/src/main/java/io/spine/tools/gradle/compiler/Module.java
```diff
@@ -0,0 +1,164 @@
...
+/**
+ * A source code module.
+ *
+ * <p>A module is a set of source code, generated artifacts and temporary files aligned in a certain
+ * layout. In terms of Gradle, a module is all the contents of a Gradle project.
+ *
+ * <p>It is assumed that the model compiler plugin is applied to the Gradle project represented by
+ * this module.
+ */
+final class Module {
```
armiol: Judging on its API it looks more `ProtoModule`, than just a `Module`.

tools/plugin-base/src/main/java/io/spine/tools/gradle/SourceScope.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+package io.spine.tools.gradle;
+
+/**
+ * A name of a source set scope.
+ */
+public enum SourceScope {
+
+    MAIN("main"),
```
armiol: I think there should be something like that in Gradle plugins, such as Java plugin. Not sure we should duplicate that.

alexander-yevsyukov:
Why Do We Have Title Capitalization For This PR?

Is it after some popular song, I'm not aware of? :)

dmdashenkov:
@armiol, PTAL again.

PR #359, opened 2019-03-13
Add plugin classpath to gradle project

dmdashenkov:
This PR allows `GradleProject` to configure `GradleRunner` to include the "plugin-under-test" classpath into the test run.

This helps to avoid publishing Gradle plugins to the local Maven repository just to test them.

We should migrate all our plugin tests to this way of establishing dependencies in the near future.


dmdashenkov:
@armiol, PTAL.

PR #371, opened 2019-03-15
Fix generation of vbuilders for nested types

### LINKED ISSUE #364: Validating Builder is not generated for a `Message` which has a repeated field with nested `Message`
In this case the `OrderVBuilder` is not generated:
```
message OrderId {
    string uuid = 1;
}

message Order {
    OrderId id = 1;
    string name = 2 [(required) = true];
    repeated Item item = 3;

    message Item {
        string name = 1 [(required) = true];
        string description = 2;
    }
}
```
(_see [this test example](https://github.com/SpineEventEngine/core-java/blob/master/server/src/test/proto/spine/test/model/contexts/orders/order.proto)_.)

In the generated sources `OrderItemVBuilder` is generated properly. However, `OrderVBuilder` is not — probably because it should have been referencing `OrderItemVBuilder` in its operation.

alexander-yevsyukov:
This PR fixes #364.


alexander-yevsyukov:
@armiol, PTAL.

PR #375, opened 2019-03-22
Store Spine Protoc Plugin configuration in files

### LINKED ISSUE #373: Protoc could not be run on Windows when the project classpath is huge
Windows OS has a hard [limit](https://support.microsoft.com/en-us/help/830473/command-prompt-cmd-exe-command-line-string-limitation) of the command-line string length equal to 8191 characters and in some cases, when the method factory classpath contains a lot of entries, we can hit this limit.

This is not an issue in UNIX OSs, cause the general limit in UNIX systems is 131072 characters.

yuri-sergiichuk:
This PR fixes #373.

In order to avoid hitting the terminal's max chars limit in Windows OS now and, potentially, the same limit in UNIX OSs, the Spine Protoc Plugin configuration is now stored in the Gradle build temporary folder and the configuration path is supplied to the plugin as the input parameter.

The configuration itself is composed and written between Gradle `clean` and `generateProto` tasks, but the path to the configuration file is determined during the `generateProto` task configuration.


tools/plugin-base/src/main/java/io/spine/tools/gradle/TaskName.java
```diff
@@ -134,6 +134,20 @@
...
+     * Creates the Protobuf compiler plugin configuration.
+     *
+     * <p>Works only with the {@code main} scope.
+     */
+    writePluginConfiguration,
+
+    /**
+     * Creates the Protobuf compiler plugin test configuration.
+     *
+     * <p>Works only with the {@code test} scope.
```
dmdashenkov: It's OK not to specify this if the headline comment already mentions `test`.

tools/protoc-plugin/src/test/java/io/spine/tools/protoc/given/CodeGeneratorRequestGiven.java
```diff
@@ -65,38 +68,46 @@ private CodeGeneratorRequestGiven() {
...
+     * Creates a {@link SpineProtocConfig} out of the supplied {@code GeneratedInterfaces} and
+     * {@code GeneratedMethods} and stores it in the supplied {@code path}.
+     *
+     * @return configuration file absolute path
      */
-    public static String encodedProtocConfig(GeneratedInterfaces interfaces,
-                                             GeneratedMethods methods) {
-        SpineProtocConfig protocConfig = SpineProtocConfig
+    public static Path
+    protocConfig(GeneratedInterfaces interfaces, GeneratedMethods methods, Path configPath) {
```
dmdashenkov: Please reduce the param count. Seems like this method does 2 different things: building and writing.

PR #382, opened 2019-03-26
Encode Spine Protoc Plugin custom parameters

### LINKED ISSUE #379: Protoc fails to split custom option from output directory on Windows
Custom Protoc plugin options are passed via the `protoc` executable to the plugin itself as the part of `pluginName_out` command-line argument, where custom options are comma-delimited and actual `out` option is separated from custom options with a colon (`:`):

```
protoc ... --customPlugin_out=<first_custom_option>,<second_custom_option>...,<last_custom_option>:plugin_output ...
```

In our case, the only custom parameter we specify for the Spine Protoc Plugin is the path to the plugin configuration:

```
protoc ... --spinePlugin_out=/path/to/plugin/configuration/config.pb:/plugin/output/path ...
```

The Protoc plugin itself receives only the custom option in the passed `CodeGeneratorRequest`, so the plugin output is stripped by the Protoc executable.

Such flow works just fine in UNIX OSs, but for some reason, Protoc cannot strip the output path on Windows OS:

```
protoc ... --spinePlugin_out=C:\path\to\plugin\configuration\config.pb:C:\plugin\output\path ...
```

Probably, the issue here is that Windows uses a colon as part of the path.

The easiest way to fix the issue is to Base64-encode all custom parameters. It'll prevent any possible path-related issues.

yuri-sergiichuk:
This PR fixes #379.

Now Spine Protoc Plugin parameters are base64-encoded.

### DIFF
--- a/tools/model-compiler/src/main/java/io/spine/tools/gradle/compiler/ProtocConfigurationPlugin.java
+++ b/tools/model-compiler/src/main/java/io/spine/tools/gradle/compiler/ProtocConfigurationPlugin.java
@@ -20,6 +20,7 @@
 
 package io.spine.tools.gradle.compiler;
 
+import com.google.common.base.Charsets;
 import com.google.common.collect.ImmutableList;
 import com.google.protobuf.gradle.ExecutableLocator;
 import com.google.protobuf.gradle.GenerateProtoTask;
@@ -48,6 +49,7 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
+import java.util.Base64;
 import java.util.Collection;
 
 import static com.google.common.base.Preconditions.checkNotNull;
@@ -246,8 +248,16 @@ public class ProtocConfigurationPlugin extends SpinePlugin {
                           options -> {
                               options.setOutputSubDir("java");
                               String option = spineProtocConfigPath.toString();
-                              options.option(option);
+                              String encodedOption = base64Encoded(option);
+                              options.option(encodedOption);
                           });
+    }
+
+    private static String base64Encoded(String value) {
+        Base64.Encoder encoder = Base64.getEncoder();
+        byte[] valueBytes = value.getBytes(Charsets.UTF_8);
+        String result = encoder.encodeToString(valueBytes);
+        return result;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


tools/model-compiler/src/main/java/io/spine/tools/gradle/compiler/ProtocConfigurationPlugin.java
```diff
@@ -246,10 +248,18 @@ private void configureTaskPlugins(GenerateProtoTask protocTask, Task dependency)
...
                               String option = spineProtocConfigPath.toString();
-                              options.option(option);
+                              String encodedOption = base64Encoded(option);
+                              options.option(encodedOption);
                           });
     }
 
+    private static String base64Encoded(String value) {
+        Base64.Encoder encoder = Base64.getEncoder();
+        byte[] valueBytes = value.getBytes(Charsets.UTF_8);
```
dmdashenkov: Please add static imports for inner classes and constants.

tools/protoc-plugin/src/main/java/io/spine/tools/protoc/Plugin.java
```diff
@@ -90,6 +92,13 @@ private static SpineProtocConfig readConfig(CodeGeneratorRequest request) {
         }
     }
 
+    private static String decodeBase64(String value) {
+        Base64.Decoder decoder = Base64.getDecoder();
+        byte[] decodedBytes = decoder.decode(value);
+        String result = new String(decodedBytes, Charsets.UTF_8);
```
dmdashenkov: Same as above.

tools/protoc-plugin/src/test/java/io/spine/tools/protoc/given/CodeGeneratorRequestGiven.java
```diff
@@ -107,7 +109,15 @@ public static Path protocConfig(GeneratedInterfaces interfaces, Path configPath)
         } catch (IOException e) {
             throw new IllegalStateException(e);
         }
-        return configPath.toAbsolutePath();
+        return base64Encoded(configPath.toAbsolutePath()
+                                       .toString());
+    }
+
+    private static String base64Encoded(String value) {
+        byte[] valueBytes = value.getBytes(Charsets.UTF_8);
```
dmdashenkov: Same as above.

PR #384, opened 2019-03-28
Delete generated Google `.proto`s

### LINKED ISSUE #246: Do not generate code for standard Protobuf types
The directory `base/generated/main/java` contains Java code generated for standard Protobuf types.

We should use classes from the library instead.

dmdashenkov:
In this PR we prune the code generated from the Google `.proto` files as soon as it is generated.

We still need to feed `protoc` with the Google `.proto` files in order to receive correct descriptor files.

Fixes #246.

Also, in this PR we clean up the signature of `Valid` option accessor class.

### DIFF
--- a/base/script/protobuf.gradle
+++ b/base/script/protobuf.gradle
@@ -78,27 +78,8 @@ static boolean isGoogleProtoSource(final FileTreeElement file) {
 }
 
 /**
- * Checks if the given directory hosts files generated from the Google `.proto` sources.
- * 
- * <p>The directory should be called `com` and should have a single child directory called `google`.
- */
-static boolean isCompiledGoogleProtoDir(final FileTreeElement file) {
-    if (!file.directory) {
-        return false
-    }
-    if (!file.name.equals('com')) {
-        return false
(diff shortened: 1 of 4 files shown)


dmdashenkov:
@armiol, PTAL.

base/script/protobuf.gradle
```diff
@@ -31,8 +31,24 @@ dependencies {
...
 }
 
+ext {
+    compiledProtoRoot = "$projectDir/generated"
+    mainCompiledProto = "$compiledProtoRoot/main/java"
+    testCompiledProto = "$compiledProtoRoot/test/java"
+}
+
+final def mainPruneTask = task pruneGoogleProtos(type: Delete) {
+    delete "$mainCompiledProto/com"
```
armiol: I think it's safer to be more specific here. I.e. if we mean both Google Protobuf and gRPC, let's have two.

Also, I would extract the package prefix somewhere — since you have `com` listed more than once. That would also be a good place to comment on the relation between the prefixes and the file origin.

PR #385, opened 2019-03-29
Extensible validation

dmdashenkov:
This PR introduces an extension API for the Spine validation framework.

Previously, the validation options used to be hardwired into the validators. Now, the options are dynamically registered at runtime.

In order to register custom validation options, a user should implement the `ValidationOptions` interface and make the implementation discoverable for the `ServiceLoader` by adding the `io.spine.validate.ValidationOptions` file to the resources. The file should contain the FQN of the implementation of `ValidationOptions`.

Since now the users have to be able to implement own `FieldValidatingOption`s and `Constraint`s, some methods in those classes and the related classes are made `protected` or `public` instead of package-private.

In this PR we move the `(when)` option into the `time` module and adds it to the validation framework using the mechanism described above. `Timestamps2` utility is also moved to `time`.

Also, this PR updates Gradle to version 5.3.1.


dmdashenkov:
@armiol, PTAL.

dmdashenkov:
@armiol, PTAL again.

dmdashenkov:
@armiol, PTAL.

dmdashenkov:
@armiol, PTAL again.

base/src/main/java/io/spine/validate/PrimitiveValidatingOptionFactory.java
```diff
@@ -0,0 +1,77 @@
...
+package io.spine.validate;
+
+import com.google.common.collect.ImmutableSet;
+import com.google.errorprone.annotations.Immutable;
+import io.spine.annotation.Internal;
+
+import java.util.Set;
+
+/**
+ * An implementation of {@link ValidatingOptionFactory} which adds common validation options.
```
armiol: Now it does not add any common validation options. Please update the docs.

PR #392, opened 2019-04-15
`SerializableMessage` as part of API

alexander-yevsyukov:
This PR removes `@Internal` annotation from `SerializableMessage`. This interface is used in `core-java` for binding generic parameters. So we cannot hide it.


alexander-yevsyukov:
@dmdashenkov, PTAL.

PR #393, opened 2019-04-16
Enable license report generation

armiol:
In this PR a report on licenses of 3rd-party dependencies was added. 

Its generation is automatic and is triggered upon each build. In this way, we don't miss any potential changes in transitive dependencies of libraries Spine uses.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -116,6 +116,8 @@ subprojects {
 
     apply plugin: 'maven-publish'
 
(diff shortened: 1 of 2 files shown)


armiol:
@alexander-yevsyukov PTAL.

armiol:
As discussed vocally, it's fine to have a single `license-report.md` file, which is modified upon every build.

PR #390, opened 2019-04-11
Gradle utils from bootstrap

dmdashenkov:
This PR refactors the Gradle tools used in Spine and by our users:
 - all the Gradle-related and source-code-related utilities are moved into `plugin-base` and `tool-base`:
    - Gradle utils now reside in `io.spine.tools.gradle`;
    - code generation tools reside in `io.spine.code.gen` in `tool-base`;
    - project structure tools reside in `io.spine.code.fs` in `tool-base`;
 - Gradle utilities declared in the `bootstrap` repository are moved into `plugin-base`;
 - the unused Gradle Reflections plugin is deleted.

API changes:
 - the unused `StringOption` class is deleted;
 - the `ResourceFiles` utility replaces with the `Resource`—a wrapper object;
 - some methods of `ClassName` and other `base` types are removed and substituted with methods of `tool-base` types (e.g. `ClassName.resolveFile()` is substituted with `SourceFile.whichDeclares(ClassName)`).

As the result of the refactoring, we do not expose code generation tools to the clients and do not force the client code to implicitly depend on the JavaPoet library.

Some file-operating tools are left in `base`. These reside in `io.spine.code.proto`. The reason for this is the fact that files are part of the Protobuf runtime model.

PR #394, opened 2019-04-17
Update config

dmdashenkov:
This PR updates the config submodule.

Since #390, the config submodule pointer was set to a remote branch. This branch is now merged into `master` and deleted. Thus, we now revert the pointer to `master`.


dmdashenkov:
@armiol, PTAL.

PR #398, opened 2019-04-19
Add repeating testing utility

alexander-yevsyukov:
This PR adds a testing utility for repeating an action a number of times.

It is test-only utility to avoid such constructs in a production code because cycles should have domain-specific semantics.

### DIFF
--- a/testlib/src/main/java/io/spine/testing/Tests.java
+++ b/testlib/src/main/java/io/spine/testing/Tests.java
@@ -30,6 +30,7 @@ import java.lang.reflect.Constructor;
 import java.lang.reflect.Modifier;
 import java.util.List;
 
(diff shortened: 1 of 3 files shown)


alexander-yevsyukov:
@dmdashenkov, PTAL.

testlib/src/main/java/io/spine/testing/Tests.java
```diff
@@ -251,4 +252,14 @@ public static void assertInDelta(long expectedValue, long actualValue, long delt
         long actualDelta = abs(expectedValue - actualValue);
         assertTrue(actualDelta <= delta);
     }
+
+    /**
+     * Repeats the passed action the {@code 'count'} number of times.
```
dmdashenkov: I think, `'` symbols may be omitted.

PR #400, opened 2019-04-22
Provide time zone

dmdashenkov:
This PR adds the `Time.currentTimeZone()` method so that all the time-related downstream code can rely on it.

Also, we rename the default value of `EntityOption.Visibility` into `DEFAULT` so that it has more reach semantics than a plain "UNDEFINED".

The `EntityOptions` utility class is removed. This functionality is moved to `server`, as that is where it's used.

Also, this PR fixes the `UseValidatingBuilder` check. Previously, the check was applied to all the Protobuf types, including the standard Google types. Now, it only applies to non-standard types.


dmdashenkov:

base/src/main/proto/spine/options.proto
```diff
@@ -705,8 +705,18 @@ message EntityOption {
...
+
+        // Default visibility is different for different types of entities:
+        //  - for projections, "FULL" is default;
+        //  - for aggregates and process managers, "NONE" is default.
+        //
+        DEFAULT = 0;
+
+        // Same as "DEFAULT".
+        //
+        // This alias is kept for backward compatibility.
```
armiol: Can we do that at all?

dmdashenkov:
@armiol, PTAL again.

tools/errorprone-checks/src/main/java/io/spine/tools/check/vbuilder/matcher/CustomProtobufType.java
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * A predicate which matches custom (i.e. non-Google) Protobuf types.
+ */
+final class CustomProtobufType implements TypePredicate {
+
+    private static final long serialVersionUID = 0L;
+
+    @SuppressWarnings("DuplicateStringLiteralInspection") // Used in another context.
+    private static final String GOOGLE_PROTOBUF_PACKAGE = "com.google.protobuf";
```
armiol: We use a different approach is `IsNotGoogleProto`. I.e. we test for "google" to be included in the package name. That seems to be a more clever solution — see the numerous cases [here](https://github.com/googleapis/googleapis). 

It would also be good to extract the predicate from both scenario implementations and reuse.

dmdashenkov: Let's discuss.

dmdashenkov:
@armiol, PTAL.

PR #401, opened 2019-04-23
Build validation rules from known types

### LINKED ISSUE #396: External validation rules defined in `validation_rules.properties` erase each other during the WAR assembly
Apparently we currently still rely on a file called `validation_rules.properties` to read the definitions of external rules.

External validation rules are tricky. They are used to define constraints for `Message` types, which code isn't accessible (e.g. Google Protobuf definitions). And potentially there could be a number of such external definition rules in a project classpath. So to ease the classpath scanning we just export all the external rules to a file and count it an optimization.

However, when assembling a WAR file or a fat-JAR, such an approach leads to failure — since each of the modules to include into a WAR/JAR/EAR has its own `validation_rules.properties` file. So they are going to overwrite each other. It's far from obvious to the naked eye of an end-user.

As long as we read all the `.desc` files on startup and initialize `KnownTypes` instance, we may as well fetch the external rule definitions from the descriptors. Then cache the rules in an in-memory cache and apply them as soon as some `Message` instance is validated.

yuri-sergiichuk:
This PR fixes #396.

External constraints (previously validation rules) are now assembled from `KnownTypes`. Also, any `KnownTypes` extension will trigger an update of `ExternalConstraints` and `ExternalConstraintOptions` singletons.

As a part of the PR we have also renamed `validationOf` Protobuf option to `constraintFor`. During an update to the current `base` module version, the compilation will fail in case of previous `validationOf` usage.


yuri-sergiichuk:
@armiol PTAL.

base/src/main/java/io/spine/validate/rule/ValidationRules.java
```diff
@@ -46,73 +57,128 @@
...
+            rules.addAll(newRules);
+            instance = new ValidationRules(ImmutableSet.copyOf(rules));
+            ValidationRuleOptions.Holder.updateFrom(instance.rules);
+        }
+    }
+
+    /**
+     * Determines if a {@link MessageType} contains a validation rule.
+     */
+    private static class IsValidationRule implements Predicate<MessageType>, Logging {
```
armiol: It looks `HasValidationRule`.

base/src/main/java/io/spine/type/KnownTypes.java
```diff
@@ -269,6 +278,7 @@ public static void extendWith(TypeSet moreKnownTypes) {
             try {
                 TypeSet newKnownTypes = instance.typeSet.union(moreKnownTypes);
                 instance = new KnownTypes(newKnownTypes);
+                ValidationRules.updateFrom(instance.typeSet.messageTypes());
```
armiol: I am not sure why we update the rules with all the message types every time. We could use just the message types from `moreKnownTypes` — and just add the newly discovered rules to the existing set.

base/src/main/java/io/spine/validate/rule/ValidationRules.java
```diff
@@ -21,22 +21,33 @@
...
+import static io.spine.util.Exceptions.newIllegalArgumentException;
 
 /**
  * A collection of {@linkplain ValidationRule validation rules} known to the application.
  *
  * <p>During initialization of this class, definitions of validation rules are verified.
  * If an invalid validation rule was found, a runtime exception will be thrown.
  */
-public final class ValidationRules {
+public final class ValidationRules implements Serializable {
```
armiol: Are these external message constraints?

For a while, we have been referring to the "validation rules" as to the "message constraints. And as I understand, this class is about `ExternalMessageConstraints`.

yuri-sergiichuk:
@armiol PTAL.

base/src/main/java/io/spine/type/KnownTypes.java
```diff
@@ -115,9 +116,11 @@ private static TypeSet load() {
      * Retrieves a Java class name generated for the Protobuf type by its type URL
      * to be used to parse {@link Message Message} from {@link Any}.
      *
-     * @param type {@link Any} type url
+     * @param type
+     *         {@link Any} type url
```
armiol: I think it's `URL`.

base/src/main/java/io/spine/type/KnownTypes.java
```diff
@@ -183,7 +187,8 @@ public TypeSet asTypeSet() {
     /**
      * Shows if the given {@link TypeUrl} is known the system.
      *
-     * @param typeUrl the {@link TypeUrl} to look up
+     * @param typeUrl
+     *         the {@link TypeUrl} to look up
```
armiol: Let's substitute `{@link}` with `{@code}`, as in the generated Javadoc there is going to be a link to the parameter type anyway.

base/src/main/java/io/spine/validate/constraint/ConstraintOf.java
```diff
@@ -28,13 +28,13 @@
...
 
 /**
- * An external validation for a field.
+ * An external constraint for a field.
  *
- * <p>Contains information about validation of another field, described by the option value.
+ * <p>Contains information about constraint of another field, described by the option value.
  */
-final class ValidationOf extends MessageOption<String> {
+final class ConstraintOf extends MessageOption<String> {
```
armiol: This option needs to be called `ValidationOf` — as we are speaking of the process. And in Protobuf it's still named `validation_of`.

base/src/main/java/io/spine/validate/constraint/ExternalConstraintOptions.java
```diff
@@ -0,0 +1,162 @@
...
+     * Obtains value of the specified option by the specified field context.
+     *
+     * @param fieldContext
+     *         the field descriptor to obtain the option
+     * @param option
+     *         the option to obtain
+     * @param <T>
+     *         the type of the option value
+     * @return the {@code Optional} of option value or {@code Optional.empty()}
+     *         if there is not option for the field descriptor
```
armiol: `if there is no such an option for the field descriptor`

base/src/main/java/io/spine/validate/constraint/ConstraintOf.java
```diff
@@ -28,13 +28,13 @@
...
 
-    ValidationOf() {
+    ConstraintOf() {
         super(OptionsProto.validationOf);
```
armiol: We will need to rename the Protobuf option as well. To me, `constraints_for` looks good:

```proto
message UkraineOnly {
    option (constraint_for) = "com.acme.User.country";
    
    string name = 1 [(pattern).regex = "^Ukraine$"];
}
```
@alexander-yevsyukov @dmdashenkov FYI.

base/src/main/java/io/spine/validate/constraint/ExternalConstraints.java
```diff
@@ -0,0 +1,187 @@
...
+import static com.google.common.collect.ImmutableSet.toImmutableSet;
+import static com.google.common.collect.Sets.newHashSetWithExpectedSize;
+import static io.spine.util.Exceptions.newIllegalArgumentException;
+
+/**
+ * A collection of {@linkplain ExternalMessageConstraint external constrains} known to
+ * the application.
+ *
+ * <p>During initialization of this class, definitions of external constraints are verified.
+ * If an invalid constraint was found, a runtime exception will be thrown.
```
armiol: `If an invalid constraint is found, a runtime exception is thrown.`

Also, I would explain what you mean by "invalid" and "verified".

base/src/main/java/io/spine/validate/constraint/ExternalConstraintOptions.java
```diff
@@ -0,0 +1,162 @@
...
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.spine.validate.constraint;
```
armiol: As discussed, let's move classes from this package to `io.spine.validate`.

And create an issue to review `io.spine.validate` contents, as it's getting too big. Perhaps, once the validation framework becomes more pluggable, it gets possible to move the constraints to sub-packages.

yuri-sergiichuk: I've created #403 issue.

yuri-sergiichuk:
@armiol PTAL

yuri-sergiichuk:
The PR is closed because it was not showing the latest changes. The #404 is a new attempt.

PR #404, opened 2019-04-26
Build validation rules from known types

### LINKED ISSUE #396: External validation rules defined in `validation_rules.properties` erase each other during the WAR assembly

yuri-sergiichuk:
This PR fixes #396 and is a copy of the #401 PR.

External constraints (previously validation rules) are now assembled from `KnownTypes`. Also, any `KnownTypes` extension will trigger an update of `ExternalConstraints` and `ExternalConstraintOptions` singletons.

As a part of the PR we have also renamed `validationOf` Protobuf option to `constraintFor`. During an update to the current `base` module version, the compilation will fail in case of previous `validationOf` usage.


base/src/main/java/io/spine/validate/ExternalConstraintOptions.java
```diff
@@ -0,0 +1,162 @@
...
+     *         the field descriptor to obtain the option
+     * @param option
+     *         the option to obtain
+     * @param <T>
+     *         the type of the option value
+     * @return the {@code Optional} of option value or {@code Optional.empty()}
+     *         if there is no such an option for the field descriptor
+     */
+    public static <T> Optional<T>
+    getOptionValue(FieldContext fieldContext, GeneratedExtension<FieldOptions, T> option) {
```
armiol: As discussed, let's try to optimize the flow.

1. Find all in-place options and collect validators for them.
2. Find all external constraint options and append the external validators to the collection of validators from p.1.
3. Perform the validation with each of the validators.

yuri-sergiichuk: As discussed vocally, let's check again the current flow after the current PR, while the mentioned optimization is not obvious right now.

Also, the current validation flow relies on by-field validation and making the described change requires reconsidering the whole validation framework underneath functionality.

yuri-sergiichuk:
@armiol PTAL again.

PR #405, opened 2019-04-30
Fix ErrorProne check crash

dmdashenkov:
The `UseValidatingBuilders` ErrorProne check was found to crash when calling a method on a message of a generic type:
```java
<T extends Message> void Descriptor descriptorFromInstance(T instance) {
    return instance.getDescriptorForType();
}
```
In the example above, we fail to extract a package name from `T` and fail.

Now, we only accept `final` types. This covers also the cases of custom anonymous classes etc.


dmdashenkov:
@armiol, PTAL.

tools/smoke-tests/error-prone-checker-tests/src/test/java/io/spine/test/tools/check/VariousValidMessageUsages.java
```diff
@@ -0,0 +1,51 @@
...
+    @Test
+    @DisplayName("pass")
+    void pass() {
+        assertTrue(true, "This class must compile in order for the test to pass.");
+        Empty empty = Empty.getDefaultInstance();
+        assertEquals(generic(empty), empty);
+    }
+
+    /**
+     * Usage or {@code toBuilder()} methods with generic types is allowed, since there is no
```
armiol: `of`.

tools/smoke-tests/error-prone-checker-tests/src/test/java/io/spine/test/tools/check/VariousValidMessageUsages.java
```diff
@@ -0,0 +1,51 @@
...
+     * abstract base with the {@code toVBuilder()} method.
+     */
+    @SuppressWarnings("unused")
```
armiol: I also do not understand the suppression. The `value` is used. The method is called. The `toBuilder()` return result is used.

Also, you seem to comment on a different issue. Or is this a method documentation?

PR #413, opened 2019-05-14
Add missing Gradle Java plugin configuration names.

serhii-lekariev:
This PR adds more Gradle Java plugin configuration names to allow `io.spine.tools.gradle.project.Dependant` class users to call `depend(ConfigurationName configurationName, String notation)` with more options to choose from.


serhii-lekariev:
@dmdashenkov, PTAL.

PR #415, opened 2019-05-15
API comments revised and edited in protobuf, reflect, and string packages

meizabel:
`protobuf`, `reflect`, and `string` package API comments fixes.


base/src/main/java/io/spine/protobuf/MessageField.java
```diff
@@ -46,7 +46,7 @@
     private static final String GETTER_METHOD_PREFIX = "get";
 
     /**
-     * By convention underscore is used for separating words in field names
+     * By convention underscore it is used for separating words in field names
```
alexander-yevsyukov: This does not read to me. Under “underscore” we mean this character `_`. It may be: “By convention, the underscore character is used...”

PR #418, opened 2019-05-16
Recognize ID type by field descriptor

alexander-yevsyukov:
This PR:
  * Extends `Identifier.Type` with an ability to match itself to a `FieldDescriptor`.
  * Removes `MessageField` class, which is no longer used by `core-java`.


alexander-yevsyukov:
@armiol, @dmdashenkov, PTAL.

base/src/main/java/io/spine/base/Identifier.java
```diff
@@ -317,6 +322,42 @@ private Any pack() {
...
+                               return false;
+                           }
+                           if (idType == Type.MESSAGE) {
+                               @SuppressWarnings("unchecked") // safe since it's Message type.
+                               TypeUrl messageType = TypeUrl.of(
+                                       (Class<? extends Message>) idClass);
+                               TypeUrl fieldType = TypeUrl.from(f.getMessageType());
+                               return fieldType.equals(messageType);
+                           }
+                           return true;
```
dmdashenkov: Please split into several separate `filter` calls for better readability.

base/src/main/java/io/spine/base/Identifier.java
```diff
@@ -500,6 +567,8 @@ Message fromMessage(Message message) {
 
         abstract <I> I defaultValue(Class<I> idClass);
 
+        abstract boolean matchField(FieldDescriptor field);
```
dmdashenkov: Let's use `FieldDeclaration` instead of a descriptor.

alexander-yevsyukov: As discussed vocally let's keep it as is to simplify filtering of fields.

PR #424, opened 2019-05-22
Relax id field requirement

alexander-yevsyukov:
This PR relaxes the requirements to ID field of entity state. Before this PR the field has to be named `id` or have the suffix `_id`. Now it can have any name like `key` or `author` or `whatever`.

This PR also annotates some validation-related classes as `@Immutable`.


alexander-yevsyukov:
@armiol, @serhii-lekariev, PTAL.

PR #426, opened 2019-05-23
ID field cannot be a collection

alexander-yevsyukov:
This PR makes the field declaration consider a repeated or a map field being _not_ an ID field when it comes first in the state message.


alexander-yevsyukov:
@dmdashenkov, PTAL.

PR #427, opened 2019-05-24
Add the `protobuf` configuration name.

serhii-lekariev:
This PR adds `protobuf` to the list of Gradle configuration names.


serhii-lekariev:
@dmdashenkov, PTAL.

serhii-lekariev:
@dmdashenkov, PTAL again.

tools/plugin-base/src/main/java/io/spine/tools/gradle/ConfigurationName.java
```diff
@@ -41,84 +31,125 @@
...
+     * <p>Users cannot add dependencies directly to this configuration. However, this configuration
+     * may be resolved.
+     */
+    testRuntimeClasspath,
 
-    ConfigurationName(String value) {
-        this.value = value;
-    }
+    /**
+     * {@code protobuf} extension block exposed by the Gradle Protobuf plugin.
```
dmdashenkov: That's not the case. Those two elements of the build model have the same name, though are independent. 

What this doc should say is that this is a configuration, which allows compiling `.proto` files from the dependencies. Also, please mention that the users should use `compile` when it is only required to add some Protobuf definition into the Protobuf compilation "classpath".

PR #429, opened 2019-05-30
Fix compiled Protobuf classes publication

dmdashenkov:
Since recently, the `base` JAR was generated with two versions of Protobuf classes. This bug caused compile-time errors in downstream repositories when compiling with `javac` of version 8.

This PR fixes the JAR exclusion process.


dmdashenkov:
@armiol, PTAL.

PR #434, opened 2019-06-07
Extendable validation exception

dmdashenkov:
This PR makes the `ValidationException` extensible and provides it with some new convenience API


base/src/main/java/io/spine/validate/option/PatternConstraint.java
```diff
@@ -49,12 +50,12 @@
...
                 values.stream()
                       .filter(value -> !value.matches(regex))
-                      .map(value -> newViolation(fieldValue))
+                      .map(value -> newViolation(fieldValue, value))
                       .collect(toImmutableList());
         return violations;
     }
 
-    private ConstraintViolation newViolation(FieldValue<String> fieldValue) {
+    private ConstraintViolation newViolation(FieldValue<String> fieldValue, String rawValue) {
```
armiol: I don't get it. The `rawValue` is the first item of `fieldValue.values` field. I would extend the API of `FieldValue` to fetch the first item instead of having a call like this — as it's potentially possible to pass something irrelevant to the `rawValue` parameter.

dmdashenkov:
@armiol, after out discussion, PTAL again.

PR #439, opened 2019-06-11
Fix `(range)` option boundaries

### LINKED ISSUE #436: `(range)` option does not treat boundaries properly
Currently `(range)` option enforces users to have the boundaries of the same type as the checked value, but the boundaries themselves are not always interpreted correctly.

The following boundaries always fail the validation:

```
message NumRanges {
    int64 minute = 2 [(range) = "[0..59]"];
    float degree = 3 [(range) = "[0.0..360.0)"];
}
```

The boundary type above for minute is `Integer` while the value type is `Long`.
The boundary type above for degree is `Double` but the value type is `Float`.

yuri-sergiichuk:
This PR fixes #436.

The boundaries are now cast to the actual number type that is under validation. Such a cast allows us to have a unified number model for the boundaries and apply them to various numeric types just as Java does it for automatic primitive numbers conversions.

### DIFF
--- a/base/src/main/java/io/spine/validate/ComparableNumber.java
+++ b/base/src/main/java/io/spine/validate/ComparableNumber.java
@@ -48,6 +48,11 @@ public final class ComparableNumber extends Number implements Comparable<Number>
         return new NumberText(value);
     }
 
+    /** Returns the actual wrapped number. */
+    public Number value() {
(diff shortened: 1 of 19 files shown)


yuri-sergiichuk:
@armiol PTAL.

PR #448, opened 2019-07-04
Slow tests annotation

dmdashenkov:
The repository, as well as in the others, contains a number of slow functional tests. When building the project, it is not always necessary to run those tests.

This PR introduces the `@SlowTest` annotation. The annotation marks slow test cases/classes in order for the test runner to exclude them from the default build process.

The respective [Gradle configuration](https://github.com/SpineEventEngine/config/pull/76) is added to the `config` repository in order to apply it where it's needed.


dmdashenkov:
@armiol, PTAL.

PR #452, opened 2019-07-09
 Convert ValidationExceptions into Errors with ValidationError

dmdashenkov:
`Errors` utility converts Java's `Throwable`s into `spine.base.Error`s. 

If a given `Throwable` was caused by a constraint violation, the resulting `Error` should reflect that.

### DIFF
--- a/base/src/main/java/io/spine/base/Errors.java
+++ b/base/src/main/java/io/spine/base/Errors.java
@@ -20,6 +20,8 @@
 
 package io.spine.base;
 
+import io.spine.validate.ValidationException;
+
 import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.base.Strings.nullToEmpty;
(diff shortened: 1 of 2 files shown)


dmdashenkov:
@armiol, PTAL.

base/src/main/java/io/spine/base/Errors.java
```diff
@@ -84,4 +70,25 @@ public static Error causeOf(Throwable throwable, int errorCode) {
         Error.Builder error = toBuilderCauseOf(throwable).setCode(errorCode);
         return error.build();
     }
+
+    private static Error.Builder toBuilderCauseOf(Throwable throwable) {
+        return toErrorBuilder(getRootCause(throwable));
+    }
+
+    private static Error.Builder toErrorBuilder(Throwable throwable) {
```
armiol: As agreed in person, let's document this method.

PR #455, opened 2019-07-22
Performance Improvements

armiol:
This PR improves the performance of `base` classes placing most of its firepower onto the validation framework.

Highlights:

- `FieldAwareMessage` mixin introduced to allow the framework-specific `Message`s to define how to read their values. It helps to avoid native Protobuf reflective calls when validating messages.
- Creation of an `ImmutableList` or `ImmutableSet` is costly if performed ~29.000 times. Now it is optimized where the performance impact was especially noticeable.
- `AnyPacker.unpack` is optimized by re-using the default instance of the unpacked `Message`. To do that we had to avoid the native Protobuf `Any.unpack`, which creates an extra instance and feeds it to GC immediately.
- `AnyPacker.unpack` is fixed and now checks the type URL to match the one defined in the `.class` declaration of the `Message` when unpacking objects.
- `hashCode()` and `equals()` are defined for `AbstractOption` as it is used with the collections (BTW, huge impact!).


base/src/main/java/io/spine/protobuf/Messages.java
```diff
@@ -38,6 +42,14 @@
     /** The name of a message builder factory method. */
     public static final String METHOD_NEW_BUILDER = "newBuilder";
 
+    /**
+     * The cache of the default instances per {@link Message} class.
+     *
+     * <p>Creates and caches objects in a lazy mode.
+     */
+    private static final
+    LoadingCache<Class<? extends MessageLite>, MessageLite> defaultInstances = loadingCache(1_000);
```
dmdashenkov: Why use `MessageLite` instead of `Message`?

base/src/main/java/io/spine/type/Type.java
```diff
@@ -42,6 +44,20 @@
...
+     * <p>Saves the memory by limiting the maximum number of cached objects.
+     *
+     * @implNote {@link com.google.common.cache.LoadingCache LoadingCache} implementation
+     *         would make dealing with the exceptions a bit more difficult to maintain and read.
+     *         Which is why a straightforward get-put approach with a plain {@code Cache}
+     *         implementation is used.
+     */
+    private static final Cache<String, Class<?>> knownClasses = CacheBuilder.newBuilder()
+                                                                            .maximumSize(1_000)
+                                                                            .build();
```
dmdashenkov: Please use a loader instead of manually creating instances in `javaClass()`.

Also, I would consider making the `1_000` a constant, or better yet, a configurable property, since different environments may require different cache sizes.

base/src/main/java/io/spine/validate/FieldAwareMessage.java
```diff
@@ -0,0 +1,78 @@
...
+
+/**
+ * The mixin for {@link Message Message}s which are aware of their fields and are able to provide
+ * their values with no reflection {@linkplain Message#getField(FieldDescriptor) used by Protobuf}.
+ *
+ * <p>This mixin is an experimental part of the framework and may be changed or removed in future.
+ */
+@GeneratedMixin
+@Experimental
+public interface FieldAwareMessage extends Message {
```
dmdashenkov: Please mark this type as `Internal`. Being an interface, it exposes all its methods into `public` scope.

base/src/main/java/io/spine/validate/FieldAwareMessage.java
```diff
@@ -0,0 +1,78 @@
...
+    default Object readValue(FieldDescriptor field) {
+        return getField(field);
+    }
+
+    /**
+     * A test-only method that checks that the implementation of {@link #readValue(FieldDescriptor)}
+     * gives the same results as the {@link Message#getField(FieldDescriptor)}.
+     */
+    @VisibleForTesting
+    default boolean checkFieldsReachable() {
```
dmdashenkov: This looks like a test case. Maybe, we could make it into an abstract test class instead of adding it into production code.

base/src/test/java/io/spine/validate/given/FieldAwareMessageTestEnv.java
```diff
@@ -0,0 +1,341 @@
...
+        }
+
+        @Override
+        public String toString() {
+            return delegate.toString();
+        }
+
+        @Override
+        public void writeTo(CodedOutputStream output) throws IOException {
+            // no-op.
```
dmdashenkov: Why don't we delegate this and all the following methods to the `delegate`?

PR #461, opened 2019-08-22
Allow to force individual dependency versions in `Dependant`

dmitrykuzmin:
This PR is a part of https://github.com/SpineEventEngine/bootstrap/pull/30.

The `Dependant` is expanded with `force(...)` and `removeForcedDependency(...)` methods which allow to force the required dependency version and remove the forced dependency from the project respectively.

Spine version is updated to `1.0.2`.

### DIFF
--- a/tools/plugin-base/src/main/java/io/spine/tools/gradle/project/Dependant.java
+++ b/tools/plugin-base/src/main/java/io/spine/tools/gradle/project/Dependant.java
@@ -62,12 +62,28 @@ public interface Dependant {
     void force(Artifact artifact);
 
     /**
+     * Forces all project configurations to fetch the particular dependency version.
+     *
+     * @param notation
+     *         the dependency spec, e.g. {@code com.google.protobuf:protoc:3.9.0}
+     */
+    void force(String notation);
+
+    /**
      * Removes a forced dependency from resolution strategies of all project configurations.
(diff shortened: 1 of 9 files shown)


dmitrykuzmin:
@dmdashenkov PTAL.

tools/plugin-base/src/main/java/io/spine/tools/gradle/project/Dependant.java
```diff
@@ -53,6 +53,24 @@
...
 
+    /**
+     * Forces all project configurations to fetch the particular dependency version.
+     *
+     * @param dependency
+     *         the forced dependency
+     * @param version
+     *         the required version of the dependency
+     */
+    void force(Dependency dependency, String version);
```
dmdashenkov: We have a composite type for a dependency with a version — `Artifact`. Please use it as the single param type instead.

version.gradle
```diff
@@ -25,7 +25,7 @@
  * as we want to manage the versions in a single source.
  */
 
-final def SPINE_VERSION = '1.0.1'
+final def SPINE_VERSION = '1.0.2'
```
dmdashenkov: Did you discuss this with Alex? If not, please use `-SNAPSHOT`.

PR #462, opened 2019-08-23
Fix `Stringifiers.toString(...)` not processing enum values

dmitrykuzmin:
This PR fixes `Stringifiers.toString(...)` throwing an exception on receiving an `enum` value.

The `enum` values were not supported because the `EnumStringifier` was actually abstract and was intended for inheritance only. It was done so because the `EnumStringifier` is a `SerializableStringifier` and all serializable stringifiers are persisted as singletons. But it's impossible to persist an arbitrary enum stringifier as a singleton.

This PR makes an `EnumStringifier` a concrete class which is still a `SerializableStringifier` but is not persisted as a singleton (but rather as an ordinary serializable object).

### DIFF
--- a/base/src/test/java/io/spine/reflect/TypesTest.java
+++ b/base/src/test/java/io/spine/reflect/TypesTest.java
@@ -25,18 +25,21 @@ import com.google.common.reflect.TypeToken;
 import com.google.common.testing.NullPointerTester;
 import com.google.protobuf.Message;
 import com.google.protobuf.StringValue;
+import io.spine.reflect.given.TypesTestEnv.ListOfMessages;
+import io.spine.reflect.given.TypesTestEnv.TaskStatus;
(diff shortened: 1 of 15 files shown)


dmitrykuzmin:
@armiol PTAL.

base/src/test/java/io/spine/reflect/TypesTest.java
```diff
@@ -52,21 +55,41 @@
...
-    @DisplayName("create list type")
+    @DisplayName("create a list type")
     void createListType() {
         Type type = listTypeOf(String.class);
         Type expectedType = new TypeToken<List<String>>(){}.getType();
         assertEquals(expectedType, type);
     }
 
+    @Test
+    @DisplayName("tell if the type is a enum class")
```
armiol: `an enum class`.

PR #463, opened 2019-08-27
Update Gradle and Truth

dmdashenkov:
In this PR we update the dependencies versions:
 - Google Truth is updated to `1.0`;
 - Gradle is updated to `5.6.1`.


dmdashenkov:
@armiol, PTAL.

PR #467, opened 2019-09-04
Path of fields

alexander-yevsyukov:
This PR introduces the `Field` class which aims to encapsulate working with field paths and thus replace utilities from `FieldPaths` class.

### DIFF
--- a/base/src/main/java/io/spine/base/FieldPaths.java
+++ b/base/src/main/java/io/spine/base/FieldPaths.java
@@ -121,7 +121,7 @@ public final class FieldPaths {
      *         the message from which to obtain a value of the field
      * @return the value of the field
      * @throws IllegalArgumentException if the passed message does not define such a field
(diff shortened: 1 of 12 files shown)


base/src/main/java/io/spine/base/FieldPaths.java
```diff
@@ -59,25 +61,46 @@ private FieldPaths() {
...
-        checkArgument(!stringPath.isEmpty(), "Path must not be empty.");
+    @Deprecated
+    public static FieldPath parse(String fieldPath) {
+        checkNotNull(fieldPath);
+        return doParse(fieldPath);
+    }
 
-        List<String> pathElements = dotSplitter.splitToList(stringPath);
-        return fromElements(pathElements);
+    static FieldPath doParse(String fieldPath) {
```
dmdashenkov: Instead of adding `doSmth` methods, let's move the logic into the `Field` altogether. This way, when the time comes, we can just delete the `FieldPaths.java` without any further migration.

PR #469, opened 2019-09-06
GitHub Actions

dmdashenkov:
This PR introduces GitHub Actions into `base`.

For now, we use Actions to run tests on Windows for PRs.

### DIFF
--- a/.idea/dictionaries/common.xml
+++ b/.idea/dictionaries/common.xml
@@ -43,7 +43,6 @@
       <w>processmanager</w>
       <w>procman</w>
       <w>proto's</w>
-      <w>protobuf</w>
       <w>protos</w>
       <w>sfixed</w>
       <w>stderr</w>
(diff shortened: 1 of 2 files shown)


dmdashenkov:
@dmitrykuzmin, PTAL.

.idea/dictionaries/common.xml
```diff
@@ -43,7 +43,6 @@
       <w>processmanager</w>
       <w>procman</w>
       <w>proto's</w>
-      <w>protobuf</w>
```
dmitrykuzmin: Let's add this to `config`'s level dictionary instead of removing here. Especially since it already contains words like `proto's`.

dmdashenkov: This change was caused by a change in `config`. I suppose, IDEA now knows this word by default.

PR #472, opened 2019-09-16
Remove Mockito

dmitrykuzmin:
This PR removes Mockito from test suites of the project.

Also, the Mockito dependency is no longer exposed as an API by the `testlib`. The Mockito-based utility  `Spy` is removed.

The work of `ErrorProneChecksPlugin` is slightly tweaked, it does not rely on Model Compiler dependency present anymore.

Spine version advances to `1.1.1-SNAPSHOT+1`.


dmitrykuzmin:
@armiol PTAL.

tools/model-compiler/src/test/java/io/spine/tools/compiler/check/DependencyConfigurerTest.java
```diff
@@ -21,91 +21,101 @@
...
-import static org.junit.jupiter.api.Assertions.assertTrue;
-import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.Mockito.spy;
-import static org.mockito.Mockito.when;
-
-@SuppressWarnings({"CheckReturnValue", "ResultOfMethodCallIgnored"})
+import static io.spine.tools.gradle.ConfigurationName.classpath;
+import static io.spine.tools.gradle.compiler.given.Project.newProject;
+
+// TODO:2019-09-15:dmytro.kuzmin: This test configures the project with real dependencies and
```
armiol: Please transform this into a class-level comment with an `@implNote` and a GitHub issue.

tools/model-compiler/src/test/java/io/spine/tools/compiler/check/DependencyConfigurerTest.java
```diff
@@ -21,91 +21,101 @@
...
+@SlowTest
 @DisplayName("DependencyConfigurer should")
 class DependencyConfigurerTest {
 
-    private static final String STUB_VERSION = "versionStub";
-
-    private Configuration annotationProcessorConfig;
-    private DependencyConfigurer helper;
-    private DependencyConfigurer helperMock;
+    private static final String SPINE_VERSION = "1.1.0";
```
armiol: Can we read this property from the `version.gradle` file? It will never be obvious this file references the version that has to be updated each release.

Besides, the `releases` repository is also specific to this version. The `SNAPSHOT` versions won't resolve properly.

dmitrykuzmin:
@armiol PTAL again.

PR #475, opened 2019-09-25
Dart code generation

dmdashenkov:
Introducing Protobuf Dart Gradle plugin.

---
The new plugin manages code generation config for Dart, as well as generating custom Dart code.
In particular, the plugin generated `types.dart` files — type registries for generated message types.

`types.dart`:
 - describes all the types which are compiled in this module, ignoring imports and dependencies;
 - maps Dart messages to type URLs;
 - maps type URLs to `BuilderInfo`s — the Protobuf-specific type descriptors.

For generating `types.dart` files, the plugin defines a simple library which allows describing Dart language structured in an object-oriented way in Java. This approach may later be used for further Dart codegen (e.g. for validation code).

---
Also, this PR refactors the `TaskName` enum. Previously, it contained all the names of Gradle tasks used in the plugin. Now, `TaskName` is an interface implemented in a number of case-specific enums.

Also, in this PR we delete the `TypeRef` concept as it is no longer used.


dmdashenkov:
@armiol, PTAL.
@dmitrykuzmin, PTAL.

dmitrykuzmin:
LGTM

dmdashenkov:
@armiol, PTAL.

dmdashenkov:
@alexander-yevsyukov, PTAL.

PR #479, opened 2019-10-04
Dart code generation improvements

dmdashenkov:
This PR improves the code generation for Dart type registries.

Previously, the generated registries provided two maps as a public API. The issue with this approach is that there was no handy way to transfer the knowledge about types — the API users would have to pass around two maps all the time or construct own abstractions.

Now, the registry provides the same information in form of an object. The compile-time type of the object is `dynamic`. This allows the code generator not to have a dependency on another Dart library which would declare the contract explicitly (i.e. as a class). The type registry is an internal API of the framework.

Also, this PR improves test coverage of the `proto-dart-plugin` and gets rid of some dead code (see `io.spine.code.properties`).


dmdashenkov:
@armiol, PTAL.

PR #480, opened 2019-10-04
Custom resource class loader

alexander-yevsyukov:
This PR allows the `Resource` instance to be created with a custom `ClassLoader`. By default, `Resource.class.getClassLoader()` will be used.

### DIFF
--- a/base/src/main/java/io/spine/io/Resource.java
+++ b/base/src/main/java/io/spine/io/Resource.java
@@ -77,12 +77,13 @@ public final class Resource {
      *
      * @param path
      *         the path to the resource file, relative to {@code contextClass}
-     * @param classLoader
- *         the class relative to which the resource is referenced */
-    public static Resource file(String path, ClassLoader classLoader) {
-        checkNotNull(classLoader);
+     * @param customLoader
+     *         the class relative to which the resource is referenced
+     */
+    public static Resource file(String path, ClassLoader customLoader) {
+        checkNotNull(customLoader);
         checkNotEmptyOrBlank(path);
-        return new Resource(path, classLoader);
+        return new Resource(path, customLoader);
     }
 
     private boolean loaderCustomized() {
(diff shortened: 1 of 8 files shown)


base/src/main/java/io/spine/io/Resource.java
```diff
@@ -48,20 +52,65 @@
 public final class Resource {
 
     private final String path;
+    private final @Nullable ClassLoader classLoader;
```
dmdashenkov: Please update the class-level doc.

base/src/main/java/io/spine/io/Resource.java
```diff
@@ -48,20 +52,65 @@
...
+        return new Resource(path, null);
+    }
+
+    /**
+     * Creates a new reference to a resource at the context of the given class.
+     *
+     * @param path
+     *         the path to the resource file, relative to {@code contextClass}
+     * @param classLoader
+ *         the class relative to which the resource is referenced */
```
dmdashenkov: Please fix the alignment.

base/src/main/java/io/spine/io/Resource.java
```diff
@@ -83,8 +132,7 @@ public boolean exists() {
      * @return the resource URL
      */
     public URL locate() {
-        URL resource = classLoader().getResource(path);
-        checkFound(resource);
+        URL resource = toUrl();
```
dmdashenkov: Please inline `toUrl()` into `locate()`. There is no point in having two methods.

base/src/main/java/io/spine/io/Resource.java
```diff
@@ -48,20 +52,65 @@
...
+
+    private String classLoaderDiag() {
+        ClassLoader loader = classLoader();
+        String fmt = inContext() ? "the assigned class loader `%s`." : "the class loader `%s`.";
+        return format(fmt, loader);
+    }
+
+    private URL toUrl() {
+        URL url = findUrl();
+        if (url == null) {
```
dmdashenkov: Let's just not allow `null` value. Instead, use `Resource.class.getClassLoader()` as the default. Please call `file(String, ClassLoader)` from `file(String)` with the default class loader. Then, we won't have to use `firstNonNull`.

PR #483, opened 2019-10-15
Dart validators for all scalar fields

dmdashenkov:
In this PR we further the code generation for Dart Protobuf message validators.

Now, all the non-recursive scalar field validation is run.

For `string`, `bytes`, `message`, and `enum` fields, the `(required)` option is applied.
For `string` fields, the `(pattern)` option is applied.
For numeric fields, `(min)`, `(max)`, and `(range)` options are applied.

Also, now we have tests for the generated code. See `tools/smoke-tests/dart-codegen-tests`.


dmdashenkov:
@armiol, PTAL.

dmdashenkov:
@dmitrykuzmin, PTAL.

dmdashenkov:
@dmitrykuzmin, PTAL again.

PR #486, opened 2019-10-22
Improvements for precondition checks

alexander-yevsyukov:
This PR moves methods for checking arguments and instances of `Message` to corresponding utility classes. Previously, most of the checks were in the `Validate` class. This made unnecessary dependency on the `validate` package. This PR fixes that.

Methods of the `Validate` class were deprecated in favor of corresponding counterparts from `Messages` and `Preconditions2` utility classes.

Updated `Preconditions2` class now has pairs of methods that has `Arg` and `State` suffixes that throw `IllegalArgumentException` and `IllegalStateException`. Previously, in the `Validate` class it was always `IllegalStateException`, which was confusing.

Also message validation was simplified. Previously, it required the following construct:
```java
List<ConstraintViolations> violations = 
    MessageValidator.newInstance(msg)
    .validate();
```
Now, it's simply:
```java
List<ConstraintViolations> violations = MessageValidator.validate(msg);
```
The version was bumped to `1.1.6`.


base/src/main/java/io/spine/util/Preconditions2.java
```diff
@@ -37,31 +44,230 @@ private Preconditions2() {
...
-     * @throws IllegalArgumentException if the string is empty or blank
+     * @throws IllegalArgumentException
+     *         if the string is empty or blank
+     * @throws NullPointerException
+     *         if the passed string is {@code null}
      */
     @CanIgnoreReturnValue
-    public static String checkNotEmptyOrBlank(String stringToCheck) {
-        return checkNotEmptyOrBlank(stringToCheck, "");
+    public static String checkNotEmptyOrBlank(String str) {
```
yuri-sergiichuk: Please re-use the `checkNotEmptyOrBlank` method with 2 args here as discussed vocally.

base/src/main/java/io/spine/util/Preconditions2.java
```diff
@@ -37,31 +44,230 @@ private Preconditions2() {
...
         return stringToCheck;
     }
+
+    /**
+     * Ensures that the passed value is positive.
+     *
+     * @param value the value to check
+     * @throws IllegalArgumentException if the value is negative or zero
+     */
+    public static long checkPositive(long value) {
```
yuri-sergiichuk: I'd better reuse the `checkPositive` method with 2 args here as well.

base/src/main/java/io/spine/util/Preconditions2.java
```diff
@@ -37,31 +44,230 @@ private Preconditions2() {
...
+     *
+     * @param value
+     *         the value to check
+     * @param errorMessage
+     *         the exception message to use if the check fails
+     * @throws IllegalArgumentException
+     *         if the value is negative or zero
+     */
+    public static long checkPositive(long value, @Nullable String errorMessage) {
+        checkNotNull(errorMessage);
```
yuri-sergiichuk: I believe we should not throw the NPE on `null` error message.

base/src/main/java/io/spine/util/Preconditions2.java
```diff
@@ -37,31 +44,230 @@ private Preconditions2() {
...
+     *         will be converted to a string using {@link String#valueOf(Object)}
+     * @throws IllegalStateException
+     *         if the object is in its default state
+     * @throws NullPointerException
+     *         if the passed message is {@code null}
+     */
+    @CanIgnoreReturnValue
+    public static <M extends Message>
+    M checkNotDefaultArg(M object, @Nullable Object errorMessage) {
+        checkNotNull(object);
```
yuri-sergiichuk: Shouldn't we pass our custom error message here (and below) as well?

base/src/main/java/io/spine/util/Preconditions2.java
```diff
@@ -37,31 +44,231 @@ private Preconditions2() {
...
-        checkNotNull(stringToCheck);
-        checkNotNull(message);
-        checkArgument(!stringToCheck.isEmpty(), message);
-        String trimmed = stringToCheck.trim();
-        checkArgument(trimmed.length() > 0, message);
-        return stringToCheck;
+    public static String checkNotEmptyOrBlank(String str,
+                                              @Nullable String errorMessageTemplate,
+                                              @Nullable Object @Nullable ... errorMessageArgs) {
+        checkNotNull(str, errorMessageTemplate);
```
yuri-sergiichuk: Shouldn't we pass `errorMessageArgs` as well?

base/src/main/java/io/spine/util/Preconditions2.java
```diff
@@ -37,31 +44,231 @@ private Preconditions2() {
...
+     *          if the passed message is {@code null}
+     */
+    @CanIgnoreReturnValue
+    @SuppressWarnings("OverloadedVarargsMethod")
+    public static <M extends Message>
+    M checkNotDefaultState(M object,
+                           @Nullable String errorMessageTemplate,
+                           @Nullable Object @Nullable ... errorMessageArgs) {
+        checkNotNull(object);
+        boolean value = isNotDefault(object);
```
yuri-sergiichuk: why do we need `value` as a separate variable here?

base/src/main/java/io/spine/validate/Validate.java
```diff
@@ -287,6 +306,8 @@ public static void checkValid(Message message) throws ValidationException {
      *         the new state of the message
      * @param <M>
      *         the type of the message
+     * @throws ValidationException
+     *          the the value transition is not valid
```
yuri-sergiichuk: remove extra `the`

base/src/test/java/io/spine/util/Preconditions2Test.java
```diff
@@ -20,44 +20,145 @@
...
         @Test
         @DisplayName("empty")
         void emptyString() {
             assertThrowsOn("");
+            assertThrowsWithMessage("", randomString());
         }
 
         @Test
         @DisplayName("blank")
         void blankString() {
```
yuri-sergiichuk: Maybe `@ParameterizedTest` will do the job better?

base/src/test/java/io/spine/util/Preconditions2Test.java
```diff
@@ -20,44 +20,145 @@
...
             assertThrowsOn("   ");
+            assertThrowsWithMessage("   ", randomString());
+        }
+    }
+
+    @Nested
+    @DisplayName("Check that a value is positive")
+    class PositiveValue {
+
+        void assertThrowsOn(long value) {
```
yuri-sergiichuk: Why aren't these methods private?

base/src/test/java/io/spine/util/Preconditions2Test.java
```diff
@@ -20,44 +20,145 @@
...
+            assertThrows(IllegalStateException.class,
+                         () -> checkNotDefaultState(StringValue.getDefaultInstance()));
+        }
+
+        @Test
+        @DisplayName("return non-default value on check")
+        void returnValue() {
+            StringValue nonDefault = newUuidValue();
+            assertEquals(nonDefault, checkNotDefaultArg(nonDefault));
+            assertEquals(nonDefault, checkNotDefaultState(nonDefault, "with error message"));
```
yuri-sergiichuk: shouldn't it be a separate test case? we're testing a different function

base/src/main/java/io/spine/util/Preconditions2.java
yuri-sergiichuk: So what about the `errorMessage`? Why don't we pass it to the `notNull` check?

base/src/main/java/io/spine/util/Preconditions2.java
alexander-yevsyukov: Please see in the latest code.

base/src/test/java/io/spine/util/Preconditions2Test.java
alexander-yevsyukov: With parametrized tests the resulting code would be bigger and harder to understand.

base/src/test/java/io/spine/util/Preconditions2Test.java
alexander-yevsyukov: This suite test both `state` and `arg` functions. I've added more checks, it should look clearer now.

base/src/main/java/io/spine/util/Preconditions2.java
alexander-yevsyukov: Because it's nullable.

base/src/main/java/io/spine/util/Preconditions2.java
yuri-sergiichuk: I see that in our shortcut method we are passing the template string with parameter:
```
checkNotEmptyOrBlank(str, "Non-empty and non-blank string expected. Encountered: \"%s\".", str);
```
In this case, we'll have the `notNull` message equal to `"Non-empty and non-blank string expected. Encountered: \"%s\"."`, without the parameter being substituted with `null`, because we do not pass the parameters. Am I right?

base/src/main/java/io/spine/util/Preconditions2.java
yuri-sergiichuk: So what? We still pass it to the `checkArgument` even though it still might be `null`. It's just inconsistent. Why do we omit the custom error message for `notNull` checks, but pass it to the `checkArgument`? We do the same in almost every method in this class.

Let's define some consistent behavior and use it across all methods. If we think that we should not pass the custom error message to the `checkNotNull` check, then maybe we should explicitly state it.

base/src/test/java/io/spine/util/Preconditions2Test.java
yuri-sergiichuk: I was thinking of smth like:
```
        @DisplayName("empty or blank")
        @ParameterizedTest(name = "\"{arguments}\"")
        @ValueSource(strings = {"", " ", "  ", "   "})
        void emptyOrBlank(String value) {
            assertThrowsOn(value);
            assertThrowsWithMessage(value, randomString());
        }
```

PR #487, opened 2019-10-23
Pass custom messages to null checks

alexander-yevsyukov:
This PR fixes treating custom messages passed to precondition checks on `null` values. Previously, formatted messages were used only on non-null values, which could be misleading.

Version bumped to 1.1.7.

### DIFF
--- a/base/src/main/java/io/spine/util/Preconditions2.java
+++ b/base/src/main/java/io/spine/util/Preconditions2.java
@@ -54,9 +54,12 @@ public final class Preconditions2 {
      */
     @CanIgnoreReturnValue
     public static String checkNotEmptyOrBlank(String str) {
-        return checkNotEmptyOrBlank(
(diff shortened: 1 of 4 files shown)


base/src/main/java/io/spine/util/Preconditions2.java
```diff
@@ -54,9 +54,31 @@ private Preconditions2() {
      */
     @CanIgnoreReturnValue
     public static String checkNotEmptyOrBlank(String str) {
```
yuri-sergiichuk: Why don't we reuse the overloaded `checkNotEmptyOrBlank(str, errorMessage)` here?

alexander-yevsyukov: To save on formatting or array allocation for vararg if all goes well.

PR #489, opened 2019-10-28
Move Dart away

dmdashenkov:
Previously, this repository hosted a number of Dart modules. Now, as we have a [separate repository](https://github.com/SpineEventEngine/dart) for that, all those modules are moved away.


dmdashenkov:
@armiol, PTAL.

PR #490, opened 2019-10-28
Include `proto-dart-plugin`

dmdashenkov:
The `proto-dart-plugin` is now included in the build again. The plugin does not require a Dart environment and thus can stay in this repository, along with all the other Gradle plugins.

Also, this PR enables tasks added by `ProtoDartPlugin` to be triggered by `assemble`.


dmdashenkov:
@armiol, PTAL.

PR #492, opened 2019-10-28
The `(column)` option

dmitrykuzmin:
This PR addresses https://github.com/SpineEventEngine/core-java/issues/1135.

The entity columns will now be declared in Protobuf with the help of `(column)` option:
```proto
message UserProfile {
    option (entity).kind = PROJECTION;
    // ...
    int32 year_of_registration = 8 [(column) = true];    
}
```
This PR introduces a `ColumnGenPlugin` which is a part of Spine Model Compiler and which generates an `_X_WithColumns` interface for every entity defining columns. For example:
```java
interface UserProfileWithColumns extends EntityWithColumns {
    
    int getYearOfRegistration();
}
```
This interface can then be implemented by the `UserProfileProjection` to define the column values:
```java
class UserProfileProjection 
            extends Projection<UserId, UserProfile, UserProfile.Builder>
            implements UserProfileWithColumns {

     @Override
     int getYearOfRegistration() {
         return 2019;
     }
}
```

Apart from defining the `ColumnGenPlugin`, this PR declares a `(column)` option itself and also adds a number of tools needed for changes in `core` (to enable the `(column)` option support).

The Spine version advances to `1.1.10`.


dmitrykuzmin:
@armiol, @dmdashenkov PTAL. The coverage is low because the new and refactored code generation code is invoked from the smoke tests.

base/src/main/java/io/spine/base/EntityWithColumns.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+package io.spine.base;
+
+import io.spine.annotation.Internal;
+
+/**
+ * A marker interface for entities that manually implement their columns.
+ *
+ * <p>Is used by Spine routines and should never be implemented directly, but rather through the
```
armiol: Let it be just `<p>Used ...`.

base/src/main/java/io/spine/reflect/Methods.java
```diff
@@ -0,0 +1,61 @@
...
+     * @throws IllegalArgumentException
+     *         if the target is not of the type that declares the given method
+     * @throws IllegalStateException
+     *         if an exception is thrown during the method invocation
+     */
+    public static Object setAccessibleAndInvoke(Method method, Object target) {
+        try {
+            method.setAccessible(true);
+            Object result = method.invoke(target);
+            method.setAccessible(false);
```
armiol: Shouldn't that be in `finally`?

base/src/main/proto/spine/options.proto
```diff
@@ -232,7 +232,54 @@ extend google.protobuf.FieldOptions {
...
+    //    int getYearOfRegistration();
+    // }
+    //
+    // // An entity class which implements the `UserProfileWithColumns` interface.
+    // class UserProfileProjection
+    //            extends Projection<UserId, UserProfile, UserProfile.Builder>
+    //            implements UserProfileWithColumns {
+    //      ...
+    //
+    //      @Override
```
armiol: I am a bit confused by the Java examples in the language-agnostic Protobuf definition of the column.

Can we please think of some other place to list the Java examples and leave just the generic definition of what a column is. Think of this Protobuf file as of such used by JavaScript and C++ clients as well.

tools/model-compiler/src/main/java/io/spine/tools/gradle/compiler/ColumnGenPlugin.java
```diff
@@ -0,0 +1,151 @@
...
+
+        _debug().log("Column generation phase initialized with tasks: `%s`, `%s`.",
+                     mainTask, testTask);
+    }
+
+    private static Action<Task> createAction(Project project,
+                                             Supplier<FileSet> files,
+                                             Supplier<String> targetDirPath,
+                                             Supplier<String> protoSrcDir) {
+        return new GenAction(project, files, targetDirPath, protoSrcDir);
```
dmdashenkov: Seems like we could just call the constructor directly.

tools/model-compiler/src/main/java/io/spine/tools/gradle/compiler/ColumnGenPlugin.java
```diff
@@ -0,0 +1,151 @@
...
+
+    @Override
+    protected Supplier<File> testDescriptorFile(Project project) {
+        return () -> getTestDescriptorSet(project);
+    }
+
+    /**
+     * Does the code generation.
+     */
+    private static class GenAction extends CodeGenerationAction {
```
dmdashenkov: Please extract as a top-level type and document.

tools/plugin-base/src/main/java/io/spine/tools/gradle/ModelCompilerTaskName.java
```diff
@@ -43,6 +43,16 @@
      */
     generateTestRejections,
 
+    /**
+     * Generates the helper types for declaring columns in the {@code main} scope.
+     */
+    generateColumns,
```
dmdashenkov:
```suggestion
    generateColumnInterfaces,
```

tools/plugin-base/src/main/java/io/spine/tools/gradle/ModelCompilerTaskName.java
```diff
@@ -43,6 +43,16 @@
...
+
+    /**
+     * Generates the helper types for declaring columns in the {@code test} scope.
+     */
+    generateTestColumns,
```
dmdashenkov:
```suggestion
    generateTestColumnInterfaces,
```

dmitrykuzmin:
@armiol, @dmdashenkov PTAL again.

base/src/main/java/io/spine/code/proto/SourceFile.java
```diff
@@ -103,6 +103,14 @@ public FileDescriptor descriptor() {
         return descriptor;
     }
 
+    public List<MessageType> topLevelMessages() {
```
dmdashenkov: Please document.

tools/model-compiler/src/main/java/io/spine/tools/compiler/gen/Writer.java
```diff
@@ -0,0 +1,73 @@
...
+import io.spine.logging.Logging;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+/**
+ * A writer of a {@link GeneratedTypeSpec}.
+ */
+final class Writer implements Logging {
```
dmdashenkov: The meaning of `Writer` in `io.spine.tools.compiler.gen` is not clear enough to me. Please make the name more descriptive.

PR #493, opened 2019-11-06
Allow skipping validating builder generation

dmdashenkov:
In this PR we allow disabling `ValidatingBuilder` generation in the model compiler plugin.

This is needed for the users who don't use Spine framework as a whole, but only the tools from this repository.

In order to disable the feature, configure the plugin as follows:
```gradle
modelCompilder {
    generateValidatingBuilders = false
}
```

The value of `generateValidatingBuilders` is `true` by default.

### DIFF
--- a/tools/protoc-api/src/main/proto/spine/tools/protoc/config.proto
+++ b/tools/protoc-api/src/main/proto/spine/tools/protoc/config.proto
@@ -24,7 +24,7 @@ message SpineProtocConfig {
     // Methods generation tasks.
     AddMethods add_methods = 2;
 
-    // Should the plugin skip generating `ValidatingBuilder` implementations.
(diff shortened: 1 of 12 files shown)


dmdashenkov:
@armiol, PTAL.

tools/protoc-api/src/main/proto/spine/tools/protoc/config.proto
```diff
@@ -23,6 +23,14 @@ message SpineProtocConfig {
 
     // Methods generation tasks.
     AddMethods add_methods = 2;
+
+    // Should the plugin skip generating `ValidatingBuilder` implementations?
```
armiol: Determines if the generation of `ValidatingBuilder` should be skipped.

PR #496, opened 2019-11-11
Update dependency versions

dmdashenkov:
In this PR we update third-party dependency versions. This includes the new Gradle 6.0.

The version of the lib is updated to `1.2.1`.


dmdashenkov:
@dmitrykuzmin, PTAL.

PR #497, opened 2019-11-12
Generate Java validation code at compile time

dmdashenkov:
This PR starts the process of moving Java Protobuf validation code to build stage.

Before this PR, all the validation work was done at runtime. This included assembling metadata, scanning classpath for custom validation strategies, and, last but least, destructing messages into fields via costly reflective calls.

Starting from now, we perform most of the validation work at build time. The runtime only invokes the functions for validating a specific message type.

### Performance

As measured on load tests in `core-java`, the performance impact of validation has been cut significantly. 
Here are the measurements for a test which dispatches thousands of messages to several entities:

 Running mode | CPU time, ms | GC calls per test | Max RAM usage, MB 
:--------------:|:------------:|:--------:|:-----------------:
 Old runtime | 2000-3000 | n/a | n/a 
 New runtime | 1700-2500 | ~2 | ~1800 
 Generated | 1200-2000 | ~1 | ~1600 

### API status

Currently, this feature is in beta. The old runtime-only validation is still functional. To switch to the new implementation, configure the code generation in Gradle:
```gradle
modelCompiler {
    generateValidation = true
}
```
The `base` types are now supplied with the generated validation, as will be `time` and `core` types after those modules are migrated to the new version of `base`.

## Interop

When invoking generated validation, the runtime validation may still be used for types for which the validation code was not generated. Note that the vice-versa interop does not work, i.e. runtime validation never invokes generated validation code.

### Limitations

It is planned that the generated validation will cover all the runtime validation features and effectively replace it in future releases. For now, external validation rules (i.e. `(validation_for)`) are not supported.

### Behaviour changes

Previously, when validating a `string` field, the value was checked upon the `pattern` constraint even if the string was empty. This was inconsistent with all the other validation. Now, the string value is only checked if it's not empty.


dmdashenkov:
@armiol, PTAL.

dmdashenkov:
@alexander-yevsyukov, PTAL.

tools/protoc-plugin/src/main/java/io/spine/tools/protoc/validation/ValidatorGenerator.java
```diff
@@ -0,0 +1,102 @@
...
+
+import java.util.Collection;
+
+import static io.spine.tools.protoc.InsertionPoint.builder_scope;
+import static io.spine.tools.protoc.InsertionPoint.class_scope;
+import static io.spine.tools.protoc.TypeParameters.empty;
+import static io.spine.tools.protoc.iface.MessageImplements.implementInterface;
+
+/**
+ * The message validation code generator.
```
alexander-yevsyukov: How about something like “Generates the code for validating messages according to constraints defined in the message type.”? 4 nouns coming together are not that easy to understand.

tools/protoc-plugin/src/main/java/io/spine/tools/protoc/validation/ValidatorGenerator.java
```diff
@@ -0,0 +1,102 @@
...
+ *
+ * <p>Generates code which validates message fields upon the constraints, as well as the API which
+ * exposes validation to the user of the message class.
+ */
+public final class ValidatorGenerator extends SpineProtoGenerator {
```
alexander-yevsyukov: I think we need to re-visit the name `SpineProtoGenerator` (including its description). Now it reads as if we were generating `.proto` files, which is not the case.

`SpineCodeGenerator` or simply `CodeGenerator` would work fine, assuming we mention that we do take `.proto` code as the input.

alexander-yevsyukov: While the prefix `Spine` in `SpineProtocConfig` serves for the purpose of saying that the configuration is specific to the Spine framework, having the same prefix for the code generation classes looks excessive. We already have this notion in the package name, and we don't really need to highlight the fact again (as for `SpineProtocConfig`).

tools/protoc-plugin/src/main/java/io/spine/tools/protoc/validation/package-info.java
```diff
@@ -0,0 +1,28 @@
...
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+@CheckReturnValue
```
alexander-yevsyukov: Please give brief documentation of the package.

tools/smoke-tests/generated-validation-tests/src/test/java/io/spine/test/tools/validate/NumberConstraintTest.java
```diff
@@ -0,0 +1,102 @@
...
+
+import static com.google.common.truth.Truth.assertThat;
+
+@DisplayName("Number boundaries constraints should be compiled so that")
+class NumberConstraintTest {
+
+    @Test
+    @DisplayName("min value is checked")
+    void min() {
+        assertViolation(InterestRate
```
alexander-yevsyukov: How about this layout:

```java
assertViolation(
    InterestRate
        .newBuilder()
        .setPercent(-3)
        .buildPartial(),
    "greater than 0.0"
);
```
This way we have constant position for all the properties, and for the types.

tools/validation-generator/src/main/java/io/spine/tools/validate/code/Expression.java
```diff
@@ -0,0 +1,67 @@
...
+import com.google.errorprone.annotations.FormatString;
+import com.squareup.javapoet.CodeBlock;
+
+import static java.lang.String.format;
+
+/**
+ * A Java code structure which may yield a value.
+ *
+ * @param <R>
+ *         the type of the value yielded by this expression; the type parameter is unused and serves
```
alexander-yevsyukov: This sounds strange. Perhaps, a couple of examples would help the reader to grasp the idea behind this?

tools/validation-generator/src/main/java/io/spine/tools/validate/code/ViolationTemplate.java
```diff
@@ -0,0 +1,154 @@
...
+        this.type = builder.type;
+        this.field = builder.field;
+        this.nestedViolations = builder.nestedViolations;
+    }
+
+    @Override
+    public CodeBlock toCode() {
+        CodeBlock.Builder builder = CodeBlock
+                .builder()
+                .add("$T.newBuilder()", ConstraintViolation.class)
```
alexander-yevsyukov: Here we use `ConstraintViolation.class` which looks as the duplication of the fact that this class is an expression of type `ConstraintViolation`. Perhaps, we can have something like `classOf()` in `Expression` so that the `<R>` parameter becomes more useful?

tools/validation-generator/src/main/java/io/spine/tools/validate/field/Constraint.java
```diff
@@ -0,0 +1,80 @@
...
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * A validation constraint based on a Protobuf option.
+ */
+final class Constraint implements Rule {
+
+    private final Function<? super Expression<?>, ? extends Expression<Boolean>> condition;
+    private final
+    Function<? super Expression<?>, ? extends Expression<ConstraintViolation>> violationFactory;
```
alexander-yevsyukov: Can we wrap `Function<? super Expression<?>, ...>` into something which tells what it's all about?

tools/validation-generator/src/main/java/io/spine/tools/validate/field/FieldValidatorFactories.java
```diff
@@ -0,0 +1,130 @@
...
+            case BOOLEAN:
+            default:
+                return NoOpFactory.INSTANCE;
+        }
+    }
+
+    /**
+     * A {@link FieldValidatorFactory} which never produces any field validation.
+     */
+    private enum NoOpFactory implements FieldValidatorFactory {
```
alexander-yevsyukov: Please pull it to the package level. It would make the null object more visible for the people exploring the package.

tools/validation-generator/src/main/java/io/spine/tools/validate/field/NumberFieldValidatorFactory.java
```diff
@@ -0,0 +1,268 @@
...
+        Boundary low = null;
+        MinOption minOption = field().findOption(min);
+        if (isNotDefault(minOption)) {
+            low = new Boundary(numberKind.parse(minOption.getValue()), !minOption.getExclusive());
+
+        }
+        Boundary high = null;
+        MaxOption maxOption = field().findOption(max);
+        if (isNotDefault(maxOption)) {
+            high = new Boundary(numberKind.parse(maxOption.getValue()), !maxOption.getExclusive());
```
alexander-yevsyukov: The construction of `Boundary` for `low` and `high` shares the same principle. Can it be extracted into a method?

tools/validation-generator/src/main/java/io/spine/tools/validate/field/NumberFieldValidatorFactory.java
```diff
@@ -0,0 +1,268 @@
...
+            return NumberBoundaries.unbound();
+        }
+    }
+
+    /**
+     * The representation of the number.
+     *
+     * <p>Each element corresponds to multiple types in Java and multiple types in Protobuf.
+     */
+    private enum NumberKind {
```
alexander-yevsyukov: Please pull it to the package level.

tools/validation-generator/src/main/java/io/spine/tools/validate/field/Rule.java
```diff
@@ -0,0 +1,59 @@
...
+package io.spine.tools.validate.field;
+
+import com.squareup.javapoet.CodeBlock;
+import io.spine.tools.validate.code.Expression;
+import io.spine.validate.ConstraintViolation;
+
+import java.util.function.Function;
+
+/**
+ * A validation rule.
```
alexander-yevsyukov: How is it differ from `ConstraintViolation`? I can get the understanding about it by looking at the code, but we need something more about this interface at its top Javadoc.

tools/validation-generator/src/main/java/io/spine/tools/validate/field/RecursiveValidation.java
```diff
@@ -0,0 +1,84 @@
...
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * A validation rule which checks that a message field is valid.
+ *
+ * <p>If a field message is invalid, the message's {@link ConstraintViolation}s are wrapped into
+ * a single violation and added to the rest of the violations of the top-level message.
+ */
+final class RecursiveValidation implements Rule {
```
alexander-yevsyukov: Why `Validation` is the `Rule`?

alexander-yevsyukov: Why do I need to know about the recursive thing from the name of the class?

tools/validation-generator/src/main/java/io/spine/tools/validate/number/NumberBoundaries.java
```diff
@@ -0,0 +1,99 @@
...
+import io.spine.validate.ComparableNumber;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * Set of numerical boundaries for a certain number.
+ */
+public final class NumberBoundaries {
```
alexander-yevsyukov: Please consider using `Range` from Guava.

tools/validation-generator/src/main/java/io/spine/tools/validate/MessageValidatorFactory.java
```diff
@@ -0,0 +1,237 @@
...
+     * declared {@code private} and {@code static}.
+     *
+     * <p>The name of the class is most often {@code Validator}. However, when a name clash occurs,
+     * the name is padded with the {@code $} (dollar sign). For instance if the message itself is
+     * called {@code Validator}, the validator class will be called {@code Validator$}. Since
+     * the class is {@code private}, this naming is not exposed to the outer scope.
+     *
+     * <p>The only method of the class is:
+     * <pre>
+     *     {@code
```
armiol: Let's kill the `{@code}`. Please refer to one of our annotations (such as `@Command`) where I had experimented with the code samples in Javadoc and come to some conclusions.

tools/validation-generator/src/main/java/io/spine/tools/validate/MessageValidatorFactory.java
```diff
@@ -0,0 +1,237 @@
...
+                .addStatement(RETURN_N, MESSAGE_VARIABLE)
+                .build();
+        return MethodSpec
+                .methodBuilder("vBuild")
+                .addAnnotation(Override.class)
+                .addModifiers(PUBLIC)
+                .returns(messageClass)
+                .addException(exceptionClass)
+                .addCode(body)
+                .build();
```
armiol: I wonder why not to use `vBuild()` in the messages like this one...

tools/validation-generator/src/main/java/io/spine/tools/validate/field/IsEmpty.java
```diff
@@ -0,0 +1,44 @@
...
+package io.spine.tools.validate.field;
+
+import io.spine.tools.validate.code.Expression;
+
+import static io.spine.tools.validate.code.Expression.formatted;
+
+/**
+ * Constructs the validation rule which checks if a value if {@code .isEmpty()}.
+ */
+final class IsEmpty {
```
armiol: It's weird it's not a descendant of some more general item. Or — not a part of some other utility. 

Also `IsEmpty.isEmpty()` looks weird to me.

tools/protoc-plugin/src/main/java/io/spine/tools/protoc/validation/ValidatorGenerator.java
dmdashenkov: In `SpineProtocConfig` the name points at the fact that the configurated Protoc plugin is called `spineProtoc`.
For `SpineProtoGenerator`, I accept your suggestion.

tools/validation-generator/src/main/java/io/spine/tools/validate/MessageValidatorFactory.java
dmdashenkov: `MethodSpec` is not a message, just a Java API of JavaPoet.
If this doesn't answer your question, let's discuss.

base/src/main/proto/spine/options.proto
```diff
@@ -86,8 +86,7 @@ extend google.protobuf.FieldOptions {
     // If the field type is a `message`, it must be set to a non-default instance.
     // If it is `string` or `bytes`, the value must not be an empty string or an array.
     // Other field types are not applicable.
-    // If the field is repeated, it must have at least one value set, and all its values must be
-    // valid.
+    // If the field is repeated, it must have at least one value set.
```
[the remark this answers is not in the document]
dmdashenkov: I'd skip the `(validate)` part completely. The rewording suggestion accepted.

tools/validation-generator/src/main/java/io/spine/tools/validate/number/NumberBoundaries.java
dmdashenkov: I've tried and reverted to using `NumberBoundaries`. `Range` is too generic and requires compile-time knowledge of the boundary type (inclusive/exclusive). As for a simple data structure, `NumberBoundaries` is more useful.

tools/validation-generator/src/main/java/io/spine/tools/validate/code/ViolationTemplate.java
dmdashenkov: I intentionally reserved from adding a `GenericTypeIndex` for an `Expression`. I'd like to keep it as simple as possible and avoid exposing extra reflective API where it's not necessary.

dmdashenkov:
@alexander-yevsyukov, @armiol, PTAL again.
I've addressed most of your comments and answered a few of them. Let's discuss the leftover in person if necessary.

tools/validation-generator/src/main/java/io/spine/tools/validate/code/ViolationTemplate.java
```diff
@@ -0,0 +1,155 @@
...
+import io.spine.validate.ConstraintViolation;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static java.util.Arrays.asList;
+
+/**
+ * An {@link Expression} which creates a new {@link ConstraintViolation}.
+ */
+public final class ViolationTemplate implements Expression<ConstraintViolation> {
```
alexander-yevsyukov: Is it really a template? If yes, then this should be documented somehow.

To me it looks like the code generated to report constraint violation(s). It feels so because all the types in the `toCode()` method are already resolved by information on fields supplied to the `Builder`.

tools/validation-generator/src/main/java/io/spine/tools/validate/field/Rule.java
```diff
@@ -0,0 +1,68 @@
...
+import io.spine.tools.validate.ViolationAccumulator;
+import io.spine.tools.validate.code.Expression;
+import io.spine.validate.ConstraintViolation;
+
+import java.util.function.Function;
+
+/**
+ * A message validation rule.
+ *
+ * <p>A {@code Rule} may represent one or several validation constraints applied to a message. In
```
alexander-yevsyukov: Is it all of the constraints defined in the message?

I don't understand the second sentence. How can following the rule be _not present_ in it?

tools/validation-generator/src/main/java/io/spine/tools/validate/field/NestedConstraints.java
```diff
@@ -0,0 +1,93 @@
...
+import io.spine.validate.Validate;
+
+import java.lang.reflect.Type;
+import java.util.List;
+import java.util.function.Function;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * A validation rule which checks that a message field does not violate its own constraints.
```
alexander-yevsyukov: What do you mean by "its own constraints"? 

Does this class serve the recursive definitions of messages where a message may have a field of the same type?

tools/validation-generator/src/main/java/io/spine/tools/validate/field/Constraint.java
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * A validation constraint based on a Protobuf option.
+ */
+final class Constraint implements Rule {
+
+    private final Condition condition;
+    private final ViolationFactory violationFactory;
+
+    /**
+     * Creates a new {@code Rule}.
```
alexander-yevsyukov:
```suggestion
     * Creates a new {@code Constraint}.
```

tools/validation-generator/src/main/java/io/spine/tools/validate/field/ViolationFactory.java
```diff
@@ -0,0 +1,36 @@
...
+
+package io.spine.tools.validate.field;
+
+import io.spine.tools.validate.code.Expression;
+import io.spine.validate.ConstraintViolation;
+
+import java.util.function.Function;
+
+/**
+ * A function which transforms a field value into a {@link ConstraintViolation} featuring
```
alexander-yevsyukov: This description doesn't correspond to generic params supplied to the parent interface. These generic params tell me it transforms any expression into an expression with `ConstraintViolation`.

dmdashenkov:

base/src/main/java/io/spine/validate/ConstraintInterpreter.java
```diff
@@ -0,0 +1,300 @@
...
+import static io.spine.validate.MessageValue.nestedIn;
+import static java.util.stream.Collectors.toList;
+
+/**
+ * Interprets validation constraints by applying them to a given message value.
+ *
+ * <p>The output result of this {@link ConstraintTranslator} is a {@link ValidationError}.
+ */
+@SuppressWarnings("OverlyCoupledClass")
+final class ConstraintInterpreter implements ConstraintTranslator<Optional<ValidationError>> {
```
armiol: As discussed, among the triad `ConstraintTranslator`, `ConstraintInterpreter` and `ConstraintComplier` it takes some several seconds to realize which is which. Let's try to rename them so that each would sound straight "in your face" and require minimal energy for identification.

base/src/main/java/io/spine/validate/ConstraintTranslator.java
```diff
@@ -0,0 +1,68 @@
...
+ * A translator need not be thread-safe. An instance of translator is, most likely, a mutable
+ * object. If any resources must be closed when finishing the translation job, {@link #translate()}
+ * is the right place to do so.
+ *
+ * @see <a href="https://en.wikipedia.org/wiki/Translator_(computing)">
+ *         Definition of a "Translator"</a>
+ */
+public interface ConstraintTranslator<T> {
+
+    void visitRange(RangedConstraint<?> constraint);
```
armiol: As long as this is a part of the public API I would document this and the rest of the methods.

base/src/main/java/io/spine/validate/FieldValue.java
```diff
@@ -212,27 +148,68 @@ public JavaType javaType() {
...
+     * Creates a stream of non-default values of this field.
+     *
+     * <p>Behaves similarly to {@link #values()} but also filters out default values.
+     *
+     * <p>{@code 0} number values, {@code false} boolean values, {@code 0}-number enum instances,
+     * empty char and byte strings, and empty messages are considered default values.
+     */
+    public final Stream<?> nonDefault() {
+        return values()
+                .filter(val -> !isDefault(val));
```
armiol: I would move this one line back.

base/src/main/java/io/spine/validate/FieldValue.java
```diff
@@ -212,27 +148,68 @@ public JavaType javaType() {
...
+    }
+
+    /**
+     * Obtains the single value of this field.
+     *
+     * <p>If the field is a {@linkplain FieldDeclaration#isCollection() collection}, obtains
+     * the first element. If the fist element is not available, throws
```
armiol: `first`

base/src/main/java/io/spine/validate/FieldValue.java
```diff
@@ -212,27 +148,68 @@ public JavaType javaType() {
...
+     * an {@code IllegalStateException}.
+     *
+     * @return a single value of this field
+     */
+    public Object singleValue() {
+        checkState(!values.isEmpty(),
+                   "Unable to get the first element of an empty collection for the field `%s`.",
```
armiol: As far as I understand, the underlying field may be not a collection at all.

base/src/main/java/io/spine/validate/NumberConversionChecker.java
```diff
@@ -0,0 +1,187 @@
...
+ */
+
+package io.spine.validate;
+
+import com.google.common.collect.ImmutableList;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * Allows to determine safe variants of a number conversions without loosing precision.
```
armiol: `losing`.

tools/validation-generator/src/main/java/io/spine/tools/validate/package-info.java
```diff
@@ -0,0 +1,31 @@
...
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * Contains utilities for generating Java message validation code.
```
armiol: `Contains the utilities generating the Java code used for message validation.`

tools/validation-generator/src/main/java/io/spine/tools/validate/field/ConstraintCode.java
```diff
@@ -0,0 +1,228 @@
...
+
+import java.util.function.Function;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static io.spine.tools.validate.FieldAccess.element;
+import static io.spine.tools.validate.code.Blocks.empty;
+import static java.lang.System.lineSeparator;
+
+/**
+ * A validation constraint based on a Protobuf option of a message field.
```
armiol: This is a generated piece of code performing the validation.

tools/validation-generator/src/main/java/io/spine/tools/validate/code/Expression.java
```diff
@@ -0,0 +1,81 @@
...
+    CodeBlock toCode();
+
+    default CodeBlock returnStatement() {
+        return CodeBlock.of("return $L;$L", this.toCode(), lineSeparator());
+    }
+
+    /**
+     * Creates an {@code Expression} from the given value.
+     */
+    static <R> Expression<R> of(String code) {
```
armiol: As long as this one and those below are `public static` methods, I would check at least for `null`s.

tools/validation-generator/src/main/java/io/spine/tools/validate/code/IsSet.java
```diff
@@ -0,0 +1,199 @@
...
+
+    public IsSet(FieldDeclaration field) {
+        this.field = checkNotNull(field);
+        this.methodName = format("is%sSet", field.name()
+                                                 .toCamelCase());
+        this.methodBody = methodBody();
+    }
+
+    /**
+     * Invokes the method which checks if the value is set.
```
armiol: Not quite `invokes`. I think it generates the expression, in turn, checking whether the value is set.

tools/validation-generator/src/main/java/io/spine/tools/validate/code/IsSet.java
```diff
@@ -0,0 +1,199 @@
...
+import static io.spine.tools.validate.code.BooleanExpression.falseLiteral;
+import static io.spine.tools.validate.code.BooleanExpression.trueLiteral;
+import static java.lang.String.format;
+import static javax.lang.model.element.Modifier.PRIVATE;
+import static javax.lang.model.element.Modifier.STATIC;
+
+/**
+ * A factory of code which check is a given field is set.
+ */
+public final class IsSet {
```
armiol: My IDEA says that some code in this file isn't formatted.

tools/validation-generator/src/main/java/io/spine/tools/validate/code/BooleanExpression.java
```diff
@@ -0,0 +1,168 @@
...
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkState;
+
+/**
+ * An expression which yields a {@code boolean} value.
+ */
+public final class BooleanExpression
+        extends CodeExpression<Boolean> {
```
armiol: It's like the `BooleanExpression` is ashamed it's an `instanceof Expression`.

tools/validation-generator/src/main/java/io/spine/tools/validate/field/ConstraintCode.java
```diff
@@ -0,0 +1,228 @@
...
+ */
+public final class ConstraintCode implements Logging {
```
armiol: It's also weird that something called `...Code` isn't in the `.code` package. That also leads to the over-exposure of `Blocks` which otherwise would be package-private.

tools/validation-generator/src/main/java/io/spine/tools/validate/MessageValidatorFactory.java
```diff
@@ -0,0 +1,201 @@
...
+                .nestedDeclarations()
+                .stream()
+                .anyMatch(nested -> nested.simpleJavaClassName()
+                                          .value()
+                                          .equals(name));
+        return withNestedName;
+    }
+
+    /**
+     * Generates the {@code Validator} class for the associated message type.
```
armiol: I don't think we have `Validator`s as a concept anymore. Maybe you could even reconsider the name of the `MessageValidatorFactory` itself.

base/src/main/java/io/spine/validate/FieldValue.java
dmdashenkov: If the field is not a collection, the list **must** be non-empty, as by `resolveType(..)`.

tools/validation-generator/src/main/java/io/spine/tools/validate/MessageValidatorFactory.java
dmdashenkov: This is not the case. The Javadoc is up to date right now: the validation code is generated into a `private static final class Validator`.

dmdashenkov:
@armiol, PTAL again.

base/src/main/java/io/spine/validate/Constraints.java
```diff
@@ -0,0 +1,205 @@
...
+import static io.spine.util.Exceptions.illegalStateWithCauseOf;
+import static io.spine.validate.FieldConstraints.customFactoriesExist;
+
+/**
+ * Validation constraints of a single Protobuf message type.
+ */
+@Immutable
+public final class Constraints {
+
+    private static final int CACHE_SIZE = 1000;
```
alexander-yevsyukov: Please extract work with cache in a package-private class.

base/src/main/java/io/spine/validate/Duplicates.java
```diff
@@ -20,42 +20,30 @@
...
+import java.util.HashSet;
 import java.util.Set;
 
-import static java.lang.Math.abs;
-
-/**
- * Validates fields of {@link Float} types.
- */
-class FloatFieldValidator extends FloatFieldValidatorBase<Float> {
+public final class Duplicates {
```
alexander-yevsyukov: From what I see the sole method of this utility class is not used. Please check and eliminate if it is indeed not used. Otherwise, please document.

base/src/main/java/io/spine/validate/NumberConversionChecker.java
```diff
@@ -0,0 +1,187 @@
...
+import com.google.common.collect.ImmutableList;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * Allows to determine safe variants of a number conversions without losing precision.
+ *
+ * <p>Mimics the actual automatic conversions that are applied to primitive number types.
+ */
+final class NumberConversionChecker {
```
alexander-yevsyukov: Can we have `NumberConversion` instead, and get rid of the copy from the `option` sub-package?

base/src/main/java/io/spine/validate/Validate.java
```diff
@@ -43,11 +47,11 @@
 import static com.google.protobuf.TextFormat.shortDebugString;
 import static io.spine.util.Exceptions.newIllegalArgumentException;
 import static io.spine.util.Exceptions.newIllegalStateException;
-import static io.spine.validate.MessageValidator.validate;
 
 /**
  * This class provides general validation routines.
  */
+@SuppressWarnings("ClassWithTooManyMethods") // Many deprecated methods will be removed in future.
```
alexander-yevsyukov: I think it should be “in the future” in [this context](https://jakubmarian.com/in-the-future-vs-in-future-in-british-and-american-english/).

base/src/main/java/io/spine/validate/Validate.java
```diff
@@ -290,12 +294,56 @@ public static void checkBounds(int value, String paramName, int lowBound, int hi
...
-        List<ConstraintViolation> violations = validate(message);
+        List<ConstraintViolation> violations = violationsOf(message);
         if (!violations.isEmpty()) {
             throw new ValidationException(violations);
         }
     }
 
+    /**
+     * Validates the given message according to its definition and returns the constrain violations
+     * if any.
```
alexander-yevsyukov: Consider having comma before “if any”.

base/src/main/java/io/spine/validate/diags/ViolationText.java
```diff
@@ -67,6 +70,26 @@ public static String ofAll(Collection<ConstraintViolation> violations) {
         return result;
     }
 
+    /**
+     * Returns a validation error message which may have formatting placeholders
```
alexander-yevsyukov: Please have the period at the end of the summary sentence.

base/src/main/java/io/spine/validate/option/Distinct.java
```diff
@@ -21,36 +21,30 @@
...
-    public static <@ImmutableTypeParameter T> Distinct<T> create() {
-        return new Distinct<>();
+    public static Distinct create() {
+        return new Distinct();
     }
 
     @Override
-    public Constraint<FieldValue<T>> constraintFor(FieldValue<T> fieldValue) {
-        return new DistinctConstraint<>(optionValue(fieldValue));
+    public Constraint constraintFor(FieldContext fieldValue) {
```
alexander-yevsyukov: Please document.

base/src/main/java/io/spine/validate/option/Min.java
```diff
@@ -21,32 +21,29 @@
...
 /**
  * An option that defines a minimum value for a numeric field.
- *
- * @param <V>
- *         numeric value type that this option is applied to
  */
 @Immutable
-final class Min<@ImmutableTypeParameter V extends Number & Comparable<V>>
-        extends FieldValidatingOption<MinOption, V> {
+final class Min
```
alexander-yevsyukov: Since we no longer having lengthy generic params, consider having the class declaration on the same line (like in `Max`).

alexander-yevsyukov: Please also consider such a layout for declarations of the classes derived from `RangedConstraint`.

base/src/main/java/io/spine/validate/option/RangeConstraint.java
```diff
@@ -21,99 +21,120 @@
...
-        EdgeValues edgeValues = edgeValues(withoutBraces(trimmed));
-        ComparableNumber left = edgeValues.leftEdge.toNumber();
-        ComparableNumber right = edgeValues.rightEdge.toNumber();
-        Range<ComparableNumber> result = range.create(left, right);
-        return result;
+    static Range<ComparableNumber> rangeFromOption(String rangeOption, FieldDeclaration field) {
+        Matcher rangeMatcher = NUMBER_RANGE.matcher(rangeOption.trim());
+        if (!rangeOption.isEmpty()) {
+            checkState(rangeMatcher.matches(),
+                       "Range '%s' on field `%s` is invalid.", rangeOption, field);
```
alexander-yevsyukov: Why is it invalid? Please give the programmer a way out of this problem.

base/src/main/java/io/spine/validate/option/RangeConstraint.java
```diff
@@ -21,99 +21,120 @@
...
+    @Override
+    public String errorMessage(FieldContext field) {
+        return format("Field %s is out of range.", field());
     }
 
-    private static String withoutBraces(String value) {
-        return value.substring(1, value.length() - 1);
+    @Override
+    protected String compileErrorMessage(Range<ComparableNumber> range) {
+        return new StringBuilder(format("Value of `%s`", field()))
```
alexander-yevsyukov: Isn't it “The value of the...”?

tools/validation-generator/src/main/java/io/spine/tools/validate/code/IsSet.java
```diff
@@ -0,0 +1,200 @@
...
+     * @param parameter
+     *         the parameter to the method - the message containing the field
+     * @return an expression which checks if the field is set
+     * @see #method()
+     */
+    public BooleanExpression invocation(MessageAccess parameter) {
+        checkNotNull(parameter);
+        return alwaysTrue
+               ? trueLiteral()
+               : BooleanExpression.fromCode("$N($L)",
```
alexander-yevsyukov: Why do we need this branch wrapped this way?

tools/validation-generator/src/main/java/io/spine/tools/validate/code/VoidExpression.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+public final class VoidExpression extends CodeExpression<Void> {
+
+    private static final long serialVersionUID = 0L;
+
+    private VoidExpression(String value) {
+        super(value);
+    }
+
+    public static VoidExpression formatted(String template, Object... args) {
```
alexander-yevsyukov: Please document.

version.gradle
```diff
@@ -25,7 +25,7 @@
  * as we want to manage the versions in a single source.
  */
 
-final def SPINE_VERSION = '1.3.0'
+final def SPINE_VERSION = '1.3.1'
```
alexander-yevsyukov: I think with this feature, it can be deemed `1.4`. This is a big move forward.

base/src/main/java/io/spine/validate/option/Distinct.java
dmdashenkov: This method is overridden and is documented in `ValidatingOption`. If you suggest adding some specifics regarding `Distinct`, let's discuss.

base/src/main/proto/spine/options.proto
```diff
@@ -545,7 +544,7 @@ message MinOption {
     //   1) "or equal to " string (if the `exclusive` parameter is false) or an empty string;
     //   2) the minimum number.
     //
-    option (default_message) = "Number must be greater than %s %s.";
+    option (default_message) = "Number must be greater than %s%s.";
```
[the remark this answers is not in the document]
dmdashenkov: I've updated to `The number`, `The string`, `The message` and so forth. Also instead of `Value must be set.`  —  `A value must be set.`.

tools/validation-generator/src/main/java/io/spine/tools/validate/code/Blocks.java
```diff
@@ -0,0 +1,44 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.spine.tools.validate.code;
+
+import com.squareup.javapoet.CodeBlock;
+
+/**
+ * A factory of {@link CodeBlock}s.
```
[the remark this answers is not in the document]
dmdashenkov: It seems more descriptive to use a statically imported `empty()`.

dmdashenkov:

base/src/main/java/io/spine/validate/option/RangeConstraint.java
```diff
@@ -21,99 +21,112 @@
...
+        } else {
+            return maxInclusive ? openClosed(minValue, maxValue) : open(minValue, maxValue);
+        }
     }
 
-    private static String withoutBraces(String value) {
-        return value.substring(1, value.length() - 1);
+    @Override
+    protected String compileErrorMessage(Range<ComparableNumber> range) {
+        return format("The value of the field `%s` is out of range. Must be %s%s and %s%s.",
```
alexander-yevsyukov: I've updated the language here myself, removing the word “between” which was redundant, assuming that following text says “less than”. But there's one more thing about this error message. What _is_ the value, which is out of range?

I think, we need to tell it to make finding the problem easier. Can we do that?

tools/validation-generator/src/main/java/io/spine/tools/validate/code/Containers.java
```diff
@@ -0,0 +1,61 @@
...
+     *         an expression which yields an object which has a {@code isEmpty()} method, e.g.
+     *         a {@code String}
+     */
+    public static BooleanExpression isEmpty(Expression<?> value) {
+        return fromCode("$L.isEmpty()", value.toCode());
+    }
+
+    /**
+     * Obtains the expression which calls {@code Messages.isDefault())} method on the given
+     * {@code value}.
```
alexander-yevsyukov: A common issue with Javadocs we have is [widow words](https://opusdesign.us/wordcount/typographic-widows-orphans/) at the end of some paragraphs. 

I'm not suggesting to review and fix it throughout the Javadocs now (especially assuming this big and important PR). I just want to tell you about such a thing so that we can start improving this aspect of the documentation in future works.

base/src/main/java/io/spine/validate/option/RangeConstraint.java
dmdashenkov: It is possible to add the field value to the message. `(min)` and `(max)` also don't provide the value in the message. However, in both cases (of `(range)` and `(min)`/`(max)`) provide the field value in the violation body.
I'd rather leave this behaviour as is. This gives the users an additional perk by making the error message non-sensitive, i.e. such that does not contain potentially sensitive data.

dmdashenkov:
@armiol, PTAL if you will.

PR #499, opened 2019-12-03
Change single quotes to backticks in error messages

dmdashenkov:
Message values and validation parameters look more distinct when wrapped into backticks:
```
The field 'foo' can only be set when the field 'bar' is defined.
```
VS
```
The field `foo` can only be set when the field `bar` is defined.
```


dmdashenkov:
@alexander-yevsyukov, PTAL.

PR #501, opened 2019-12-16
Improve enum types conversion

dmitrykuzmin:
This PR improves the handling of enum types in the `TypeConverter`.

Now, the converter will be able to handle all types of `EnumValue` instances: with the enum constant name specified, with the enum value number specified and with both of them. In the latter case the name will be used for conversion as it's presumably more efficient.

Such change is required to simplify the code of clients in languages where it's hard to obtain a Proto-generated enum constant name, for example, JS.

Spine version advances to `1.2.4`.


dmitrykuzmin:
@armiol PTAL.

dmitrykuzmin:
@dmdashenkov PTAL.

PR #502, opened 2019-12-30
Remove `maven` Gradle plugin

dmdashenkov:
This PR removes the usages of `maven` plugin.

The plugin hasn't been used for publishing for a long time, if ever. However, the build relied heavily on the side effects caused by applying the plugin. For example, the plugin also applied `java` which was required by unrelated configurations (such as `jacoco.gradle`).

Now, the plugin is not used and the side effects are reached by more explicit ways, e.g. applying the `base` plugin.

The version of `base` is bumped to 1.3.2, primarily to make sure that publishing is not broken.

### DIFF
--- a/base/build.gradle
+++ b/base/build.gradle
@@ -22,8 +22,6 @@ import java.nio.file.Files
 
 group = 'io.spine'
 
-apply plugin: 'java'
-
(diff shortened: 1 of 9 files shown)


dmdashenkov:
@armiol, PTAL.

PR #507, opened 2020-01-22
`Method` to `MethodHandle`

dmdashenkov:
This PR adds a utility for converting an arbitrary `java.lang.reflect.Method` into a `java.lang.invoke.MethodHandle`.

With `MethodHandles`, we do not have to change the accessibility attribute upon each invocation.


dmdashenkov:
@armiol, PTAL.

PR #510, opened 2020-02-20
Bulk unpack `Any`s

dmdashenkov:
In order to unpack an `Any`, we perform some costly operations, such as finding type URLs and default instances of a type.

In this PR we add a function, which does that once and unpacks multiple instances of the same type more efficiently.


dmdashenkov:
@armiol, PTAL.

base/src/main/java/io/spine/protobuf/AnyPacker.java
```diff
@@ -143,6 +144,47 @@ public static Message unpack(Any any) {
...
+     * <p>The function returns {@code null} for {@code null} input.
+     *
+     * <p>The function throws a {@link UnexpectedTypeException} if the actual type of the message
+     * does not match the given class.
+     *
+     * @param type
+     *         expected class of the messages
+     */
+    public static <T extends Message> Function<@Nullable Any, @Nullable T>
+    unpackFunc(Class<T> type) {
```
armiol: I would go with `unpackFn`.

dmdashenkov: We've already got an overload called `unpackFunc`. I'm leaving this as is for the sake of consistency.

PR #513, opened 2020-02-25
Static logger API

dmdashenkov:
In this PR we add the `Logging.loggerFor(Class)` method which obtains a logger for a given class without making the class implement `Logging` interface.


dmdashenkov:
@armiol, PTAL.

PR #514, opened 2020-02-28
Add more assertions for `LogRecordSubject`

dmdashenkov:
In this PR we add two assertion methods for checking `LogRecordSubject`.


dmdashenkov:
@armiol, PTAL.

PR #523, opened 2020-04-03
Required `oneof`

### LINKED ISSUE #521: Allow requiring `oneof` values
In some cases, it is needed to require that one of the fields coming under `oneof` is set. For example, if we create a user account it can be either `person` or `service` field must be set. In order to arrange such a requirement, we can use the following construct:

```protobuf
option (required_field) = "person|service";

// ...

oneof type {
    // The user is a human.
    PersonProfile person = 2;

    // The user is a computer program.
    ServiceProfile service = 3;
}

```
The problem with this approach is that it's fragile and breaks when we:
  1. rename fields;
  2. add fields under `oneof`.

We cannot annotate `oneof` itself, but we cannot reduce the dependency on the content of it by telling the compiler: “We do want one of the things coming under that `oneof`.” Here's how:

```protobuf
// One of the fields coming under `type` `oneof` must be populated.
option (required).field = "type";
```
This is still fragile for renaming the `oneof` itself, but they tend to be more stable, and it's still safer as we refer only one element.

dmdashenkov:
In this PR we add a new validation option: `(is_required)` for `oneof` group fields.

Previously, to make a `oneof` "required" (i.e. to enforce that one field is always set), the users had to write a `(required_field)` expression:
```proto
message Meal {
    option (required_field) = "fish | meat | veggies";

    oneof choice {
        Fish fish = 1;
        Meat meat = 2;
        Veggitables veggies = 3;
    }
}
```
This approach is error-prone, as the field names may be misspelled, the fields may be added or deleted, etc. 

Now, the users can mark the `oneof` group with a `(is_required)` option:
```proto
message Meal {
    oneof choice {
        option (is_required) = true;

        Fish fish = 1;
        Meat meat = 2;
        Veggitables veggies = 3;
    }
}
```
The second snippet is equivalent to the first one but has less room for accidental errors.

Fixes #521.


dmdashenkov:

base/src/main/java/io/spine/type/OneofDeclaration.java
```diff
@@ -18,22 +18,18 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-package io.spine.code.gen.java;
+package io.spine.type;
```
alexander-yevsyukov, line 21: Why do we change the package?

alexander-yevsyukov: Currently the `type` package contains only few classes that deal with known types.

base/src/test/java/io/spine/type/OneofDeclarationTest.java
```diff
@@ -18,14 +18,13 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-package io.spine.code.gen.java;
+package io.spine.type;
```
alexander-yevsyukov, line 21: The same question regarding the package as for the class under the test. Why do change the package?

dmdashenkov: We move `OneofDeclaration` from another module. I've changed the package to a more appropriate `io.spine.code.proto`, which is also the package where the `FieldDeclaration` resides.

dmdashenkov:

PR #524, opened 2020-04-08
Resolve Dart Protobuf imports

dmdashenkov:
# The problem

Dart, just like JavaScript, has a file-based import system (as opposed to class-based in Java).

When compiling Protobuf into Dart, we have to compile all the accessible `.proto` files, including those from the project's dependencies. Otherwise, the compiled files would have broken imports pointing at third-party types which are not there. 

## Example

Module `my-app` uses library `my-lib`. Both `my-app` and `my-lib` declare Protobuf types and add utilities for working with those types.

If `my-app` types use `my-lib` types, the imports generated in `my-app` will point to relative paths where the `my-lib` generated Dart classes should be. So, if we don't compile `my-lib` types in `my-app`, `my-app` types contain broken imports.

# Solution

Just like in JS, we fix the imports in the generated Dart code. For that, we require the user to construct a map of all the known Dart modules with generated code to their Protobuf packages (or package prefixes).

```gradle
protoDart {
    modules = [
            'spine-client' : ['spine/*', 'google/*'],
            'spine-users'  : ['spine/users/*']
    ]
}
```


dmdashenkov:
@armiol, PTAL.

tools/plugin-base/src/main/java/io/spine/tools/code/ExternalModule.java
```diff
@@ -18,9 +18,10 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-package io.spine.js.generate.resolve;
+package io.spine.tools.code;
```
armiol, line 21: According to the class-level comment, this is a `JavaScript module`. However, as you move it out from the JS-specific package, it no longer looks as such. `NPM` reference is also misleading to me.

tools/proto-dart-plugin/src/main/java/io/spine/generate/dart/Extension.java
```diff
@@ -50,6 +59,38 @@
     private final DirectoryProperty testDir;
     private final DirectoryProperty mainGeneratedDir;
     private final DirectoryProperty testGeneratedDir;
+
+    /**
+     * Names of JavaScript modules and directories they provide.
```
armiol, line 64: I don't understand why it is a `JavaScript module` you speak about.

tools/proto-dart-plugin/src/main/java/io/spine/generate/dart/Extension.java
```diff
@@ -50,6 +59,38 @@
...
+     *
+     * <p>Additionally to modules specified via the property,
+     * the {@linkplain ExternalModule#predefinedModules() predefined Spine} modules are used.
+     *
+     * <p>An example of the definition:
+     * <pre>{@code
+     * modules = [
+     *      // The module provides `company/client` directory (not including subdirectories).
+     *      // So, an import path like {@code ../company/client/file.js}
+     *      // becomes {@code client/company/client/file.js}.
```
armiol, line 76: And `.js` files in the reference.

tools/proto-dart-plugin/src/main/java/io/spine/generate/dart/SourceFile.java
```diff
@@ -0,0 +1,131 @@
...
+
+    private List<String> lines;
+
+    private SourceFile(Path path, List<String> lines) {
+        super(path);
+        this.lines = lines;
+    }
+
+    /**
+     * Reads the file from disc.
```
armiol, line 61: I think it's from "a local filesystem".

tools/proto-dart-plugin/src/main/java/io/spine/generate/dart/SourceFile.java
```diff
@@ -0,0 +1,131 @@
...
+                || file.endsWith(".pbjson.dart");
+    }
+
+    /**
+     * Resolves the relative imports in the file into absolute ones with the given modules.
+     */
+    void resolveImports(ImmutableList<ExternalModule> modules, Path libPath) {
+        lines = lines.stream()
+                     .map(line -> resolveImportInLine(line, modules, libPath))
+                     .collect(toList());
```
armiol, line 89: This one is confusing. So you could just call the method for each of the modules, but instead, you are collecting them into a redundant list to invoke the methods.

tools/proto-js-plugin/src/main/java/io/spine/js/generate/imports/package-info.java
```diff
@@ -0,0 +1,27 @@
...
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+@CheckReturnValue
+@ParametersAreNonnullByDefault
+package io.spine.js.generate.imports;
```
armiol, line 23: Please document the package.

tools/proto-js-plugin/src/test/java/io/spine/js/generate/imports/ResolveImportsTest.java
```diff
@@ -148,4 +150,10 @@ private static IterableSubject afterResolve(Path file, ResolveImports task) thro
     private ResolveImports newTask(ExternalModule module) {
         return new ResolveImports(generatedProtoDir, ImmutableList.of(module));
     }
+
+    private static ExternalModule newModule(String moduleName, String directoryPattern) {
```
armiol, line 154: It's weird that this file has just this one `private static` method appended. Please review.

tools/proto-js-plugin/src/test/java/io/spine/js/generate/imports/given/package-info.java
```diff
@@ -0,0 +1,28 @@
...
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+@CheckReturnValue
+@ParametersAreNonnullByDefault
+package io.spine.js.generate.imports.given;
+
+import com.google.errorprone.annotations.CheckReturnValue;
```
armiol, line 26: Please document this package, for consistency sake.

dmdashenkov:
@armiol, PTAL again.

PR #525, opened 2020-04-16
Publish base Protos as a separate artifact

yuri-sergiichuk:
I've added publishing of `base` module protos as a separate artifact to ease integrations for programming languages that are not yet supported by the framework.


yuri-sergiichuk:
@dmdashenkov PTAL.

PR #526, opened 2020-04-17
Advance version to 1.5.8

alexander-yevsyukov:
This PR advances the version to be able to build the `bootstrap` plugin (which requires all the dependencies to have the same version number).

### DIFF
--- a/version.gradle
+++ b/version.gradle
@@ -25,7 +25,7 @@
  * as we want to manage the versions in a single source.
  */
 
-final def SPINE_VERSION = '1.5.6'
+final def SPINE_VERSION = '1.5.8'
 
 ext {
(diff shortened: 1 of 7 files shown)


version.gradle
```diff
@@ -25,7 +25,7 @@
  * as we want to manage the versions in a single source.
  */
 
-final def SPINE_VERSION = '1.5.6'
+final def SPINE_VERSION = '1.5.8'
```
dmdashenkov, line 28: Please use version `1.5.7`.

alexander-yevsyukov: I need 1.5.8 because `base` and `core-java` (which is going to have this version soon too) need to have the same number so that `bootstrap` can be built.

PR #530, opened 2020-04-27
`ConstraintViolation`s for `(set_once)`

dmdashenkov:
Currently, we only provide a method for throwing a `ValidationException` if the `(set_once)` constraint is violated. However, it can also be helpful to be able to obtain violations without an exception. For that, we introduce the new `Validate.validateChange(..)` method.


dmdashenkov:
@armiol, PTAL.

PR #531, opened 2020-04-27
Allow `(is)` & `(every_is)` simultaneously

dmdashenkov:
Consider options `(is)` and `(every_is)` applied to a type at the same time:

```proto
option (every_is).java_type = "Fruit";

message Tomato {
    option (is).java_type = "Vegetable";

    // ...
}
```

Currently, Java class `Tomato` only implements `Vegetable` interface.

With this PR we make `(is)` and `(every_is)` complementary. This way, Java class `Tomato` implements both `Vegetable` and `Fruit` interfaces.


dmdashenkov:
@armiol, PTAL.

PR #539, opened 2020-05-26
Custom environment types

serhii-lekariev:
This PR extends the `Environment` API to allow the definition of custom environment types. It also migrates the current boolean test/not-a-test setup to a class-based solution.

To register a new environment type, define a custom one like so:

```java
import io.spine.base.EnvironmentType;

public final class Local extends EnvironmentType {
        
         // ...

        @Override
         protected boolean enabled() {
            // Check an environment variable or a file.
         }

        public static Local type() {
        // Obtain the instance somehow. Encouraged to be a singleton instance.
        }
  }
```

```java
Environment environment = Environment.instance();

// As the type is not registered yet.
assertThat(environment.is(Local.type)).isFalse();

Environment.register(Local.type());

// Assuming that "Local" is actually enabled
assertThat(environment.is(Local.type)).isTrue();
assertThat(environment.type()).isSameInstanceAs(Local.type());

```

When checking the environment types, user-defined types are checked first. If no type was found to be `enabled`, falls back to `io.spine.base.Production`.

#### Deprecation notes

The following methods are now deprecated in favor of the new API:

- `isTests`: use `is(Tests.type());`;
- `setToTests`: use `setTo(Tests.type());`;
- `isProduction`: use `is(Production.type());`;
- `setToProduction`: use `setTo(Production.type());`.


serhii-lekariev:
@armiol, PTAL.

base/src/main/java/io/spine/base/BaseEnvironmentType.java
```diff
@@ -0,0 +1,112 @@
...
+         * <ul>
+         *     <li>{@code org.junit}
+         *     <li>{@code org.testng}
+         * </ul>
+         *
+         * @return {@code true} if the code runs under a testing framework, {@code false} otherwise
+         */
+        @Override
+        public boolean currentlyOn() {
+            // Check the environment variable. We may run under unknown testing framework or
```
armiol, line 55: The long pieces like this should go under `@implNote` in the method-level Javadoc.

base/src/main/java/io/spine/base/BaseEnvironmentType.java
```diff
@@ -0,0 +1,112 @@
...
+
+package io.spine.base;
+
+import com.google.common.base.Throwables;
+import com.google.common.collect.ImmutableList;
+
+import java.util.regex.Pattern;
+
+/**
+ * Environment types provided by the base library.
```
armiol, line 29:
```
...{@code base} library.
```

base/src/main/java/io/spine/base/BaseEnvironmentType.java
```diff
@@ -0,0 +1,112 @@
...
+ *
+ * <p>Provides a {@code TESTS} option, that looks into the stack trace to find mentions of
```
armiol, line 31: This paragraph and the next one should go as a documentation of the enum members, not the class-level one.

base/src/main/java/io/spine/base/BaseEnvironmentType.java
```diff
@@ -0,0 +1,112 @@
...
+         * in the stacktrace:
+         * <ul>
+         *     <li>{@code org.junit}
+         *     <li>{@code org.testng}
+         * </ul>
+         *
+         * @return {@code true} if the code runs under a testing framework, {@code false} otherwise
+         */
+        @Override
+        public boolean currentlyOn() {
```
armiol, line 54: Let's think of some better naming:

```
if(TESTS.currentlyOn()) {
    //...
}
```
It does not quite read to me. We are not "on tests" per se. The current environment is a test environment, that's true.

base/src/main/java/io/spine/base/BaseEnvironmentType.java
```diff
@@ -0,0 +1,112 @@
...
+            }
+
+            // Check stacktrace for known frameworks.
+            String stacktrace = Throwables.getStackTraceAsString(new RuntimeException(""));
+            return KNOWN_TESTING_FRAMEWORKS.stream()
+                                           .anyMatch(stacktrace::contains);
+        }
+
+        @Override
+        public void reset() {
```
armiol, line 73: Do we need this method at all?

For `TESTS` the `reset()` won't work, as you cannot change whether you inside the JUnit call stack or not.

base/src/main/java/io/spine/base/Environment.java
```diff
@@ -21,37 +21,76 @@
...
 
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkState;
+import static io.spine.base.BaseEnvironmentType.PRODUCTION;
+import static io.spine.base.BaseEnvironmentType.TESTS;
+
 /**
  * Provides information about the environment (current platform used, etc.).
+ *
+ * <p>By default, knows only about {@link BaseEnvironmentType}. Library users may extend the list
```
armiol, line 36: `BaseEnvironmentType` is a package-private enum. `Environment` is a public class. Something isn't right.

I wouldn't metion that at all.

base/src/main/java/io/spine/base/Environment.java
```diff
@@ -21,37 +21,76 @@
...
+     * <p>Note that the {@linkplain BaseEnvironmentType default types} are still present.
+     * When trying to {@linkplain #envType() determine which environment type} is currently on,
+     * the user defined types are checked first.
+     *
+     * @param enumClass
+     *         an enum class that specifies the environment types
+     * @param <E>
+     *         a type that defines possible environment types
+     */
+    public static <E extends Enum & EnvironmentType> void registerCustom(Class<E> enumClass) {
```
armiol, line 82: As discussed in person, we may get rid of the explicit registration and make our peace with it.

base/src/main/java/io/spine/base/Environment.java
```diff
@@ -69,101 +108,65 @@ public Environment createCopy() {
...
     }
 
     /**
-     * Verifies if the code runs in the production mode.
-     *
-     * <p>This method is opposite to {@link #isTests()}
-     *
-     * @return {@code true} if the code runs in the production mode, {@code false} otherwise
-     */
-    public boolean isProduction() {
```
armiol, line 135: Let's bring this public API back.

base/src/main/java/io/spine/base/Environment.java
```diff
@@ -69,101 +108,65 @@ public Environment createCopy() {
...
+     * {@code false}, this method falls back on {@link BaseEnvironmentType#PRODUCTION}.
      *
-     * @return {@code true} if the code runs under a testing framework, {@code false} otherwise
+     * @return the current environment type.
      */
-    @SuppressWarnings({
-            "DynamicRegexReplaceableByCompiledPattern", // OK as we cache the result
-            "DuplicateStringLiteralInspection" // used in another context
-    })
-    public boolean isTests() {
```
armiol, line 98: Let's bring this public API method back. Its implementation will be simpler, that's for sure.

base/src/main/java/io/spine/base/EnvironmentType.java
```diff
@@ -0,0 +1,56 @@
...
+public interface EnvironmentType {
+
+    /**
+     * Returns {@code true} if the underlying system is currently in this environment type.
+     *
+     * <p>For example, let's say that an application is deployed to a fleet of virtual machines.
+     * Let's say an environment variable is set for every virtual machine.
+     * An application developer may use this type of knowledge to determine the current environment.
+     */
+    boolean currentlyOn();
```
armiol, line 38: In the class-level docs I read 
```
A type of an environment that knows whether it is enabled.
```
Why this method isn't called `isEnabled()`?

base/src/main/java/io/spine/base/EnvironmentType.java
```diff
@@ -0,0 +1,56 @@
...
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.spine.base;
+
+/**
+ * A type of an environment that knows whether it is enabled.
```
armiol, line 24: I think `an` may be removed. Please use some tools to verify this text.

base/src/main/java/io/spine/base/EnvironmentType.java
```diff
@@ -0,0 +1,56 @@
...
+     */
+    boolean currentlyOn();
+
+    /**
+     * Makes it so the underlying system no longer has this environment type.
+     *
+     * <p>Be careful, since this method may mutate the state that is outside the scope of the
+     * Java application, such as clear an environment variable, remove a file, etc.
+     */
+    void reset();
```
armiol, line 46: See my comment above. I don't think we need this method at all.

And Javadocs make not much sense to me. 
```
...underlying system no longer has this environment type.
```
is clearly not true, as the classes and enum members are still there.

base/src/main/java/io/spine/base/EnvironmentType.java
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * Makes it so the underlying system has this environment type.
+     *
+     * <p>Be careful, since this method may mutate the state that is outside the scope of the
+     * Java application, such as set an environment variable, create or move a file, etc.
+     */
+    void setTo();
```
armiol, line 54: I don't think we need this method here. It's a courtesy of `Environment` to maintain the current env type.

base/src/main/java/io/spine/base/EnvironmentType.java
```diff
@@ -0,0 +1,56 @@
...
+}
+
```
armiol, line 56: This line is redundant.

serhii-lekariev:
@armiol, PTAL.

base/src/main/java/io/spine/base/BaseEnvironmentType.java
```diff
@@ -0,0 +1,102 @@
...
+                return String.valueOf(true)
+                             .equalsIgnoreCase(testProp) || "1".equals(testProp);
+            }
+
+            String stacktrace = Throwables.getStackTraceAsString(new RuntimeException(""));
+            return KNOWN_TESTING_FRAMEWORKS.stream()
+                                           .anyMatch(stacktrace::contains);
+        }
+    },
+    /**
```
armiol, line 75: Please have an empty line above.

base/src/main/java/io/spine/base/Environment.java
```diff
@@ -21,37 +21,71 @@
...
 import org.checkerframework.checker.nullness.qual.Nullable;
 
+import static com.google.common.base.Preconditions.checkState;
+import static io.spine.base.BaseEnvironmentType.PRODUCTION;
+import static io.spine.base.BaseEnvironmentType.TESTS;
+
 /**
  * Provides information about the environment (current platform used, etc.).
+ *
+ * <p><b>When extending, please note</b> that this class does not handle the situations when two
```
armiol, line 36: Let's rephrase in terms of the responsibitliy of the framework user. It's not that this class handles or does not handle any use cases. It just delegates the responsibility to tell between different `PRODUCTION` environments to the developer writing the code.

base/src/main/java/io/spine/base/Environment.java
```diff
@@ -21,37 +21,71 @@
...
+     *
+     * <p>Note that the {@linkplain BaseEnvironmentType default types} are still present.
+     * When trying to {@linkplain #currentType() determine which environment type} is enabled,
+     * the user defined types are checked first.
+     *
+     * @param environmentType
+     *         a user-defined environment type
+     */
+    @Internal
+    public static void registerCustom(EnvironmentType environmentType) {
```
armiol, line 78: Let's remove `custom` from the name.

base/src/main/java/io/spine/base/Environment.java
```diff
@@ -21,37 +21,71 @@
...
+ * think that they are currently enabled, <b>the behaviour of {@link #currentType()} is
+ * undefined.</b>
  */
 @SPI
-@SuppressWarnings("AccessOfSystemProperties") // OK as we need system properties for this class.
 public final class Environment {
 
-    private static final Environment INSTANCE = new Environment();
+    private static final ImmutableList<EnvironmentType> BASE_TYPES =
+            ImmutableList.of(TESTS, PRODUCTION);
```
armiol, line 46: Instead of putting the items explicitly, let's use `values()` of the enum type and  `copyOf()`.

base/src/main/java/io/spine/base/Environment.java
```diff
@@ -21,37 +21,71 @@
...
+    /**
+     * Remembers the specified environment type, allowing {@linkplain #currentType()}
+     * to determine whether it's enabled} later.
+     *
+     * <p>If the specified environment type has already been registered, throws an
+     * {@code IllegalStateException}.
+     *
+     * <p>Note that the {@linkplain BaseEnvironmentType default types} are still present.
+     * When trying to {@linkplain #currentType() determine which environment type} is enabled,
+     * the user defined types are checked first.
```
armiol, line 72: `user-defined`.

base/src/main/java/io/spine/base/Environment.java
```diff
@@ -69,101 +103,55 @@ public Environment createCopy() {
...
-    @VisibleForTesting
-    public void restoreFrom(Environment copy) {
-        // Make sure this matches the set of fields copied in the copy constructor.
-        this.tests = copy.tests;
-    }
-
-    /**
-     * Verifies if the code currently runs under a unit testing framework.
+     * <p>If {@linkplain #registerCustom(EnvironmentType) custom env types have been defined},
+     * goes through them in an undefined order. Then, checks the {@linkplain BaseEnvironmentType
```
armiol, line 109: I don't think the order is undefined. Let's check those registered by the user in the order of their registration. I.e. if I register some `STAGING` first, and `PRE_PRODUCTION` second, we should check `STAGING` first.

base/src/main/java/io/spine/base/Environment.java
```diff
@@ -69,101 +103,55 @@ public Environment createCopy() {
...
-     *
-     * <p>This method is opposite to {@link #setToTests()}.
+     * Forces the specified environment type to be the current one.
      */
     @VisibleForTesting
-    public void setToProduction() {
-        this.tests = false;
-        System.setProperty(ENV_KEY_TESTS, String.valueOf(false));
+    public void setTo(EnvironmentType type) {
+        this.currentEnvType = type;
```
armiol, line 146: Let's check that the value is not `null`.

base/src/main/java/io/spine/base/Environment.java
```diff
@@ -69,101 +103,55 @@ public Environment createCopy() {
...
     }
 
     /**
-     * Resets the instance and clears the {@link #ENV_KEY_TESTS} variable.
+     * Resets the instance.
```
armiol, line 150: Why the behavior of clearing the variable is changed?

base/src/main/java/io/spine/base/EnvironmentType.java
```diff
@@ -0,0 +1,39 @@
...
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.spine.base;
+
+/**
+ * A type of environment that knows whether it is enabled.
```
armiol, line 24: `that knows whether it is enabled` portion duplicates the contract of the interface. Please remove.

base/src/main/java/io/spine/base/EnvironmentType.java
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * <p>Useful examples may include distinguishable {@code STAGING} or {@code LOCAL} environments.
```
armiol, line 26: `Some examples may be {@code STAGING} or {@code LOCAL} environments.`

base/src/main/java/io/spine/base/EnvironmentType.java
```diff
@@ -0,0 +1,39 @@
...
+ * {@code base} library provides {@linkplain BaseEnvironmentType default environment types}.
```
armiol, line 27: Instead of this sentence, use `@see ` with the reference to the `BaseEnvironmentType`.

base/src/main/java/io/spine/base/Environment.java
serhii-lekariev: It's actually defined, the last environment to `register` is the first one that checked.

I thought about the following scenario:

Having two env types that are not mutually exclusive is a bug, but if the env types are checked in a predictable order, the bug becomes harder to discover.

serhii-lekariev:
@armiol, PTAL again.

base/src/test/java/io/spine/base/EnvironmentTest.java
```diff
@@ -159,15 +159,15 @@ void mutateKnownEnvTypesOnce() {
         @Test
         @DisplayName("throw if a user attempts to create register the same environment twice")
         void throwOnDoubleCreation() {
```
armiol, line 161: Is this a `creation`?

base/src/main/java/io/spine/base/Environment.java
```diff
@@ -26,24 +26,21 @@
...
 
 /**
  * Provides information about the environment (current platform used, etc.).
  *
- * <p><b>When extending, please note</b> that this class does not handle the situations when two
- * or more {@linkplain EnvironmentType environment types} return {@code true} on the
- * {@link EnvironmentType#enabled()}. As such, if two or more user-defined environment types
- * think that they are currently enabled, <b>the behaviour of {@link #currentType()} is
- * undefined.</b>
+ * <p><b>When extending, please ensure</b> the mutual exclusivity on your {@code EnvironmentType}s.
```
armiol, line 35: This class cannot be extended. Please put a bit more into this thought.

base/src/main/java/io/spine/base/BaseEnvironmentType.java
```diff
@@ -0,0 +1,103 @@
...
+     * Testing environment.
+     *
+     * <p>Detected by checking stack trace for mentions of the known testing frameworks.
+     *
+     * <p>This option is mutually exclusive with {@link #PRODUCTION}, i.e. one of them is always
+     * enabled.
+     */
+    @SuppressWarnings("AccessOfSystemProperties" /* OK as we need system properties for this class. */)
+    TESTS {
+        /**
```
armiol, line 43: Please have an empty line above.

serhii-lekariev:
@armiol, PTAL again.

base/src/main/java/io/spine/base/Environment.java
```diff
@@ -32,17 +32,82 @@
...
+ *
+ * <ul>
+ *     <li><em>{@link BaseEnvironmentType#TESTS TESTS}</em> is detected
+ *     if the current call stack has a reference to the unit testing framework.
+ *
+ *     <li><em>{@link BaseEnvironmentType#PRODUCTION PRODUCTION}</em> is set in all other cases.
+ * </ul>
+ *
+ * <p>The framework users may define their custom settings depending on the {@linkplain}current
+ * environment type:
```
serhii-lekariev, line 48: Something isn't right.

base/src/main/java/io/spine/base/Environment.java
```diff
@@ -32,17 +32,82 @@
...
+ *
+ *     private Application() {
+ *         EnvironmentType type = Environment.instance()
+ *                                           .type();
+ *         if (type == BaseEnvironmentType.TESTS) {
+ *              // Single connection is enough for tests.
+ *             this.pool = new ConnectionPoolImpl(PoolCapacity.of(1));
+ *         } else {
+ *             if(LoadTestingType.instance().enabled()) {
+ *                 this.pool =
```
serhii-lekariev, line 96: Is this what we would expect people to do?

I thought that 
```java
else {
    if(type == LoadTestingType.instance()) {
       ...
    }
}
```
was the intended way to check the environment.

base/src/main/java/io/spine/base/Production.java
```diff
@@ -0,0 +1,59 @@
...
+    @Override
+    boolean enabled() {
+        return !Tests.instance()
+                     .enabled();
+    }
+
+    /**
+     * Returns the singleton instance.
+     */
+    public static Production instance() {
```
armiol, line 42: Let's call it `type()`.

Don't forget to update the code samples in the documentation.

base/src/main/java/io/spine/base/Environment.java
```diff
@@ -69,101 +176,57 @@ public Environment createCopy() {
...
-            return this.tests;
-        }
-
-        // Check stacktrace for known frameworks.
-        String stacktrace = Throwables.getStackTraceAsString(new RuntimeException(""));
-        if (stacktrace.contains("org.junit")
-                || stacktrace.contains("org.testng")) {
-            this.tests = true;
-            return true;
+    public boolean is(EnvironmentType type) {
```
armiol, line 187: Let's bring back `isTests()` and `isProduction()` and mark them `@Deprecated` in favor of `... .enabled()`.
It's a proper way of propagating the public API changes.

base/src/main/java/io/spine/base/Tests.java
```diff
@@ -0,0 +1,115 @@
...
+     * Clears the {@linkplain #ENV_KEY_TESTS environment variable used for test detection}.
+     */
+    static void clearTestingEnvVariable() {
+        System.clearProperty(ENV_KEY_TESTS);
+    }
+
+    /**
+     * Returns the singleton instance of this class.
+     */
+    public static Tests instance() {
```
armiol, line 98: Same as above.

serhii-lekariev:
@armiol, PTAL again.

PR #540, opened 2020-06-06
Fix Pub cache location on Windows

dmdashenkov:
Recently, the Pub cache location on Windows has been [changed](https://github.com/dart-lang/pub/blob/82e60a3dcb3afe753563e7d304827fb650bc4833/lib/src/system_cache.dart#L38).

In this PR we bring our Gradle plugin for Dart projects up to the speed with the Pub tool.


dmdashenkov:
@armiol, PTAL.

PR #542, opened 2020-06-09
`EnvironmentType` design tweaks

serhii-lekariev:
This PR simplifies `EnvironmentType` and `Environment`.

#### `Environment`

Now, allows to check for the class with `boolean is(Class<? extends EnvironmentType>), instead of `boolean is(EnvironmentType)`.

#### `EnvironmentType` 

Now, no longer overrides `equals` and `hashCode` in suspicious ways. Instead, it simply acts as a stricter `interface` for a single `boolean enabled()` method.

Also, standard environment types are no longer singletons. We no longer encourage the extenders to be singletons as well.

Also resolves some Error Prone warnings.

### DIFF
--- a/base/src/main/java/io/spine/base/Environment.java
+++ b/base/src/main/java/io/spine/base/Environment.java
@@ -115,7 +115,7 @@ import static com.google.common.base.Preconditions.checkNotNull;
 public final class Environment {
 
     private static final ImmutableList<EnvironmentType> BASE_TYPES =
-            ImmutableList.of(Tests.type(), Production.type());
+            ImmutableList.of(new Tests(), new Production());
 
     private static final Environment INSTANCE = new Environment();
 
@@ -208,7 +208,7 @@ public final class Environment {
      */
     @Deprecated
     public boolean isTests() {
-        return is(Tests.type());
+        return is(new Tests());
     }
 
     /**
@@ -255,7 +255,7 @@ public final class Environment {
     @Deprecated
     @VisibleForTesting
     public void setToTests() {
-        this.currentEnvType = Tests.type();
+        this.currentEnvType = new Tests();
         Tests.enable();
     }
 
@@ -269,7 +269,7 @@ public final class Environment {
     @Deprecated
     @VisibleForTesting
     public void setToProduction() {
-        this.currentEnvType = Production.type();
+        this.currentEnvType = new Production();
         Tests.clearTestingEnvVariable();
     }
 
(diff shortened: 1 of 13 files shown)


serhii-lekariev:
@dmdashenkov, PTAL.

base/src/main/java/io/spine/base/Environment.java
```diff
@@ -78,19 +77,19 @@
  *
  *     static {
  *         Environment.instance()
- *                    .register(StagingEnvironmentType.type())
- *                    .register(LoadTestingType.type());
+ *                    .register(new StagingEnvironmentType())
```
dmdashenkov, line 80: Let's skip `Type` here, just like we don't call it `TestsType` and `ProductionType`.

base/src/main/java/io/spine/base/Environment.java
```diff
@@ -181,50 +180,22 @@ public Environment createCopy() {
...
-    private void determineCurrentType() {
-        for (EnvironmentType type : knownEnvTypes) {
-            if (type.enabled()) {
-                this.currentEnvType = type;
-                return;
-            }
-        }
+    public boolean is(Class<? extends EnvironmentType> typeClass) {
+        EnvironmentType currentType = cachedOrCalculated();
+        boolean result = currentType.getClass()
```
dmdashenkov, line 185: I'd go with `type.isInstance(..)`. That will give us an opportunity to create polymorphic types.

base/src/main/java/io/spine/base/Environment.java
```diff
@@ -181,50 +180,22 @@ public Environment createCopy() {
...
-    }
-
-    private void determineCurrentType() {
-        for (EnvironmentType type : knownEnvTypes) {
-            if (type.enabled()) {
-                this.currentEnvType = type;
-                return;
-            }
-        }
+    public boolean is(Class<? extends EnvironmentType> typeClass) {
```
dmdashenkov, line 183:
```suggestion
    public boolean is(Class<? extends EnvironmentType> type) {
```

base/src/main/java/io/spine/base/Environment.java
```diff
@@ -298,4 +269,22 @@ public void reset() {
...
+
+    private EnvironmentType currentType() {
+        for (EnvironmentType type : knownEnvTypes) {
+            if (type.enabled()) {
+                return type;
+            }
+        }
+
+        // `Production` is the default fallback.
+        return new Production();
```
dmdashenkov, line 288: Seems like an illegal state to me.

serhii-lekariev:
@dmdashenkov, PTAL again.

PR #544, opened 2020-06-09
Add `io.spine.testing` to the list of known testing libraries

serhii-lekariev:
`Tests` is the `EnvironmentType` that determines whether the system is currently undergoing tests. We check this by looking at the stack trace for mentions of known testing libraries. `io.spine.testing` is now among them, as `base` defines some classes in this package.


serhii-lekariev:
@dmdashenkov, PTAL.

PR #545, opened 2020-06-10
Register `EnvironmentType`s using their classes

serhii-lekariev:
This PR introduces several changes to `Environment` and `EnvironmentType`s.

- `Tests` and `Production` now have `package-private` constructors;
- `Environment` now caches the class of the active `EnvironmentType` instead of an instance;
- `Environment` now exposes an `@Internal` method to register an `EnvironmentType` by class. Classes to be registered using this method must have a parameterless constructor. This limitation is caused by the fact that we have to instantiate the specified `EnvironmentType`.

This leads to a signature change: `public EnvironmentType type()` -> `public Class<? extends EnvironmentType> type()`.
It also allows to add a `setTo(Class<? extends EnvironmentType>)` method, which is now necessary, as users can no longer `setTo(new Tests())`, as `Tests` now has a package-private constructor.

#### `Invokables`

This PR also extracts some of the reflection-related utilities into an `Invokables` class.

Users of `Methods` may found the utilities there.


serhii-lekariev:
@armiol, PTAL again.

serhii-lekariev:
@armiol, PTAL again.

serhii-lekariev:
@armiol, PTAL again.

serhii-lekariev:
@armiol, PTAL.

base/src/main/java/io/spine/base/Environment.java
```diff
@@ -159,6 +192,27 @@ public Environment register(EnvironmentType environmentType) {
         return this;
     }
 
+    /**
+     * Remembers the specified environment type, allowing {@linkplain #is(Class) to
+     * determine whether it's enabled} later.
+     *
+     * <p>The specified {@code type} must have a parameterless constructor.
+     * {@linkplain EnvironmentType#enabled() activity} of the specified environment type is going
```
armiol, line 200: `activity` should start from the capital letter. But I am not sure I understand this sentence:
```
activity of the specified environment type is going to be checked against an instance created by invoking the parameterless constructor.
```
Please rewrite into two or more _simple_ sentences.

base/src/main/java/io/spine/base/Environment.java
```diff
@@ -104,6 +107,36 @@
...
+ *     // is `true`.
+ *     assertThat(environment.is(AwsLambda.class)).isTrue();
+ *
+ *     environment.reset();
+ *
+ *     // When `reset` explicitly, cached value is erased.
+ *     assertThat(environment.is(AwsLambda.class)).isFalse();
+ * </pre>
+ *
+ * <p>{@linkplain #setTo(EnvironmentType) explicitly setting} the environment type overrides
```
armiol, line 137:
```
When set {@linkplain ... explicitly}, ...
```

Notice the capital letter.

base/src/main/java/io/spine/base/Environment.java
```diff
@@ -104,6 +107,36 @@
  * }
  * </pre>
  *
+ * <h1>Caching</h1>
+ * <p>{@code Environment} caches the {@code EnvironmentType} once its calculated.
```
armiol, line 111: For the sake of readability of this `.java` file, please add an empty line above.

base/src/main/java/io/spine/base/Environment.java
```diff
@@ -104,6 +107,36 @@
...
+ * This means that if one environment type has been found to be active, its instance is saved.
+ * If later it becomes logically inactive, e.g. the environment variable that's used to check the
+ * environment type changes, {@code Environment} is still going to return the cached value, unless
```
armiol, line 114: Don't use complex sentences.

```
... going to return the cached value. To overwrite the value (do something). Also, the value may be {@linkplain ... reset}.
```

serhii-lekariev:
@armiol, PTAL again.

base/src/test/java/io/spine/base/EnvironmentTest.java
```diff
@@ -217,32 +232,63 @@ void determineUsingType() {
...
+        assertThat(environment.is(Staging.class)).isTrue();
     }
 
+    @Test
+    @DisplayName("cache the `Tests` environment type")
+    void cacheTests() throws InterruptedException {
+        AtomicBoolean envCached = new AtomicBoolean(false);
+        assertThat(environment.is(Tests.class));
+        Thread thread = new Thread(() -> {
+            /*
```
armiol, line 275: Please convert this to the method-level comment.

Don't use inline comments except for emergency cases.

serhii-lekariev:
@armiol, PTAL.

base/src/main/java/io/spine/base/Environment.java
```diff
@@ -299,19 +299,26 @@ public void restoreFrom(Environment copy) {
 
     /**
      * Sets the current environment type to {@code type.getClass()}. Overrides the current value.
+     *
+     * <p>Calls {@link #register(EnvironmentType)} internally.
```
armiol, line 303:
```
If the supplied type was not {@linkplain #register(EnvironmentType) registered} previously, it is registered.
```

base/src/main/java/io/spine/base/Environment.java
```diff
@@ -299,19 +299,26 @@ public void restoreFrom(Environment copy) {
...
-        this.currentEnvType = checkNotNull(type).getClass();
+        checkNotNull(type);
+        register(type);
+        this.currentEnvType = type.getClass();
     }
 
     /**
      * Sets the current environment type to the specified one. Overrides the current value.
+     *
+     * <p>Calls {@link #register(Class)} internally.
```
armiol, line 315: Same as above.

Issue #535, opened 2020-05-24
Automatic updates of the Spine libraries


armiol:
At the moment the interdependencies between Spine libraries form a dependency tree, with the `base` library being the root:

```
`base` <═╗
  ^ ^    ╠═ `time` <═╗
  ║ │    ║           ║
  ║ │    ╚═══════════╚═ `core-java` <═╗<╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╌╮
  ║ │                                 ║                  ╎
  ║ ├─────────────────────────────────╚═ `gcloud-java`   ╎
  ║ │                                 ║                  ╎
  ║ ├─────────────────────────────────╚═ `jdbc-storage`  ╎
  ║ │                                 ║                  ╎
  ║ ╰─────────────────────────────────╚═ `web` <╌╌╌╌╌╌╌╌╌┤
  ║                                                      ╎
  ╚══════════════════════════════════════════════════════╰ `bootstrap


 <═══ - both compile- and build-time dependencies
 <─── — build-time dependencies
 <╌╌╌ - run-time-only dependencies

 ```

As one of the libraries gets updated, the dependants should be updated as well. Clearly, new releases of `base` library trigger the updates all down the tree.

In the scope of this task, the updates in the chain `base` <═ `time` <═ `core-java` should be automated. 

**Propagating `base` updates**

As soon as the `base` is updated to the version `X.Y.Z` in its `master`, `time` and `core-java` repositories should get new pull requests.

* In each of these PRs the version of the `base` library should be updated to `X.Y.Z`. 

* The own version of `core-java` and `time` should be set to either `X.Y.Z` (to match the `base`) or to their own incremental versions — whichever is greater. 

* As long as `core-java` depends on `time` as well, `core-java` should refer the _new_ version of `time` library. In this way no manual modification of the PR will be needed, once `time` is merged and published.

**Use case 1:**

The `base` library is updated to `1.5.12`.
The `core-java` library has `1.5.7` and `time` has `1.5.9`.

(Version update script is launched)

The `time` receives a new PR; the dependency onto `base` is set to `1.5.12`. The version of `time` is set to `1.5.12` as well.
The `core-java` receives a new PR, the dependency onto `base` is set to `1.5.12`. The version of `core-java` is set to `1.5.12` as well.

**Use case 2:**

The `base` library is updated to `1.5.10`.
The `core-java` library has `1.5.14` and `time` has `1.5.4`.

(Version update script is launched)

[...]

PR #551, opened 2020-07-17
Structure the `version` file

serhii-lekariev:
This PR is a part of a group of PRs that introduce consistency to how the library versions are declared.

Before, `versions` Gradle did not have any necessary structure - as long as the necessary properties were added and used in `build.gradle` they could do so in any way.

Simplifying version parsing and manipulation makes it easier to automate publishing and updating the libraries, see https://github.com/SpineEventEngine/base/issues/535 and https://github.com/SpineEventEngine/publishing.


serhii-lekariev:
See https://github.com/SpineEventEngine/base/pull/557.

Issue #553, opened 2020-08-05
`Time.currentTime()` returns invalid values when run on Java 11


nick-dolgiy:
Starting from Java 9 the implementation of the `java.time.Clock` was changed to be able to provide time in the precision that the underlying clock may offer, here is the [issue](https://bugs.openjdk.java.net/browse/JDK-8068730) on it.

Now that we have a microsecond precision of the `java.time.Clock` it sometimes leads to the problem that `Time.currentTime()` that emulates nanoseconds returns the `Timestamp` where the amount of nanoseconds is bigger than allowed, and as the result `Timestamps.checkValid()` fails for this value.

### How to reproduce

- Create the Spine-based project that can be compiled and run under the Java 11.
- Run the following code:
``` Java
for (int i = 0; i < 10000000; i++) {
    Timestamps.checkValid(Time.currentTime());
}
```
This code will produce the exception:
``` Stacktrace
java.lang.IllegalArgumentException: Timestamp is not valid. See proto definition for valid values. Seconds (1596640408) must be in range [-62,135,596,800, +253,402,300,799]. Nanos (1000000000) must be in range [0, +999,999,999].
	at com.google.protobuf.util.Timestamps.checkValid(Timestamps.java:173)
```
### Possible solution

Drop the microseconds precision if microseconds are returned from the system clock in the implementation of the `Time.currentTime()` method.

PR #554, opened 2020-08-06
Truncate time to milliseconds

nick-dolgiy:
If we run a spine-based project under the Java 11 platform we randomly may encounter the issue when `Time.currentTime()` returns an invalid timestamp.

The issue is that `Time` uses the `SystemTimeProvider` by default, which rely on the fact that `Instant.now()` returns time in millisecond precision, that was true for Java 8. The `SystemTimeProvider` emulates nanoseconds in order to split in time events that happened in the same millisecond, but if we have time with increased precision it leads us to the situation when the resulting amount of nanoseconds with emulation is bigger than allowed. This happens because starting from Java 9 `Instant.now()` may return time with increased precision if the underlying clock supports it. This issue also described in #553.

In this PR I made the `SystemTimeProvider` explicitly truncate time to the millisecond precision.

### DIFF
--- a/base/src/main/java/io/spine/base/Time.java
+++ b/base/src/main/java/io/spine/base/Time.java
@@ -153,12 +153,13 @@ public final class Time {
          * {@inheritDoc}
          *
          * <p>Starting from Java 9 the precision of time may differ from platform to platform and
-         * depends on the underlying clock used by the JVM, the
-         * <a href='https://bugs.openjdk.java.net/browse/JDK-8068730'>issue</a> on this in the
-         * JDK bug system. To be consistent on different platforms and use {@link IncrementalNanos}
-         * without risk of overflowing the {@code nano} values this method intentionally calculates
-         * time in millisecond precision even if the underlying clock may offer more precise
-         * values.
+         * depends on the underlying clock used by the JVM. See the corresponding
+         * <a href='https://bugs.openjdk.java.net/browse/JDK-8068730'>issue</a> on this matter.
+         *
+         * <p>In order to provide consistent behavior on different platforms and avoid the overflow
+         * of the emulated nanosecond value, this method intentionally uses the system time
+         * in millisecond precision. Therefore even if the system clock may offer more precise
+         * values, the {@code System.currentTimeMillis()} is used as a base for the returned values.
          */
         @Override
         public Timestamp currentTime() {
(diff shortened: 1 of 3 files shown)


nick-dolgiy:
@armiol PTAL.

nick-dolgiy:
@armiol PTAL again

base/src/main/java/io/spine/base/Time.java
```diff
@@ -140,21 +140,37 @@ default ZoneId currentZone() {
...
         private SystemTimeProvider() {
         }
 
+        /**
+         * {@inheritDoc}
+         *
+         * <p>Starting from Java 9 the precision of time may differ from platform to platform and
+         * depends on the underlying clock used by the JVM, the
+         * <a href='https://bugs.openjdk.java.net/browse/JDK-8068730'>issue</a> on this in the
+         * JDK bug system. To be consistent on different platforms and use {@link IncrementalNanos}
```
armiol, line 158: `JDK issue tracker`.

base/src/main/java/io/spine/base/Time.java
```diff
@@ -140,21 +140,37 @@ default ZoneId currentZone() {
     @VisibleForTesting
     static class SystemTimeProvider implements Provider {
 
+        public static final int NANOSECONDS_IN_MILLISECOND = 1_000_000;
```
armiol, line 143: Let's hide this constant.

Also, please look around. I would think there is a constant like this somewhere, already defined for you.

base/src/main/java/io/spine/base/Time.java
```diff
@@ -178,7 +194,8 @@ public Timestamp currentTime() {
      * <p>The returned nanosecond value starts at {@code 0} and never exceeds {@code 999 999}.
      * It is designed to keep the millisecond value provided by a typical-JVM system clock intact.
      *
-     * <p>The nanosecond value is reset for each new passed {@link Instant} value.
+     * <p>The nanosecond value is reset for each new passed {@code seconds} and {@code nanos}
+     * values.
      * That allows to receive {@code 1 000} distinct time values per millisecond.
```
armiol, line 199: Please move this line to the previous one.

base/src/main/java/io/spine/base/Time.java
```diff
@@ -140,21 +140,37 @@ default ZoneId currentZone() {
...
         static final Provider INSTANCE = new SystemTimeProvider();
 
         /** Prevent instantiation from outside. */
         private SystemTimeProvider() {
         }
 
+        /**
+         * {@inheritDoc}
+         *
+         * <p>Starting from Java 9 the precision of time may differ from platform to platform and
```
armiol, line 155: @nick-dolgiy 

Please see my suggested edit below.

```
<p>Starting from Java 9 the precision of time may differ from platform to platform and depends 
on the underlying clock used by the JVM. See the corresponding
 <a href='https://bugs.openjdk.java.net/browse/JDK-8068730'>issue</a> on this matter.

<p>In order to provide consistent behavior on different platforms and avoid the overflow 
of the emulated nanosecond value, this method intentionally uses the system time 
with the millisecond precision. Therefore even if the system clock may offer more precise values, 
the {@code System.currentTimeMillis()} is used as a base for the returned values.
```

nick-dolgiy:
@armiol PTAL again.

PR #555, opened 2020-08-19
Publish API doc

dmdashenkov:
This PR configures API doc publishing for the modules in this repository.


dmdashenkov:
Cancel this for now. Publishing our docs requires `spine-javadoc-filter`, which is a part of this repository. As this change was required for a single guide page, we can live without it.

Issue #552, opened 2020-07-17
Protobuf compiler is downloaded to the `.spine` under an included project


alexander-yevsyukov:
Currently it is not possible to create [composite builds](https://docs.gradle.org/current/userguide/composite_builds.html) of Spine-based projects because of the following problem. 

When a Spine-based project is added to a composite-build project via `includeBuild()`, and the project is evaluated during the composite build, we download the Protobuf Compiler plugin into the `.spine` directory, which is created _under_ the included project. But when the plugin is used, we try to find it in the _root_ project.

The workaround for this problem is to manually copy the `.spine` directory to the root of the composite build project.

alexander-yevsyukov:
@armiol @dmdashenkov FYI. I'll need this fixed for arranging the new build environment for spine.io development. Let's discuss.

PR #559, opened 2020-08-26
Simplify protoc plugin location

dmdashenkov:
Previously, Model Compiler downloaded the executable Jar with the `protoc` plugin into the `.spine` directory. This solution is bad for several reasons. The primary one — as it has been [discovered](https://github.com/SpineEventEngine/base/issues/552), it does not work with included builds.

In this PR we simplify the process of locating the Jar. Just like before, there are two scripts (for Windows and for *nix), which launch the Jar (via the `java` CLI tool). Unlike before, those scripts are not Maven artifacts but resources of the Model Compiler. Also, the script resources contain a placeholder for the path to the Jar.

To launch the plugin, Model Compiler resolves a dependency on the executable Jar and writes the appropriate script file into a temporary dir with the path to the resolved Jar. Then, Model Compiler configures the Protobuf Gradle plugin to launch the Spine `protoc` plugin via the written script file.

The benefit of the new solution is that all the FS paths are absolute, i.e. there is no difference in who requests the plugin: a root project, a subproject or an included build.


dmdashenkov:
@armiol, PTAL.

tools/plugin-base/src/main/java/io/spine/tools/gradle/ProtocConfigurationPlugin.java
```diff
@@ -103,32 +99,13 @@ private void configureProtocTasks(GenerateProtoTaskCollection tasks) {
      *
      * @param plugins
      *         container of all plugins
+     * @param project
```
armiol, line 102: Please have a description. In particular, in the context of the current issue, it may be either a root project or a project of a child Gradle module.

armiol (approved):
@dmdashenkov LGTM with a comment.

PR #560, opened 2020-08-26
Verify number validation options

dmdashenkov:
Starting from now, we will check that `(min)`, `(max)`, and `(range)` constraints on whole numbers (`int32`, etc.) are themselves whole numbers.

Also, some leftover debug output is trimmed.


dmdashenkov:
@armiol, PTAL.

armiol (approved):
@dmdashenkov LGTM. However, please respect https://github.com/SpineEventEngine/base/pull/561 in this changeset as well. We need those structural changes for the publishing procedure.

PR #562, opened 2020-09-01
Gradle task ordering

dmdashenkov:
In this PR we make the Gradle tasks declared by Model Compiler have more strict bounds to the task graph.

Previously, some tasks were only bound from one side, i.e. a task must have been executed after an existing one. This is insufficient, since, in some cases, such tasks might be moved in the task graph in such a way that their result it not captured by another task. For example, the task which generates a descriptor reference file (`desc.ref`) could've been executed after `processResources`, which would lead to the `desc.ref` file missing from the classpath at runtime.

In this PR we add more constraints to make sure that the tasks are ordered properly.


dmdashenkov:
@dmitrykuzmin, PTAL.

dmitrykuzmin (approved):
LGTM

PR #564, opened 2020-09-08
Query API for Entity states and Protobuf records

armiol:
This changeset brings the new API for building the queries for Entity states and plain records defined as Protobuf messages. Effectively, this API enhances the code previously used for entity columns. 

The framework version is updated to `2.0.0-jdk8.SNAPSHOT.2`. The reference to the `config` repository is set to `config/2.x-jdk8-master`.

# Generic parameters for the Entity state messages

The `EntityState` interface is now parameterized with the type of the Entity identifier. The code generated by the Spine compiler for the messages with `(entity)` option uses the first field of the message to obtain the value.

Also, the `EntityColumn` s are now parameterized with the type of the Entity state and the type of the column value. 

Example.

```proto

message ProjectDetails {
  option (entity).kind = PROJECTION;

  ProjectId id = 1;
  ProjectName name = 2 [(column) = true];
  int32 task_count = 3 [(column) = true];
}
```

For this message the Spine generation routines produce the `EntityColumn`s as follows:

```java
public final class ProjectDetails extends /* .. */
                        implements /* .. */, EntityState<ProjectId> {

  // ...

  public static final class Column {

    private Column() {
      // Prevent instantiation.
    }

    // Returns the "name" column.
    public static EntityColumn<ProjectDetails, ProjectName> name(){...}}

    // Returns the "task_count" column.
    public static EntityColumn<ProjectDetails, Integer> taskCount() {...}}

    // Returns all the column definitions for this type.
    public static ImmutableSet<EntityColumn<ProjectDetails, ?>> definitions()} {
        Set<EntityColumn<ProjectDetails, ?>> result = new HashSet<>();
        result.add(name());
        result.add(taskCount());
        return ImmutableSet.copyOf(result);
    }
  }
}
```

# Query API

`io.spine.query` package has been introduced. It describes the Query API for Protobuf Message records which are used to store arbitrary messages and Entity states. 

## Entity Query API

For the Messages representing Entity states the `Query` and `QueryBuilder` nested classes are now generated.

### Columns in `QueryBuilder`

With the power of the code generation, each QueryBuilder becomes a strongly-typed DSL which only accepts the columns of the target Entity. Also, the values to use in predicates are also matched to the column types at compile-time.

Given the message

```proto
message ProjectView {
   option (entity).kind = PROJECTION;

   // The identifier of the project.
   //
   // This field is not named `id` to ensure that the code generation
   // picks up the non-standard field name.
   //
   ProjectId project = 1;

   ProjectName project_name = 2 [(column) = true];

   Project.Status status = 3 [(column) = true];

   UserView assignee = 4;

   int32 days_since_started = 5 [(column) = true];

   bool was_reassigned = 6 [(column) = true];

   repeated string tag = 7;
}
```

the `ProjectView` class obtains `ProjectView.Query` and `ProjectView.QueryBuilder` classes along with the static factory method `newQuery(): QueryBuilder`. The code below

```java
ProjectView.Query query = ProjectView.query()
       .status().isNot(Status.CREATED)
       .daysSinceStarted().isGreaterThan(5)
       .build();
```

creates a query for all the `ProjectView`s for the projects which status is not `CREATED` and which were started more than five days ago.

Disjunction (or OR statements) are also supported:

```java
Either<ProjectView.QueryBuilder> startedMoreThanMonthAgo =
       project -> project.daysSinceStarted().isGreaterThan(30);
Either<ProjectView.QueryBuilder> isDone =
       project -> project.status().is(Status.DONE);

ProjectView.Query query =
       ProjectView.query()
                  .either(startedMoreThanMonthAgo, isDone)
                  .build();
```

This piece produces a query targeting the projects which are either done or started more than thirty days ago. 

### Identifiers

The first field of the Message is treated as an identifier. It's important to understand that Spine treats the field declared first in the order of reading. And not the one with the least index value.

The name of the ID field is preserved and is exposed for querying:

```java
ProjectView.Query query = ProjectView.query()
                                     .project().is(expectedId)
                                     .build();
```

Or, for the multiple IDs:

```java
ImmutableSet<ProjectId> expectedValues = // ...;
ProjectView.Query query = ProjectView.query()
                                     .project().in(expectedValues)
                                     .build();
```

### Custom Columns

In addition to the columns declared in the Entity state, the query API allows declaring the custom columns. This is intended to be used for lifecycle and version columns residing in `core-java`. Here is an excerpt from the `base` test code:

```java
/**
* The column presumably defined as a custom column for querying and storage.
*/
class ArchivedColumn extends CustomColumn<EntityWithLifecycle, Boolean> {

   @Override
   public ColumnName name() {
       return ColumnName.of("archived");
   }

   @Override
   public Class<Boolean> type() {
       return Boolean.class;
   }

   @Override
   public Boolean valueIn(EntityWithLifecycle source) {
       return source.isArchived();
   }
}
```

Where `EntityWithLifecycle` is a mock interface serving as a source of the lifecycle values.
In the real-world usage from `core-java` codebase, it is replaced by the `Entity` class.

```java
/**
* Custom columns that define entity lifecycle.
*
* <p>Used in the smoke testing of entity query builders.
*/
public enum Lifecycle {

   ARCHIVED(new ArchivedColumn()),

   DELETED(new DeletedColumn());

   private final CustomColumn<?, Boolean> column;

   Lifecycle(CustomColumn<?, Boolean> column) {
       this.column = column;
   }
   
   /** Returns the column declaration. */
   public CustomColumn<?, Boolean> column() {
       return column;
   }
}
```

Such an approach allows defining the singleton instances of the columns and using them in querying:

```java
ProjectView.Query queryForDeleted =
       ProjectView.query()
                  .where(DELETED.column(), true)
                  .build();
```

### Sorting and Limit

Query API also allows specifying the sorting order and the record limit for the queries:

```java
ProjectView.Query query = ProjectView.query()
       .sortAscendingBy(daysSinceStarted())
       .sortAscendingBy(projectName())
       .sortDescendingBy(wasReassigned())
       .limit(10)
       .build();
```

Here `daysSinceStarted()`, `projectName()` and `wasReassigned()` are three static methods generated in `ProjectView.Column` class. Each of them returns the declaration of the Entity column corresponding to the Message field with the `(column)` option.

### Field masks

It is possible to set the mask for each resulting Protobuf message:

```java
FieldMask mask = fieldMaskWith(status());
ProjectView.Query query = ProjectView.query()
                                     .withMask(mask)
                                     .build();
```

### Entity states with no columns

If a Proto message declares an Entity state with no columns, we still require at least one field declared. As described above, it is treated as an Entity identifier. Therefore, both `QueryBuilder` and `Query` are generated for this case. 

## Record Query API

End-users also store the Protobuf messages which aren’t marked as Entity states. In order to be able to query those, the `io.spine.query` package provides some helper utilities.

### Declaring Columns

As long as `(column)` annotation is specific to Entity state messages, there should be some way to specify the storage columns for plain `Message` records. `base` library now provides a `RecordColumn<TypeOfMessage, ValueType>` type. With it, the declaration of the columns may look as shown below.

Example.

The Proto message which instances need to be stored:

```proto
// Describes a car manufacturer as a company.
message Manufacturer {

   ManufacturerId id = 1 [(required) = true, (validate) = true];

   // ISIN value.
   Isin isin = 2 [(required) = true, (validate) = true];

   // When the company was founded.
   google.protobuf.Timestamp when_founded = 3 [(required) = true];

   // How many employees work in the company.
   int32 number_of_employees = 4 [(min).value = "1"];

   // The type of the company as a legal entity.
   LegalEntityType type = 5;

   // If traded, which stock symbols the company uses.
   repeated TradeStockSymbol stock_symbol = 6;
}
```

End-user may want to choose some pieces of the Message data to be stored as distinct columns to use them in querying later:

```java
/**
* Defines the columns for {@link Manufacturer} message record.
*/
public static final class ManufacturerColumns {

   public static final RecordColumn<Manufacturer, String> isin =
           new RecordColumn<>("isin", 
                              String.class, 
                              (r) -> r.getIsin()
                                      .getValue());

   public static final RecordColumn<Manufacturer, Timestamp> whenFounded =
           new RecordColumn<>("when_founded", 
                              Timestamp.class, 
                              Manufacturer::getWhenFounded);

   public static final RecordColumn<Manufacturer, Boolean> isTraded =
           new RecordColumn<>("is_traded",
                              Boolean.class,
                              (r) -> !r.getStockSymbolList()
                                       .isEmpty());

   private ManufacturerColumns() {
   }
}
```

### `RecordQuery` and `RecordQueryBuilder`

Once the columns are declared, they may be used in a Query. Spine provides a `RecordQueryBuilder` for this purpose:

```java
RecordQuery<ManufacturerId, Manufacturer> query =
       RecordQuery.newBuilder(ManufacturerId.class, Manufacturer.class)
                  .where(isin).isNot("JP 3633400001")
                  .where(whenFounded).isLessOrEqualTo(THURSDAY)
                  .where(isTraded).is(true)
                  .build();
```

Where `isin`, `whenFounded`, and `isTraded` are the statically imported definitions from the `ManufacturerColumns` class defined above.

### Sorting, Limit, and Field Masks

In the same manner as with the Entity Query API, the `RecordQueryBuilder` supports setting the sorting order, the maximum number of the records, and the field masks.

```java
FieldMask mask = fieldMaskWith(isTraded);
RecordQuery<ManufacturerId, Manufacturer> query =
       RecordQuery.newBuilder(ManufacturerId.class, Manufacturer.class)
                  .withMask(mask)
                  .sortAscendingBy(whenFounded)
                  .sortAscendingBy(isin)
                  .sortDescendingBy(isTraded)
                  .limit(10)
                  .build();
```

Where `isin`, `whenFounded` and `isTraded` are the statically imported definitions of the record columns from `ManufacturerColumns` type declared above.

### DIFF
--- a/base/src/main/java/io/spine/query/AbstractQuery.java
+++ b/base/src/main/java/io/spine/query/AbstractQuery.java
@@ -79,10 +79,7 @@ abstract class AbstractQuery<I, R extends Message, P extends SubjectParameter<R,
      * <p>Checks that if the limit is set, at least one ordering directive is present as well.
      */
     AbstractQuery(AbstractQueryBuilder<I, R, P, ?, ?> builder) {
-        this.subject = new Subject<>(builder.whichIds(),
-                                     builder.whichIdType(),
-                                     builder.whichRecordType(),
-                                     builder.predicates());
+        this.subject = new Subject<>(builder);
         this.ordering = checkNotNull(builder.ordering());
         this.mask = builder.whichMask().orElse(FieldMask.getDefaultInstance());
         limit = ensureLimit(builder.whichLimit());
--- a/base/src/main/java/io/spine/query/AbstractQuery.java
+++ b/base/src/main/java/io/spine/query/AbstractQuery.java
@@ -49,20 +49,20 @@ abstract class AbstractQuery<I, R extends Message, P extends SubjectParameter<R,
     private final Subject<I, R> subject;
 
     /**
-     * List of ordering directives which define the order of records in the query results.
+     * List of sorting directives which define the order of records in the query results.
      *
      * <p>Directives are applied one by one, starting with the first one. The second one
-     * and all consecutive directives specify the order of records, which are considered
-     * equal by the previous {@code OrderBy} directives.
+     * and all consecutive directives specify the sorting order of records, which are considered
+     * equal by the previous {@code SortBy} directives.
      */
-    private final ImmutableList<OrderBy<?, R>> ordering;
+    private final ImmutableList<SortBy<?, R>> sorting;
 
     /**
      * The maximum number of records in the query results.
      *
      * <p>If not set, all matching records are returned.
      *
-     * <p>This field may only be used if at least one {@link OrderBy ordering directive} is set.
+     * <p>This field may only be used if at least one {@link SortBy sorting directive} is set.
      */
     private final @Nullable Integer limit;
 
@@ -76,23 +76,23 @@ abstract class AbstractQuery<I, R extends Message, P extends SubjectParameter<R,
     /**
      * A common contract for the constructors of {@code AbstractQuery} implementations.
      *
-     * <p>Checks that if the limit is set, at least one ordering directive is present as well.
+     * <p>Checks that if the limit is set, at least one sorting directive is present as well.
      */
     AbstractQuery(AbstractQueryBuilder<I, R, P, ?, ?> builder) {
         this.subject = new Subject<>(builder);
-        this.ordering = checkNotNull(builder.ordering());
+        this.sorting = checkNotNull(builder.sorting());
         this.mask = builder.whichMask().orElse(FieldMask.getDefaultInstance());
         limit = ensureLimit(builder.whichLimit());
     }
 
     /**
-     * Checks that if the limit is set, at least one ordering directive is specified as well.
+     * Checks that if the limit is set, at least one sorting directive is specified as well.
      *
      * @return the value of query limit, {@code null}-able, as the limit may not be set
      */
     private @Nullable Integer ensureLimit(@Nullable Integer limit) {
-        checkState(limit == null || !ordering.isEmpty(),
-                      "Query limit must be used with at least one ordering directive set.");
+        checkState(limit == null || !sorting.isEmpty(),
+                      "Query limit must be used with at least one sorting directive set.");
         return limit;
     }
 
@@ -102,8 +102,8 @@ abstract class AbstractQuery<I, R extends Message, P extends SubjectParameter<R,
     }
 
     @Override
-    public final ImmutableList<OrderBy<?, R>> ordering() {
-        return ordering;
+    public final ImmutableList<SortBy<?, R>> sorting() {
+        return sorting;
     }
 
     @Override
@@ -126,13 +126,13 @@ abstract class AbstractQuery<I, R extends Message, P extends SubjectParameter<R,
         }
         AbstractQuery<?, ?, ?> query = (AbstractQuery<?, ?, ?>) o;
         return subject.equals(query.subject) &&
-                ordering.equals(query.ordering) &&
+                sorting.equals(query.sorting) &&
                 Objects.equals(limit, query.limit) &&
                 mask.equals(query.mask);
     }
 
     @Override
     public int hashCode() {
-        return Objects.hash(subject, ordering, limit, mask);
+        return Objects.hash(subject, sorting, limit, mask);
     }
 }
--- a/base/src/main/java/io/spine/query/AbstractQueryBuilder.java
+++ b/base/src/main/java/io/spine/query/AbstractQueryBuilder.java
@@ -38,6 +38,8 @@ import java.util.Optional;
 import static com.google.common.base.Preconditions.checkNotNull;
 import static com.google.common.collect.ImmutableList.toImmutableList;
 import static com.google.protobuf.util.FieldMaskUtil.fromStringList;
+import static io.spine.query.Direction.ASC;
+import static io.spine.query.Direction.DESC;
 import static io.spine.query.LogicalOperator.AND;
 import static io.spine.query.LogicalOperator.OR;
 import static io.spine.util.Preconditions2.checkPositive;
@@ -73,7 +75,7 @@ abstract class AbstractQueryBuilder<I,
 
     private QueryPredicate.Builder<R> currentPredicate = QueryPredicate.newBuilder(AND);
 
-    private final List<OrderBy<?, R>> ordering = new ArrayList<>();
+    private final List<SortBy<?, R>> sorting = new ArrayList<>();
 
     @MonotonicNonNull
     private Integer limit;
@@ -118,8 +120,8 @@ abstract class AbstractQueryBuilder<I,
     }
 
     @Override
-    public ImmutableList<OrderBy<?, R>> ordering() {
-        return ImmutableList.copyOf(ordering);
+    public ImmutableList<SortBy<?, R>> sorting() {
+        return ImmutableList.copyOf(sorting);
     }
 
     @Override
@@ -181,10 +183,17 @@ abstract class AbstractQueryBuilder<I,
 
     @Override
     @CanIgnoreReturnValue
-    public final B orderBy(RecordColumn<R, ?> column, Direction direction) {
+    public final B sortAscendingBy(RecordColumn<R, ?> column) {
         checkNotNull(column);
-        checkNotNull(direction);
-        ordering.add(new OrderBy<>(column, direction));
+        sorting.add(new SortBy<>(column, ASC));
+        return thisRef();
+    }
+
+    @Override
+    @CanIgnoreReturnValue
+    public final B sortDescendingBy(RecordColumn<R, ?> column) {
+        checkNotNull(column);
+        sorting.add(new SortBy<>(column, DESC));
         return thisRef();
     }
 
@@ -257,13 +266,13 @@ abstract class AbstractQueryBuilder<I,
     }
 
     /**
-     * Adds the ordering directive.
+     * Adds the sorting directive.
      *
      * @return this instance of query builder, for chaining
      */
-    protected final B addOrdering(OrderBy<?, R> value) {
+    protected final B addSorting(SortBy<?, R> value) {
         checkNotNull(value);
-        ordering.add(value);
+        sorting.add(value);
         return thisRef();
     }
 }
--- a/base/src/main/java/io/spine/query/AbstractQueryBuilder.java
+++ b/base/src/main/java/io/spine/query/AbstractQueryBuilder.java
@@ -31,6 +31,7 @@ import org.checkerframework.checker.nullness.qual.Nullable;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.List;
 import java.util.Optional;
 
@@ -150,9 +151,8 @@ abstract class AbstractQueryBuilder<I,
     @Override
     @CanIgnoreReturnValue
     public final B withMask(String ...maskPaths) {
-        Class<R> recordType = whichRecordType();
-        FieldMask fieldMask = fromStringList(recordType, ImmutableList.copyOf(maskPaths));
-        return withMask(fieldMask);
+        ImmutableList<String> pathList = ImmutableList.copyOf(maskPaths);
+        return withMask(pathList);
     }
 
     @Override
@@ -161,6 +161,19 @@ abstract class AbstractQueryBuilder<I,
         ImmutableList<String> paths = Arrays.stream(fields)
                                               .map(Field::toString)
                                               .collect(toImmutableList());
+        return withMask(paths);
+    }
+
+    /**
+     * Sets the paths from the passed collection to apply as a field mask
+     * to each of the resulting records.
+     *
+     * <p>This method acts similar to {@link #withMask(String...)}.
+     *
+     * @return this instance of builder
+     * @see #withMask(String...)
+     */
+    final B withMask(Collection<String> paths) {
         Class<R> recordType = whichRecordType();
         FieldMask fieldMask = fromStringList(recordType, paths);
         return withMask(fieldMask);
--- a/base/src/main/java/io/spine/query/AbstractQueryBuilder.java
+++ b/base/src/main/java/io/spine/query/AbstractQueryBuilder.java
@@ -157,10 +157,10 @@ abstract class AbstractQueryBuilder<I,
 
     @Override
     @CanIgnoreReturnValue
-    public final B withMask(Field...fields) {
+    public final B withMask(Field... fields) {
         ImmutableList<String> paths = Arrays.stream(fields)
-                                              .map(Field::toString)
-                                              .collect(toImmutableList());
+                                            .map(Field::toString)
+                                            .collect(toImmutableList());
         return withMask(paths);
     }
 
--- a/base/src/main/java/io/spine/query/ComparisonOperator.java
+++ b/base/src/main/java/io/spine/query/ComparisonOperator.java
@@ -31,7 +31,7 @@ import static java.lang.String.format;
 /**
  * Defines how the queried records are compared against the desired parameter values.
  *
- * <h1><a name="supported_types"><strong>Supported field types</strong></a></h1>
+ * <h3><a name="supported_types"><strong>Supported field types</strong></a></h3>
  *
  * <p>The equality comparisons support the fields of all types. The operation is performed
  * via the {@link Objects#equals} method. A {@code null} reference is considered equal
--- a/base/src/main/java/io/spine/query/CustomColumn.java
+++ b/base/src/main/java/io/spine/query/CustomColumn.java
@@ -29,7 +29,7 @@ import io.spine.annotation.SPI;
  *
  * <p>End-users may choose to store some arbitrary data along with the record.
  * E.g. the time of entity creation or the role of the user created the record etc, that is,
- * something not included into the definition of record {@code Message}.
+ * something not included into the definition of the {@code Message} type of the record.
  *
  * <p>End-users would need to provide their own {@link CustomColumn} implementation.
  * When storing objects with custom columns, the values are fetched according
--- a/base/src/main/java/io/spine/query/CustomColumn.java
+++ b/base/src/main/java/io/spine/query/CustomColumn.java
@@ -27,11 +27,12 @@ import io.spine.annotation.SPI;
  * A column storing the data which is not directly declared as a field in the {@code Message}
  * of an {@link io.spine.base.EntityState EntityState} or a plain record.
  *
- * <p>End-users may choose to store some arbitrary data along with the record.
- * E.g. the time of entity creation or the role of the user created the record etc, that is,
- * something not included into the definition of the {@code Message} type of the record.
+ * <p>End-users may choose to store some arbitrary or computed on-the-fly data
+ * along with the record. E.g. the time of entity creation or the role of the user created
+ * the record etc. That is, something which isn't included into the definition
+ * of the {@code Message} type of the record.
  *
- * <p>End-users would need to provide their own {@link CustomColumn} implementation.
+ * <p>The framework users would need to provide their own {@link CustomColumn} implementation.
  * When storing objects with custom columns, the values are fetched according
  * to the {@link #valueIn(Object) valueIn(S)} implementation. In it, the {@code S} value represents
  * an arbitrary object serving as a source for the value.
--- a/base/src/main/java/io/spine/query/CustomCriterion.java
+++ b/base/src/main/java/io/spine/query/CustomCriterion.java
@@ -42,7 +42,7 @@ public final class CustomCriterion<S, V, B extends QueryBuilder<?, ?, ?, B, ?>> 
     private final CustomColumn<S, V> column;
 
     /**
-     * Creates a new instance of {@code CustomCriterion}.
+     * Creates a new instance.
      *
      * @param column
      *         the column for which the {@link CustomSubjectParameter} should be set
--- a/base/src/main/java/io/spine/query/CustomCriterion.java
+++ b/base/src/main/java/io/spine/query/CustomCriterion.java
@@ -21,7 +21,6 @@
 package io.spine.query;
 
 import com.google.errorprone.annotations.CanIgnoreReturnValue;
-import io.spine.annotation.Internal;
 
 import static com.google.common.base.Preconditions.checkNotNull;
 import static io.spine.query.ComparisonOperator.EQUALS;
@@ -35,8 +34,7 @@ import static io.spine.query.ComparisonOperator.EQUALS;
  * @param <V> the type of column values
  * @param <B> the type of query builder in scope of which this criterion exists
  */
-@Internal
-public final class CustomCriterion<S, V, B extends QueryBuilder<?, ?, ?, B, ?>> {
+final class CustomCriterion<S, V, B extends QueryBuilder<?, ?, ?, B, ?>> {
 
     private final B builder;
     private final CustomColumn<S, V> column;
--- a/base/src/main/java/io/spine/query/CustomCriterion.java
+++ b/base/src/main/java/io/spine/query/CustomCriterion.java
@@ -30,9 +30,12 @@ import static io.spine.query.ComparisonOperator.EQUALS;
  *
  * <p>The custom parameters are set as desired values for the {@link CustomColumn}s.
  *
- * @param <S> the type of objects which serve as a source for the column values
- * @param <V> the type of column values
- * @param <B> the type of query builder in scope of which this criterion exists
+ * @param <S>
+ *         the type of objects which serve as a source for the column values
+ * @param <V>
+ *         the type of column values
+ * @param <B>
+ *         the type of query builder in scope of which this criterion exists
  */
 final class CustomCriterion<S, V, B extends QueryBuilder<?, ?, ?, B, ?>> {
 
--- a/base/src/main/java/io/spine/query/CustomSubjectParameter.java
+++ b/base/src/main/java/io/spine/query/CustomSubjectParameter.java
@@ -21,16 +21,19 @@
 package io.spine.query;
 
 /**
- * A subject parameter, which defines the condition for something different than a declared field
- * of a record or an {@link io.spine.base.EntityState EntityState}.
+ * A subject parameter for a query, which defines a condition for something different than
+ * a value of a field declared in a {@code Message} type of the record.
  *
  * <p>In some cases, stored records or entities include {@linkplain CustomColumn custom columns},
  * which hold the properties not directly belonging to the record declaration. In order to include
  * such a column into a {@link Subject} definition, a {@code CustomQueryParameter} is used.
  *
- * @param <S> the type of objects serving as an origin for the parameter values
- * @param <V> the type of parameter values
+ * @param <S>
+ *         the type of objects serving as an origin for the parameter values
+ * @param <V>
+ *         the type of parameter values
  * @see CustomColumn
+ * @see Subject
  */
 public final class CustomSubjectParameter<S, V> extends SubjectParameter<S, Column<S, V>, V> {
 
--- a/base/src/main/java/io/spine/query/CustomSubjectParameter.java
+++ b/base/src/main/java/io/spine/query/CustomSubjectParameter.java
@@ -35,7 +35,7 @@ package io.spine.query;
 public final class CustomSubjectParameter<S, V> extends SubjectParameter<S, Column<S, V>, V> {
 
     /**
-     * Creates a new instance of the parameter.
+     * Creates a new instance.
      *
      * @param column
      *         column to use in the query
--- a/base/src/main/java/io/spine/query/CustomSubjectParameter.java
+++ b/base/src/main/java/io/spine/query/CustomSubjectParameter.java
@@ -21,12 +21,13 @@
 package io.spine.query;
 
 /**
- * A subject parameter for a query, which defines a condition for something different than
- * a value of a field declared in a {@code Message} type of the record.
+ * A query parameter, which defines a condition for a computed value stored along with
+ * the queried record.
  *
- * <p>In some cases, stored records or entities include {@linkplain CustomColumn custom columns},
- * which hold the properties not directly belonging to the record declaration. In order to include
- * such a column into a {@link Subject} definition, a {@code CustomQueryParameter} is used.
+ * <p>Custom parameters are applied to records storing {@linkplain CustomColumn custom
+ * columns} in addition to the columns corresponding to the fields declared in a {@code Message}
+ * type of the record. Such custom columns typically store computed values, which for some reason
+ * do not belong directly to the record declaration.
  *
  * @param <S>
  *         the type of objects serving as an origin for the parameter values
--- a/base/src/main/java/io/spine/query/Direction.java
+++ b/base/src/main/java/io/spine/query/Direction.java
@@ -21,9 +21,9 @@
 package io.spine.query;
 
 /**
- * The direction of ordering for the set of records by some field value.
+ * The direction of sorting for the set of records by some field value.
  *
- * @see OrderBy
+ * @see SortBy
  */
 public enum Direction {
 
--- a/base/src/main/java/io/spine/query/Either.java
+++ b/base/src/main/java/io/spine/query/Either.java
@@ -28,8 +28,8 @@ import java.util.function.UnaryOperator;
  * A lambda serving to fill the current predicate {@linkplain QueryBuilder query builders}
  * with the parameters joined in {@linkplain LogicalOperator#OR disjunction}.
  */
-@FunctionalInterface
 @Immutable
+@FunctionalInterface
 public interface Either<B extends QueryBuilder<?, ?, ?, ?, ?>> extends UnaryOperator<B> {
 
 }
--- a/base/src/main/java/io/spine/query/EntityCriterion.java
+++ b/base/src/main/java/io/spine/query/EntityCriterion.java
@@ -40,7 +40,7 @@ public final class EntityCriterion<S extends EntityState<?>,
 
 
     /**
-     * Creates a new instance of {@code EntityCriterion}.
+     * Creates a new instance.
      *
      * @param column
      *         the column which actual value to use later in querying
--- a/base/src/main/java/io/spine/query/EntityQuery.java
+++ b/base/src/main/java/io/spine/query/EntityQuery.java
@@ -84,7 +84,7 @@ public abstract class EntityQuery<I,
     private void doCopyTo(AbstractQueryBuilder<I, S, ?, ?, ?> destination) {
         copyIdParameter(destination);
         copyPredicates(destination);
-        copyOrdering(destination);
+        copySorting(destination);
         copyLimit(destination);
         copyMask(destination);
     }
@@ -127,11 +127,11 @@ public abstract class EntityQuery<I,
     }
 
     /**
-     * Copies the ordering directives from the current instance to the destination builder.
+     * Copies the sorting directives from the current instance to the destination builder.
      */
-    private void copyOrdering(AbstractQueryBuilder<I, S, ?, ?, ?> destination) {
-        for (OrderBy<?, S> sourceOrderBy : ordering()) {
-            destination.addOrdering(sourceOrderBy);
+    private void copySorting(AbstractQueryBuilder<I, S, ?, ?, ?> destination) {
+        for (SortBy<?, S> sourceSortBy : sorting()) {
+            destination.addSorting(sourceSortBy);
         }
     }
 }
--- a/base/src/main/java/io/spine/query/EntityQuery.java
+++ b/base/src/main/java/io/spine/query/EntityQuery.java
@@ -25,8 +25,10 @@ import com.google.protobuf.FieldMask;
 import io.spine.base.EntityState;
 
 /**
- * A common contract for the classes generated for each entity state type, which defines
- * how the entities of this type may be queried.
+ * A query for the records being the stored Protobuf messages, each declared
+ * as a state of an Entity.
+ *
+ * <p>See the {@code package-info.java} of this package for more details on usage.
  *
  * @param <I>
  *         the type of entity identifiers
--- a/base/src/main/java/io/spine/query/EntityQuery.java
+++ b/base/src/main/java/io/spine/query/EntityQuery.java
@@ -44,7 +44,7 @@ public abstract class EntityQuery<I,
     private final B builder;
 
     /**
-     * A common constructor contract for all {@code EntityQuery} implementations.
+     * Creates a new query according to the passed query builder.
      */
     protected EntityQuery(B builder) {
         super(builder);
--- a/base/src/main/java/io/spine/query/EntityQuery.java
+++ b/base/src/main/java/io/spine/query/EntityQuery.java
@@ -59,7 +59,7 @@ public abstract class EntityQuery<I,
     }
 
     /**
-     * Creates a {@link RecordQuery} instance with the same properties as this entity query.
+     * Creates a {@link RecordQuery} instance with the same attributes as this entity query.
      */
     public final RecordQuery<I, S> toRecordQuery() {
         Subject<I, S> subject = subject();
--- a/base/src/main/java/io/spine/query/EntityQueryBuilder.java
+++ b/base/src/main/java/io/spine/query/EntityQueryBuilder.java
@@ -90,9 +90,9 @@ public abstract class EntityQueryBuilder<I,
     @CanIgnoreReturnValue
     public final B withMask(SubscribableField... fields) {
         ImmutableList<String> paths = Arrays.stream(fields)
-                                              .map(f -> f.getField()
-                                                         .toString())
-                                              .collect(toImmutableList());
+                                            .map(f -> f.getField()
+                                                       .toString())
+                                            .collect(toImmutableList());
         return withMask(paths);
     }
 
--- a/base/src/main/java/io/spine/query/EntityQueryBuilder.java
+++ b/base/src/main/java/io/spine/query/EntityQueryBuilder.java
@@ -27,7 +27,7 @@ import java.util.function.Function;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 /**
- * An abstract base for builders of queries which are aimed to fetch the records of entity states.
+ * Base type for builders of queries which are aimed to fetch the records of entity states.
  *
  * @param <I>
  *         the type of entity identifiers
--- a/base/src/main/java/io/spine/query/EntityQueryBuilder.java
+++ b/base/src/main/java/io/spine/query/EntityQueryBuilder.java
@@ -20,11 +20,16 @@
 
 package io.spine.query;
 
+import com.google.common.collect.ImmutableList;
+import com.google.errorprone.annotations.CanIgnoreReturnValue;
 import io.spine.base.EntityState;
+import io.spine.base.SubscribableField;
 
+import java.util.Arrays;
 import java.util.function.Function;
 
 import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.collect.ImmutableList.toImmutableList;
 
 /**
  * Base type for builders of queries which are aimed to fetch the records of entity states.
@@ -71,6 +76,24 @@ public abstract class EntityQueryBuilder<I,
     public final <V> B where(CustomColumn<?, V> column, V value) {
         return column.in(thisRef())
                      .is(value);
+    }
+
+    /**
+     * Applies the paths of the passed fields as a field mask to each of the resulting records.
+     *
+     * <p>If the mask is not set, the query results contain the records as-is.
+     *
+     * <p>Any previously set mask values are overridden by this method call.
+     *
+     * @return this instance of query builder, for chaining
+     */
+    @CanIgnoreReturnValue
+    public final B withMask(SubscribableField... fields) {
+        ImmutableList<String> paths = Arrays.stream(fields)
+                                              .map(f -> f.getField()
+                                                         .toString())
+                                              .collect(toImmutableList());
+        return withMask(paths);
     }
 
     /**
--- a/base/src/main/java/io/spine/query/EntityQueryBuilder.java
+++ b/base/src/main/java/io/spine/query/EntityQueryBuilder.java
@@ -27,7 +27,7 @@ import java.util.function.Function;
 import static com.google.common.base.Preconditions.checkNotNull;
 
 /**
- * An abstract base for builders of an {@code EntityQuery} for a particular entity state type.
+ * An abstract base for builders of queries which are aimed to fetch the records of entity states.
  *
  * @param <I>
  *         the type of entity identifiers
--- a/base/src/main/java/io/spine/query/EntitySubjectParameter.java
+++ b/base/src/main/java/io/spine/query/EntitySubjectParameter.java
@@ -31,7 +31,7 @@ import static io.spine.query.ComparisonOperator.LESS_THAN;
 
 /**
  * A parameter defining how to query a record of an entity state by the value
- * of its {@link EntityColumn column}.
+ * of its {@linkplain EntityColumn column}.
  *
  * @param <S>
  *         the type of entity state
--- a/base/src/main/java/io/spine/query/EntitySubjectParameter.java
+++ b/base/src/main/java/io/spine/query/EntitySubjectParameter.java
@@ -30,7 +30,8 @@ import static io.spine.query.ComparisonOperator.LESS_OR_EQUALS;
 import static io.spine.query.ComparisonOperator.LESS_THAN;
 
 /**
- * A parameter defining how to query an entity by the value of its {@link EntityColumn}.
+ * A parameter defining how to query a record of an entity state by the value
+ * of its {@link EntityColumn column}.
  *
  * @param <S>
  *         the type of entity state
--- a/base/src/main/java/io/spine/query/EntitySubjectParameter.java
+++ b/base/src/main/java/io/spine/query/EntitySubjectParameter.java
@@ -22,13 +22,6 @@ package io.spine.query;
 
 import io.spine.base.EntityState;
 
-import static com.google.common.base.Preconditions.checkNotNull;
-import static io.spine.query.ComparisonOperator.EQUALS;
-import static io.spine.query.ComparisonOperator.GREATER_OR_EQUALS;
-import static io.spine.query.ComparisonOperator.GREATER_THAN;
-import static io.spine.query.ComparisonOperator.LESS_OR_EQUALS;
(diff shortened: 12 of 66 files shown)


alexander-yevsyukov:
> uses the first field of the message to obtain the value.

Is it the first field by the order of the declaration in the message type, or the field with index equal 1?

alexander-yevsyukov:
> `public static Set<EntityColumn<ProjectDetails, ?>> definitions()`

When do we want to change `Column.definitions()`? 
Can we remove a column, for example?
What if we return `ImmutableSet` instead?

base/src/main/java/io/spine/code/proto/ColumnOption.java
```diff
@@ -106,20 +103,11 @@ public static boolean isColumn(FieldDeclaration field) {
     }
 
     /**
-     * Returns {@code true} if the given message type is eligible for having columns.
-     *
-     * <p>Currently, only projection and process manager types can have columns.
-     *
-     * <p>The column presence for {@linkplain io.spine.option.EntityOption.Kind#ENTITY generic}
-     * entity types is also allowed for convenience of tests.
+     * Returns {@code true} if the given message type is declared as entity and may have columns.
```
dmdashenkov, line 106:
```suggestion
     * Returns {@code true} if the given message type is declared as an entity and may have columns.
```

base/src/main/java/io/spine/code/proto/ColumnOption.java
```diff
@@ -106,20 +103,11 @@ public static boolean isColumn(FieldDeclaration field) {
...
      */
     @SuppressWarnings("BooleanMethodIsAlwaysInverted") // For readability.
-    private static boolean eligibleForColumns(MessageType messageType) {
+    private static boolean declaresAsEntity(MessageType messageType) {
```
dmdashenkov, line 109: I suggest `declaresEntity(..)` for better readability. Or, citing the doc, `declaredAsEntity(..)`.

base/src/main/java/io/spine/query/Column.java
```diff
@@ -0,0 +1,70 @@
...
+package io.spine.query;
+
+import com.google.errorprone.annotations.Immutable;
+import com.google.protobuf.Message;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+import java.util.function.Function;
+
+/**
+ * A column of the some record residing in a storage.
```
dmdashenkov, line 30:
```suggestion
 * A column of some record residing in a storage.
```

base/src/main/java/io/spine/query/ComparisonOperator.java
```diff
@@ -0,0 +1,133 @@
...
+import java.util.Objects;
+
+import static java.lang.String.format;
+
+/**
+ * Defines how the queried records are compared against the desired parameter values.
+ *
+ * <h1><a name="supported_types"><strong>Supported field types</strong></a></h1>
+ *
+ * <p>All types of the fields are supported. The check is performed via the {@link Object#equals}
```
dmdashenkov, line 36: It seems like there's something missing from the first sentence. "All types of the fields are supported for equality comparisons.", or something in this spirit.

base/src/main/java/io/spine/query/CustomColumn.java
```diff
@@ -0,0 +1,62 @@
...
+ * <p>End-users would need to provide their own {@link CustomColumn} implementation.
+ * When storing objects with custom columns, the values are fetched according
+ * to the {@link #valueIn(Object) valueIn(S)} implementation. In it, the {@code S} value represents
+ * an arbitrary object serving as a source for the value.
+ *
+ * @see CustomSubjectParameter
+ * @see QueryPredicate#customParameters()
+ *
+ * @param <S> the type of objects serving as a source for the column values
+ * @param <V> the type of column values
```
dmdashenkov, line 43: Aren't `@param`s supposed to go before `@see`s?

base/src/main/java/io/spine/query/Direction.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+package io.spine.query;
+
+/**
+ * The direction of ordering for the set of records by some field value.
+ *
+ * @see OrderBy
+ */
+public enum Direction {
```
dmdashenkov, line 28: Please consider building this into the API of the `QueryBuilder`, i.e. having `orderAscendingBy(column)` instead of `orderBy(column, ASC)`. This helps the developer to type a bit less due to code completion.

base/src/main/java/io/spine/query/Either.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.util.function.Function;
+
+/**
+ * Lambda serving to fill the current predicate {@linkplain QueryBuilder query builders}
+ * with the parameters joined in {@linkplain LogicalOperator#OR disjunction}.
+ */
+@FunctionalInterface
+@Immutable
+public interface Either<B extends QueryBuilder<?, ?, ?, ?, ?>> extends Function<B, B> {
```
dmdashenkov, line 33:
```suggestion
public interface Either<B extends QueryBuilder<?, ?, ?, ?, ?>> extends UnaryOperator<B> {
```

tools/protoc-plugin/src/main/java/io/spine/tools/protoc/column/ColumnGenerator.java
```diff
@@ -45,7 +45,7 @@ private ColumnGenerator(boolean generate) {
      */
     public static ColumnGenerator instance(SpineProtocConfig config) {
```
dmdashenkov, line 46: Should this class still be called `ColumnGenerator`?

tools/protoc-plugin/src/main/java/io/spine/tools/protoc/iface/InterfaceGenerationTask.java
```diff
@@ -54,8 +62,49 @@ TypeParameters interfaceParameters() {
...
+    private static Optional<TypeParameter>
+    detectParameter(MessageType type, FirstGenericParameter annotation) {
+        Optional<TypeParameter> firstParameter;
+        try {
+            Class<? extends DetermineType> fieldTypeDetector = annotation.is();
+            Constructor<? extends DetermineType> ctor = fieldTypeDetector.getConstructor();
+            DetermineType detector = ctor.newInstance();
+            ClassName value = detector.apply(type);
+            firstParameter = Optional.of(new PredefinedTypeParameter(value));
+
```
dmdashenkov, line 102: Please remove this line.

base/src/main/java/io/spine/query/IdCriterion.java
```diff
@@ -0,0 +1,67 @@
...
+    public final B in(I... values) {
+        ImmutableSet<I> asSet = ImmutableSet.copyOf(values);
+        IdParameter<I> parameter = IdParameter.in(asSet);
+        return builder.setIdParameter(parameter);
+    }
+
+    /**
+     * Creates an instance of this criterion with the passed identifier values.
+     */
+    public final B with(Iterable<I> values) {
```
dmitrykuzmin, line 62: I think this can be safely renamed to `in(...)`.

There is hardly any possible value for `I` for which this call becomes erroneous or could be confused with `in(I...)`.

alexander-yevsyukov:
> ProjectId project_id = 1;

Can we have it just `project` to follow our recommendations on naming fields?

alexander-yevsyukov:
What happens if an entity state does not declare any column? Would we have empty `Query` and `QueryBuilder` or no classes at all?

tools/protoc-plugin/src/main/java/io/spine/tools/protoc/column/ColumnGenerator.java
armiol: @dmdashenkov yes, we still generate the columns independently from the queries, as previously. We just don't have an option to generate _only_ the columns.

alexander-yevsyukov:
> ``` java
> RecordQuery.newBuilder(ManufacturerId.class, Manufacturer.class)
>                  .where(isin)
>                  .isNot("JP 3633400001")
>                  .where(whenFounded)
>                  .isLessOrEqualTo(THURSDAY)
>                  .where(isTraded)
>                  .is(true)
>                  .build();
> ```

Assuming we're advancing generated DSL, I think we should favour readability of the usage code _over_ the fact that IDEA does not support such wrapping. So, I'd suggest always put `is()` on the same line with `where()`, here in the PR description in our core code, and in the examples. It would highlight the convenience of the language.

base/src/main/java/io/spine/query/AbstractQuery.java
```diff
@@ -0,0 +1,141 @@
...
+     */
+    private final FieldMask mask;
+
+    /**
+     * A common contract for the constructors of {@code AbstractQuery} implementations.
+     *
+     * <p>Checks that if the limit is set, at least one ordering directive is present as well.
+     */
+    AbstractQuery(AbstractQueryBuilder<I, R, P, ?, ?> builder) {
+        this.subject = new Subject<>(builder.whichIds(),
```
alexander-yevsyukov, line 82: Can we have a (possibly package-access) constructor of `Subject` by the passed `QueryBuilder`, please?

base/src/main/java/io/spine/query/ComparisonOperator.java
```diff
@@ -0,0 +1,135 @@
...
+import org.checkerframework.checker.nullness.qual.Nullable;
+
+import java.util.Objects;
+
+import static java.lang.String.format;
+
+/**
+ * Defines how the queried records are compared against the desired parameter values.
+ *
+ * <h1><a name="supported_types"><strong>Supported field types</strong></a></h1>
```
alexander-yevsyukov, line 34:
```suggestion
 * <h3><a name="supported_types"><strong>Supported field types</strong></a></h3>
```

base/src/main/java/io/spine/query/ComparisonOperator.java
```diff
@@ -0,0 +1,135 @@
...
+        }
+    },
+
+    /**
+     * The actual value must be less than the value of the subject parameter.
+     */
+    LESS_THAN {
+        @Override
+        public boolean eval(@Nullable Object left, @Nullable Object right) {
+            return GREATER_THAN.eval(right, left);
```
alexander-yevsyukov, line 65:
```suggestion
            return !EQUALS(right, left) && !GREATER_THAN.eval(right, left);
```

base/src/main/java/io/spine/query/CustomColumn.java
```diff
@@ -0,0 +1,63 @@
...
+import com.google.errorprone.annotations.Immutable;
+import io.spine.annotation.SPI;
+
+/**
+ * A column storing the data which is not directly declared as a field in the {@code Message}
+ * of an {@link io.spine.base.EntityState EntityState} or a plain record.
+ *
+ * <p>End-users may choose to store some arbitrary data along with the record.
+ * E.g. the time of entity creation or the role of the user created the record etc, that is,
+ * something not included into the definition of record {@code Message}.
```
alexander-yevsyukov, line 32:
```suggestion
 * something not included into the definition of the {@code Message} type of the record.
```

alexander-yevsyukov (changes requested):
Please see my comments so far. I'm about 1/5 of the way.

base/src/main/java/io/spine/query/CustomSubjectParameter.java
```diff
@@ -0,0 +1,50 @@
...
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.spine.query;
+
+/**
+ * A subject parameter, which defines the condition for something different than a declared field
```
alexander-yevsyukov, line 24: How about saying something like “A custom parameter for a query, which defines a condition for something different than a value of a field declared in a record...”?

Now, the abstract somewhat repeats the class name and assumes we understand what the “subject” is. This may not be the case.

If we do want to stress on the important link of this class and `Subject` (as we probably do because we `link` to it), I'd suggest repeating the link using `@see`.

base/src/main/java/io/spine/query/CustomSubjectParameter.java
```diff
@@ -0,0 +1,50 @@
...
+ * of a record or an {@link io.spine.base.EntityState EntityState}.
+ *
+ * <p>In some cases, stored records or entities include {@linkplain CustomColumn custom columns},
+ * which hold the properties not directly belonging to the record declaration. In order to include
+ * such a column into a {@link Subject} definition, a {@code CustomQueryParameter} is used.
+ *
+ * @param <S> the type of objects serving as an origin for the parameter values
```
alexander-yevsyukov, line 31: Please check wrapping for the param descriptions. They probably should go on following lines.

base/src/main/java/io/spine/query/CustomCriterion.java
```diff
@@ -0,0 +1,72 @@
...
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static io.spine.query.ComparisonOperator.EQUALS;
+
+/**
+ * Allows to specify the values for the {@link CustomSubjectParameter}s.
+ *
+ * <p>The custom parameters are set as desired values for the {@link CustomColumn}s.
+ *
+ * @param <S> the type of objects which serve as a source for the column values
```
alexander-yevsyukov, line 34: Please check the wrapping for generic params. They usually go on next lines in our code.

base/src/main/java/io/spine/query/Either.java
```diff
@@ -0,0 +1,35 @@
...
+import com.google.errorprone.annotations.Immutable;
+
+import java.util.function.UnaryOperator;
+
+/**
+ * A lambda serving to fill the current predicate {@linkplain QueryBuilder query builders}
+ * with the parameters joined in {@linkplain LogicalOperator#OR disjunction}.
+ */
+@FunctionalInterface
+@Immutable
```
alexander-yevsyukov, line 32: Please have `@Immutable` before `@FunctionalInterface`. It would read a bit easier.

base/src/main/java/io/spine/query/EntityCriterion.java
```diff
@@ -0,0 +1,61 @@
...
+ * a {@link EntityQueryBuilder} when building an {@link EntityQuery}.
+ *
+ * @param <S>
+ *         the type of entity state
+ * @param <V>
+ *         the type of the column values for this criterion
+ * @param <B>
+ *         the type of the builder in scope of which this criterion exists
+ */
+public final class EntityCriterion<S extends EntityState<?>,
```
alexander-yevsyukov, line 36: Isn't it `EntityStateCriterion`? We pass `EntityState` not an `Entity`.

base/src/main/java/io/spine/query/EntityQuery.java
```diff
@@ -0,0 +1,137 @@
...
+ */
+public abstract class EntityQuery<I,
+                                  S extends EntityState<I>,
+                                  B extends EntityQueryBuilder<I, S, B, ?>>
+        extends AbstractQuery<I, S, EntitySubjectParameter<S, ?>> {
+
+    private final B builder;
+
+    /**
+     * A common constructor contract for all {@code EntityQuery} implementations.
```
alexander-yevsyukov, line 47:
```suggestion
     * Creates a new query by the passed builder.
```

alexander-yevsyukov: The reason I'm suggesting this change is that current Javadoc simply re-states what Java assumes as the programming language.

base/src/main/java/io/spine/query/EntityQuery.java
```diff
@@ -0,0 +1,137 @@
...
+
+    /**
+     * Returns the builder instance on top of which this query has been created.
+     */
+    public final B toBuilder() {
+        return builder;
+    }
+
+    /**
+     * Creates a {@link RecordQuery} instance with the same properties as this entity query.
```
alexander-yevsyukov, line 62:
```suggestion
     * Creates a {@link RecordQuery} instance with the same attributes as this entity query.
```

base/src/main/java/io/spine/query/EntityQueryBuilder.java
```diff
@@ -0,0 +1,95 @@
...
+package io.spine.query;
+
+import io.spine.base.EntityState;
+
+import java.util.function.Function;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * An abstract base for builders of an {@code EntityQuery} for a particular entity state type.
```
alexander-yevsyukov, line 30:
```suggestion
 * The abstract base for builders of queries of entity state types.
```

alexander-yevsyukov: Please also refer to my suggestion to rename `EntityQuery` to `EntityStateQuery`. It may reduce the need of explaining what types these queries target.

alexander-yevsyukov: Or, we may keep the name, but clarify the target type matter in a different way. The problem I'm observing is that we keep explaining this thing here and there. It is cased by the fact that `Entity` wraps over `EntityState`. We should make it a bit easier in Javadocs and/or type names.

base/src/main/java/io/spine/query/package-info.java
```diff
@@ -0,0 +1,30 @@
...
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * This package defines how the objects defined in Protobuf may be queried.
```
alexander-yevsyukov, line 22: I think we should tell a bigger story saying that:
  * Storages can store records and entity states. These pieces of data can be queried.
  * Briefly tell how records are different to entity states or link to such explanation from a different package.
  * APIs for querying two cases of storing data are similar but a bit different.
  * Tell about code generation and DSL it produces.
  * Tell how this package works with the generated DSL code.
  * Give brief descriptions to the groups of classes and interfaces in this package. Presumably, they are grouped by the kind of stored data (i.e. record vs entity state).

alexander-yevsyukov (changes requested):
Please see some more (32/158)

base/src/main/java/io/spine/query/EntitySubjectParameter.java
```diff
@@ -0,0 +1,154 @@
...
+ * @param <S>
+ *         the type of entity state
+ * @param <V>
+ *         the type of the entity column values
+ */
+public final class EntitySubjectParameter<S extends EntityState<?>, V>
+        extends SubjectParameter<S, EntityColumn<S, V>, V> {
+
+    /**
+     * Creates an instance of {@code QueryParameter} targeting entities whose column value
```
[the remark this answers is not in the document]
alexander-yevsyukov: `SubjectParamter` does not seem to descend from `QueryParameter`.

base/src/main/java/io/spine/query/EntitySubjectParameter.java
```diff
@@ -0,0 +1,154 @@
...
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static io.spine.query.ComparisonOperator.EQUALS;
+import static io.spine.query.ComparisonOperator.GREATER_OR_EQUALS;
+import static io.spine.query.ComparisonOperator.GREATER_THAN;
+import static io.spine.query.ComparisonOperator.LESS_OR_EQUALS;
+import static io.spine.query.ComparisonOperator.LESS_THAN;
+
+/**
+ * A parameter defining how to query an entity by the value of its {@link EntityColumn}.
```
alexander-yevsyukov, line 33:
```suggestion
 * A parameter defining how to query an entity by the value of its {@linkplain EntityColumn column}.
```

base/src/main/java/io/spine/query/EntitySubjectParameter.java
```diff
@@ -0,0 +1,154 @@
...
+     *         the value to compare to
+     * @param <S>
+     *         the type of the state of entities which records are queried
+     * @param <V>
+     *         the type of the value
+     * @return a new instance of a parameter
+     */
+    public static <S extends EntityState<?>, V> EntitySubjectParameter<S, V>
+    lessThan(EntityColumn<S, V> column, V value) {
+        return new EntitySubjectParameter<>(column, value, LESS_THAN);
```
alexander-yevsyukov, line 95: Please add null checks similarly to `is()`.

base/src/main/java/io/spine/query/EntitySubjectParameter.java
```diff
@@ -0,0 +1,154 @@
...
+     *         the value to compare to
+     * @param <S>
+     *         the type of the state of entities which records are queried
+     * @param <V>
+     *         the type of the value
+     * @return a new instance of a parameter
+     */
+    public static <S extends EntityState<?>, V> EntitySubjectParameter<S, V>
+    lessOrEqualTo(EntityColumn<S, V> column, V value) {
+        return new EntitySubjectParameter<>(column, value, LESS_OR_EQUALS);
```
alexander-yevsyukov, line 114: Please add null checks similarly to `is()`.

base/src/main/java/io/spine/query/EntitySubjectParameter.java
```diff
@@ -0,0 +1,154 @@
...
+     *         the value to compare to
+     * @param <S>
+     *         the type of the state of entities which records are queried
+     * @param <V>
+     *         the type of the value
+     * @return a new instance of a parameter
+     */
+    public static <S extends EntityState<?>, V> EntitySubjectParameter<S, V>
+    greaterThan(EntityColumn<S, V> column, V value) {
+        return new EntitySubjectParameter<>(column, value, GREATER_THAN);
```
alexander-yevsyukov, line 133: Please add null checks similarly to `is()`.

base/src/main/java/io/spine/query/EntitySubjectParameter.java
```diff
@@ -0,0 +1,154 @@
...
+     *         the value to compare to
+     * @param <S>
+     *         the type of the state of entities which records are queried
+     * @param <V>
+     *         the type of the value
+     * @return a new instance of a parameter
+     */
+    public static <S extends EntityState<?>, V> EntitySubjectParameter<S, V>
+    greaterOrEqualTo(EntityColumn<S, V> column, V value) {
+        return new EntitySubjectParameter<>(column, value, GREATER_OR_EQUALS);
```
alexander-yevsyukov, line 152: Please add null checks similarly to `is()`.

base/src/main/java/io/spine/query/IdCriterion.java
```diff
@@ -0,0 +1,67 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.spine.query;
+
+import com.google.common.collect.ImmutableSet;
+
+/**
+ * An expression which sets the values of record identifiers to be used in {@link Query}.
```
alexander-yevsyukov, line 26:
```suggestion
 * An expression which sets the values of record identifiers to be used in a {@linkplain Query query}.
```

base/src/main/java/io/spine/query/RecordQuery.java
```diff
@@ -0,0 +1,80 @@
...
+
+package io.spine.query;
+
+import com.google.protobuf.Message;
+import io.spine.annotation.SPI;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * A query for the records each represented by a particular Protobuf message.
```
alexander-yevsyukov, line 29: Now `RecordQuery` is different to `Query` which is also is purposed (as its Javadoc says) to “fetch the records defined as Protobuf messages”?

base/src/main/java/io/spine/query/RecordQuery.java
```diff
@@ -0,0 +1,80 @@
...
+ *
+ * <p>If the Protobuf message defines the state of an entity, {@link EntityQuery} serves querying
+ * better than this type.
```
alexander-yevsyukov, line 32: > better thank this type.

Why? How? Can I still use _this_ type? Will it hurt?

base/src/main/java/io/spine/query/RecordQuery.java
```diff
@@ -0,0 +1,80 @@
...
+ * @see EntityQuery
+ */
+@SPI
+public final class RecordQuery<I, R extends Message>
+        extends AbstractQuery<I, R, RecordSubjectParameter<R, ?>> {
+
+    private final RecordQueryBuilder<I, R> builder;
+
+    /**
+     * Creates a new instance of {@code RecordQuery} on top of the passed builder.
```
alexander-yevsyukov, line 47:
```suggestion
     * Creates a new instance on top of the passed builder.
```

base/src/main/java/io/spine/query/IdParameter.java
```diff
@@ -0,0 +1,70 @@
...
+
+package io.spine.query;
+
+import com.google.common.collect.ImmutableSet;
+import com.google.errorprone.annotations.Immutable;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * Defines the values of {@link Subject} subject identifiers.
```
alexander-yevsyukov, line 29:
```suggestion
 * Defines identifiers of a query targets.
```

base/src/main/java/io/spine/query/IdParameter.java
```diff
@@ -0,0 +1,70 @@
...
+@Immutable(containerOf = "I")
+public final class IdParameter<I> {
+
+    private final ImmutableSet<I> values;
+
+    private IdParameter(ImmutableSet<I> values) {
+        this.values = values;
+    }
+
+    public ImmutableSet<I> values() {
```
alexander-yevsyukov, line 43: Please document.

base/src/main/java/io/spine/query/IdParameter.java
```diff
@@ -0,0 +1,70 @@
...
+        return values;
+    }
+
+    public static <I> IdParameter<I> empty() {
```
alexander-yevsyukov, line 47: Please document.

base/src/main/java/io/spine/query/IdParameter.java
```diff
@@ -0,0 +1,70 @@
...
+        return new IdParameter<>(ImmutableSet.of());
+    }
+
+    public static <I> IdParameter<I> is(I value) {
```
alexander-yevsyukov, line 51: Please document.

base/src/main/java/io/spine/query/IdParameter.java
```diff
@@ -0,0 +1,70 @@
...
+        checkNotNull(value);
+        return new IdParameter<>(ImmutableSet.of(value));
+    }
+
+    /**
+     * Creates an instance of {@code IdParameter} with the identifier values restricted
+     * to the passed one.
+     *
+     * @param values
+     *         the identifier values to use
```
alexander-yevsyukov, line 61: Can the passed list be empty?

alexander-yevsyukov (changes requested):
Please see my comments (37/158).

base/src/main/java/io/spine/query/QueryCriterion.java
```diff
@@ -0,0 +1,133 @@
...
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static io.spine.query.ComparisonOperator.EQUALS;
+import static io.spine.query.ComparisonOperator.GREATER_OR_EQUALS;
+import static io.spine.query.ComparisonOperator.GREATER_THAN;
+import static io.spine.query.ComparisonOperator.LESS_OR_EQUALS;
+import static io.spine.query.ComparisonOperator.LESS_THAN;
+
+/**
+ * Expression which sets some restriction to the value of a record column when querying the records.
```
alexander-yevsyukov, line 33:
```suggestion
 * A restriction to a value of a record column when querying the records.
```

base/src/main/java/io/spine/query/QueryCriterion.java
```diff
@@ -0,0 +1,133 @@
...
+abstract class QueryCriterion<R extends Message,
+                              V,
+                              C extends RecordColumn<R, V>,
+                              B extends AbstractQueryBuilder<?, ?, ?, B, ?>> {
+
+    private final C column;
+    private final B builder;
+
+    /**
+     * Creates a new instance of {@code QueryCriterion}.
```
alexander-yevsyukov, line 44:
```suggestion
     * Creates a new instance of by the passed column and the query builder.
```

base/src/main/java/io/spine/query/QueryCriterion.java
```diff
@@ -0,0 +1,133 @@
...
+     * @param builder
+     *         the builder of an query in scope of which the criterion is created
+     */
+    QueryCriterion(C column, B builder) {
+        this.column = column;
+        this.builder = builder;
+    }
+
+    /**
+     * Creates a new subject parameter of the type specific for the query type being built.
```
alexander-yevsyukov, line 57:
```suggestion
     * Adds a parameter for comparing the values of the column with the passed value using the passed operator.
```

base/src/main/java/io/spine/query/QueryCriterion.java
```diff
@@ -0,0 +1,133 @@
...
+     * @param operator
+     *         the comparison operator
+     * @param value
+     *         the value to which actual column values will be compared
+     * @return a new instance of the subject parameter
+     */
+    protected abstract B addParameter(B builder, C col, ComparisonOperator operator, V value);
+
+    /**
+     * Appends an associated query builder with a criterion checking that the value
```
alexander-yevsyukov, line 70: As discussed, let's change the wording of public `isXxx()` methods to start with the restriction they introduce and follow with the text of appending the criterion to the associated builder.

base/src/main/java/io/spine/query/QueryPredicate.java
```diff
@@ -0,0 +1,183 @@
...
+import com.google.errorprone.annotations.CanIgnoreReturnValue;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * Joins the {@linkplain SubjectParameter subject parameters} with
```
alexander-yevsyukov, line 33:
```suggestion
 * Joins {@linkplain SubjectParameter parameters} of a query with
```

alexander-yevsyukov (changes requested):
Please see my comments (42/158)

base/src/main/java/io/spine/query/ComparisonOperator.java
armiol: The operands were swapped places, therefore the original code works properly.

base/src/main/java/io/spine/query/Subject.java
```diff
@@ -0,0 +1,120 @@
...
+import com.google.protobuf.Message;
+
+import java.util.Objects;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * Set of criteria for the records obtained via querying or subscription.
+ *
+ * <p>Subjects are typically stored Protobuf messages or entity states.
```
alexander-yevsyukov, line 33: > p>Subjects are typically stored Protobuf messages or entity states.

Now it sounds as if `Subject` implements `Message`, which is not the case. Perhaps, we can speak about query targets here. Or, delete this paragraph at all?

base/src/main/java/io/spine/query/Subject.java
```diff
@@ -0,0 +1,120 @@
...
+
+    /**
+     * The criteria put on the identifiers of the records of interest.
+     */
+    private final IdParameter<I> id;
+
+    /**
+     * The type of the queried records.
+     */
+    private final Class<R> recordType;
```
alexander-yevsyukov, line 50: Consider putting this field first. It would match `SELECT <recordType>` syntax.

base/src/main/java/io/spine/query/Subject.java
```diff
@@ -0,0 +1,120 @@
...
+
+    /**
+     * The type of the identifiers of the queried records.
+     */
+    private final Class<I> idType;
```
alexander-yevsyukov, line 55: Can we have `idType` declared before `id`? This way it would be easier to understand it as filtering thing.

base/src/main/java/io/spine/reflect/Invokables.java
```diff
@@ -0,0 +1,211 @@
...
+import java.util.function.Supplier;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static io.spine.util.Exceptions.newIllegalArgumentException;
+import static io.spine.util.Exceptions.newIllegalStateException;
+import static java.lang.String.format;
+import static java.lang.invoke.MethodHandles.publicLookup;
+
+/**
+ * A utility class for working with Java {@linkplain java.lang.reflect.Method methods} and
```
alexander-yevsyukov, line 40:
```suggestion
 * Utility class for working with Java {@linkplain java.lang.reflect.Method methods} and
```

base/src/main/java/io/spine/reflect/Invokables.java
```diff
@@ -0,0 +1,211 @@
...
+     * @throws IllegalStateException
+     *         if a {@code ReflectiveOperationException} is thrown
+     *         by {@code fn}, or other error occurs during the reflective operation execution
+     */
+    @SuppressWarnings("OverlyBroadCatchBlock" /* catching any runtimes does not hurt here. */)
+    private static <T, R> R invokePreservingAccessibility(T reflectiveObject,
+                                                          Function<T, Invokable<?, ?>> makeInvokable,
+                                                          ReflectiveFunction<T, R> fn,
+                                                          Supplier<String> onError) {
+        checkNotNull(reflectiveObject);
```
alexander-yevsyukov, line 186: We can probably save on null checks, assuming this is a private method. By the type we get here outer API should have checked all the arguments.

alexander-yevsyukov (changes requested):
Please see my comments (56/172)

base/src/main/java/io/spine/query/EntityQueryBuilder.java
```diff
@@ -0,0 +1,95 @@
...
+package io.spine.query;
+
+import io.spine.base.EntityState;
+
+import java.util.function.Function;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+
+/**
+ * An abstract base for builders of queries which are aimed to fetch the records of entity states.
```
alexander-yevsyukov, line 30:
```suggestion
 * Base type for builders of queries which are aimed to fetch the records of entity states.
```

base/src/test/java/io/spine/base/given/VariableControlledEnvironment.java
```diff
@@ -0,0 +1,48 @@
...
+ * An environment {@linkplain #enabled() that is enabled} based on the value of an env variable
+ * specified to the constructor.
+ */
+@SuppressWarnings("AbstractClassWithoutAbstractMethods")
+public abstract class VariableControlledEnvironment extends EnvironmentType {
+
+    private final String envVariable;
+
+    VariableControlledEnvironment(String variable) {
+        this.envVariable = variable;
```
alexander-yevsyukov, line 35: Please call `super();`

base/src/test/java/io/spine/base/given/VariableControlledEnvironment.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    @SuppressWarnings("unused" /* invoked via reflection. */)
+    VariableControlledEnvironment() {
+        this.envVariable = "";
```
alexander-yevsyukov, line 40: Please call `super();`

base/src/test/java/io/spine/query/RecordColumnTest.java
```diff
@@ -0,0 +1,83 @@
...
+
+@DisplayName("`RecordColumn` should")
+class RecordColumnTest {
+
+    @Test
+    @DisplayName("allow creating new instances")
+    void allowCreation() {
+        String name = "description";
+        String description = "some description";
+        RecordColumn<Manufacturer, String> column = new RecordColumn<>(name,
```
alexander-yevsyukov, line 40: Please have creation coming on the new line, with an empty line before following `assertThat()`.

base/src/test/java/io/spine/query/RecordQueryBuilderTest.java
```diff
@@ -0,0 +1,351 @@
...
+            assertThat(actualIdParam.values()).isEqualTo(expectedValues);
+        }
+
+        @Test
+        @DisplayName("by the values of several columns")
+        void byColumnValues() {
+            boolean stocksAreTraded = true;
+            String isinValue = "JP 3633400001";
+            RecordQuery<ManufacturerId, Manufacturer> query =
+                    manufacturerBuilder().where(isin)
```
alexander-yevsyukov, line 124: Please have `is()` on the same line with `.where()`.

base/src/test/java/io/spine/query/RecordQueryBuilderTest.java
```diff
@@ -0,0 +1,351 @@
...
+            assertThat(ordering.get(1)).isEqualTo(new OrderBy<>(whenFounded, ASC));
+            assertThat(query.limit()).isEqualTo(tenRecords);
+        }
+
+        @Test
+        @DisplayName("which return the same `Builder` instance if asked")
+        void returnSameBuilder() {
+            RecordQueryBuilder<ManufacturerId, Manufacturer> builder =
+                    manufacturerBuilder().where(whenFounded)
+                                         .isGreaterThan(THURSDAY)
```
alexander-yevsyukov, line 240: Please have `is()` on the same line with `where()`.

base/src/test/java/io/spine/query/RecordQueryBuilderTest.java
```diff
@@ -0,0 +1,351 @@
...
+                                            .values()).isEqualTo(ids);
+        }
+
+        @Test
+        @DisplayName("of parameters")
+        void ofParameterValues() {
+            String isinValue = "JP 3496600002";
+            List<QueryPredicate<Manufacturer>> predicates =
+                    manufacturerBuilder().where(isin)
+                                         .is(isinValue)
```
alexander-yevsyukov, line 295: Please have `is()` on the same line with `where()`.

alexander-yevsyukov (changes requested):
Please see my comments (68/172).

alexander-yevsyukov (changes requested):
Please see my comments (172/172).

base/src/main/java/io/spine/query/AbstractQueryBuilder.java
```diff
@@ -161,6 +161,19 @@ public final B withMask(Field...fields) {
         ImmutableList<String> paths = Arrays.stream(fields)
                                               .map(Field::toString)
```
alexander-yevsyukov, line 162: Please fix the alignment. Consider having `Arrays.stream()` coming on the new line after `=`.

base/src/main/java/io/spine/query/EntityQueryBuilder.java
```diff
@@ -73,6 +78,24 @@ protected EntityQueryBuilder(Class<I> idType, Class<S> stateType) {
...
+     * <p>If the mask is not set, the query results contain the records as-is.
+     *
+     * <p>Any previously set mask values are overridden by this method call.
+     *
+     * @return this instance of query builder, for chaining
+     */
+    @CanIgnoreReturnValue
+    public final B withMask(SubscribableField... fields) {
+        ImmutableList<String> paths = Arrays.stream(fields)
+                                              .map(f -> f.getField()
```
alexander-yevsyukov, line 93: Please fix the alignment.

alexander-yevsyukov (changes requested):
A couple of minor code layout requests.

base/src/main/java/io/spine/query/package-info.java
```diff
@@ -19,7 +19,386 @@
...
+ * <p>While storing a {@code Message} as a BLOB simplifies writing, it makes searching and
+ * filtering of the stored records difficult. Except for IDs, no attributes of persisted messages
+ * are exposed to the search engine of a DBMS by default. Therefore, the framework allows defining
+ * additional attributes of a message to enable querying. In terms of storage language,
+ * each declared attribute is persisted as a "column". Please note that on this level of abstraction
+ * a developer does not define the structure of an underlying database; it's still a responsibility
+ * of a particular storage port.
+ *
+ * <p>This package provides an API on building the queries for the stored records, rather than
+ * configuring the storage port. Other libraries of the framework, such as `spine-server` and
```
alexander-yevsyukov, line 51: Please replace backticks with `{@code }` here and below.

base/src/main/java/io/spine/query/package-info.java
```diff
@@ -19,7 +19,386 @@
...
+ * by the values of {@code whenFounded} column ascending, then by {@code isTraded} column values
+ * descending. Finally, the query result is limited to the top ten records.
+ *
+ * <p>Please note, that {@linkplain io.spine.query.RecordQueryBuilder#limit(int) limit(..)}
+ * method may only be used in queries which order their results. Otherwise, a runtime exception
+ * is thrown upon building the query.
+ *
+ * <h4>Field masks</h4>
+ *
+ * <p>Sometimes end-users would want to fetch only the particular fields of the stored messages.
```
alexander-yevsyukov, line 192: I'm not sure about “the particular fields”. We have not spoke about these fields yet, so the definite article may not apply.

base/src/main/java/io/spine/query/package-info.java
```diff
@@ -19,7 +19,386 @@
...
+ * <pre>
+ *     ProjectView.Query query = ProjectView.newQuery()
+ *                                          .project().is(expectedId)
+ *                                          .build();
+ * </pre>
+ *
+ * <h4>Custom Columns</h4>
+ *
+ * <p>In addition to the columns declared in the Entity state, the query API allows declaring
+ * the custom columns. They are intended to handle the cases in which some computed on-the-fly data
```
alexander-yevsyukov, line 300: > the custom columns

”the” is probably not needed here.

base/src/main/java/io/spine/query/package-info.java
```diff
@@ -19,7 +19,386 @@
...
+ * </pre>
+ *
+ * <h4>Ordering and limit</h4>
+ *
+ * <p>Similar to the plain record query API, it is possible to set the same query parameters
+ * when querying the Entity state records:
+ *
+ * <pre>
+ *     ProjectView.Query query = ProjectView
+ *        .newQuery()
```
alexander-yevsyukov, line 367: Please have `.newQuery()` on the previous line.

base/src/main/java/io/spine/query/package-info.java
```diff
@@ -19,7 +19,386 @@
...
+ * and the query DSL.
+ *
+ * <p>Notice the difference with the {@link io.spine.query.RecordQueryBuilder RecordQueryBuilder}
+ * API, as the developer no longer needs to declare any types by hand.
+ *
+ * <p>Here is an example:
+ *
+ * <pre>
+ *     ProjectView.Query query = ProjectView
+ *        .newQuery()
```
alexander-yevsyukov, line 259: Please have `.newQuery()` on the previous line.

base/src/main/java/io/spine/query/package-info.java
```diff
@@ -19,7 +19,386 @@
...
+ * the declaration of the Entity column corresponding to the Message field with
+ * the {@code (column)} option.
+ *
+ * <h4>Field masks</h4>
+ *
+ * <p>It is also possible to set the mask for each resulting Protobuf message:
+ *
+ * <pre>
+ *     FieldMask mask = fieldMaskWith(status());
+ *     ProjectView.Query query = ProjectView.newQuery()
```
alexander-yevsyukov, line 386: Assuming we are going to migrate to Java 11 soon, what if we apply the `var` syntax to the code examples _already_? It would make code samples a bit more compact. Readers who are still under Java 8 would still understand them.

alexander-yevsyukov (changes requested):
Please see my comments.

base/src/main/java/io/spine/query/package-info.java
armiol: @alexander-yevsyukov I think it should be there, as we speak of a _particular_ _kind_ of cases, right after "in which".

base/src/main/java/io/spine/query/package-info.java
armiol: @alexander-yevsyukov we'll then have to do that in every case, not only in this document.
I'd prefer to phase it out until we really migrate, rather than do that just in a few cases out of dozens.

armiol:
@alexander-yevsyukov @dmdashenkov as discussed vocally, we now have `sortAscendingBy(column)` and `sortDescendingBy(column)` methods instead of `orderBy(column, direction)`. Also, `.newQuery()` is now `query()`, making it a stronger imperative.

armiol:
@alexander-yevsyukov @dmdashenkov PTAL again.

alexander-yevsyukov (approved):
LGTM!

dmdashenkov (approved):
Please see my one comment. LGTM otherwise.

base/src/main/java/io/spine/query/package-info.java
```diff
@@ -0,0 +1,408 @@
...
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * This package defines the query DSL for the stored Protobuf messages.
```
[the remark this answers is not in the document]
alexander-yevsyukov: @dmdashenkov, this is the basis for such article. We also need to have a high-level description of the package for developers who want to understand the code.

PR #534, opened 2020-05-19
Kotlin Gradle scripts. Part 2 of 2

dmdashenkov:
This PR finalizes the transition to Kotlin Gradle scripts for the `base` repository. Here we translate `smoke-tests` and `base-validating-builders` into Kotlin.

## Custom `pull` scripts

Kotlin is a strictly-typed language which does not allow unresolved references at compile time. One of the ways to add classes into a Kotlin build script compile classpath is declaring those classes in the `buildSrc` project.

The `buildSrc` project is a conventionally-named Gradle plugin, which applied to all the projects in the hierarchy implicitly. We use this feature for extracting common dependencies and configuration routines via `deps.kt`. The whole `buildSrc` project is copied from the `config` Git submodule into the root project dir, where it is found by Gradle by convention.

However, the `smoke-tests` and `base-validating-builders`, being separate builds, do not receive the `buildSrc` from our main build. Furthermore, it is impossible to configure the `buildSrc` location.

In order for the separate builds to receive `buildSrc` in their script compile classpath, we copy the `buildSrc` directory into the root project directories of those builds: `./tools/smoke-tests` and `./base-validating-builders`. To automate this, we introduce custom scripts: `./pull` and `.\pull.bat`. Those scrips do whatever their `config` counterparts do and __also__  copy the `buildSrc` dir from the root project into the included builds.

Those scripts now should be used instead of `./config/pull` and `.\config\pull.bat`.

## Fixes in the test lib

In the `plugin-testlib` we've got tools for creating temporary Gradle projects to test our plugins on. For that, some of the project's own files were copied into a temporary directory. To find those files, we need to find the root project. That is done by looking for a `version.gradle` file. However, while renaming that file to `version.gradle.kts`, we did not leave a room for using old name and so if a project using a `version.gradle` used that tool, an obscure exception would pop up.

In this PR we allow projects to use either `version.gradle` or `version.gradle.kts`.

PR #532, opened 2020-05-08
Kotlin Gradle scripts. Part 1 of 2

dmdashenkov:
In this PR we start migrating our Gradle scripts to Kotlin.

For simplicity, we split the changes into 2 PRs. In this PR we migrate the basic scripts of the repo's main build. Other builds (`smoke-tests` and `base-validating-builders`) are to be migrated in the next part.

PR #565, opened 2020-09-08
Release `1.6.0`

dmitrykuzmin:
This PR does the release of a library version `1.6.0`.

This release brings numerous API improvements, as well as fixes and infrastructure updates to the framework.

### Execution environment
1. Custom environments support is introduced [#539, #542, #545]. 

The `Environment` now exposes API to register user-defined environment types.
This can be done by extending the `EnvironmentType` class:
```java
public final class AwsLambda extends EnvironmentType {

    @Override
    public boolean enabled() {
        // ...
    }
}
```
Then, the environment can be registered to give the user ability to check if it's active at any given moment of time:
```java
Environment.instance()
           .register(new AwsLambda());

// ...

if (Environment.instance().is(AwsLambda.class)) {
    this.sender = new AwsEmailSender();
}
```
 The Spine framework provides two environments out of the box: `Production` and `Tests`. 

2. The `io.spine.testing` package is added to the list of known testing libraries. When it occurs in the class path, the environment will always be resolved to `Tests` [#544]. 
### Protobuf Model definition
1. `(min)`, `(max)` and `(range)` constraints will now be verified to be the whole numbers when applied to a whole number field (`int32`, etc.) [#560].
2. `(is)` and `(every_is)` options will now be applied simultaneously instead of `(is)` taking priority over `(every_is)` [#531].
3. New tools are added for working with regex-es in Protobuf [#520].
4. An `(is_required)` validation option is added for `oneof` group fields [#523].
5. The `spine.ui.Language` enumeration is now defined and contains language codes defined by ISO 639-1 [#522].

### Other

JSON parser will now ignore unknown fields, allowing easier data migrations between old/new Protobuf schemas [#518].

## Fixes

1. The Spine `protoc` plugin location process is simplified. This enables creation of [composite builds](https://docs.gradle.org/current/userguide/composite_builds.html) for Spine-based projects [#559]. 
2. `Time` is tweaked to provide identical behaviour on all platforms under Java 8 and Java 9+ [#554].
3. The proper import resolution is implemented for the Dart client [#524].
4. `javax.annotations-api` dependency is added to the project so the Spine `protoc` plugin can be run safely with Java 11 [#546].

## Infrastructure

1. Stricter Gradle task ordering is introduced to eliminate randomness in build execution and improve the process reliability [#562].
2. The project build scripts are migrated to Kotlin [#532, #534].

### DIFF
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -1,5 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
+  <component name="FrameworkDetectionExcludesConfiguration">
+    <file type="web" url="file://$PROJECT_DIR$" />
+  </component>
   <component name="JavaScriptSettings">
     <option name="languageLevel" value="ES6" />
   </component>
@@ -35,5 +38,5 @@
       </value>
     </option>
   </component>
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" default="false" project-jdk-name="1.8" project-jdk-type="JavaSDK" />
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" default="true" project-jdk-name="1.8" project-jdk-type="JavaSDK" />
 </project>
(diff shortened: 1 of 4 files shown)


dmitrykuzmin:
@armiol PTAL.

armiol (approved):

PR #566, opened 2020-09-11
Always specify a ClassLoader for `Resource`

dmdashenkov:
In this PR we deprecate the `Resource.file(String)` method in favour of `Resource.file(String, ClassLoader)`.

The former method allows to load resources more easily, thus the API users generally avoid the latter method. However, in many cases, the former method reports missing resources which are actually present. This is an issue of complex managed environments, such as a Gradle build. Such environments use custom `ClassLoader` implementations. Since the former method uses the `ClassLoader` of the `Resource` class itself, it is sometimes unable to locate some resource files.

The `Resource.file(String, ClassLoader)` overload solves the issue by allowing the API user to provide a `ClassLoader`. For Gradle plugins, a resource file can be found only by the `ClassLoader` which loads the plugin classes. So, for example, to load file `plugin_runner.sh`, which is a resource of the `model-compiler` plugin, one should use the `ModelCompiler.class.getClassLoader()` or a `ClassLoader` of a class which is a part of the `model-compiler` module, but not a part of a dependency JAR, such as `Resource`, which is declared in `base`.

Also, in this PR we turn on Checkstyle checks, address Checkstyle warnings, and delete some unused code.


dmdashenkov:
@armiol, PTAL.

base/src/main/java/io/spine/net/Patterns.java
```diff
@@ -29,10 +29,11 @@
  * <a href="https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/util/Patterns.java">
  *  android.util.Patterns.java</a>
  */
-@SuppressWarnings(
-        {"DuplicateStringLiteralInspection" /* OK as they are parts of patterns. */,
-         "SingleCharacterStringConcatenation" /* ... will be concatenated during compilation*/,
-         "unused" /* For future use of other value object classes in this package. */})
+@SuppressWarnings({
+        "MultipleStringLiterals" /* OK as they are parts of patterns. */,
+        "SingleCharacterStringConcatenation" /* ... will be concatenated during compilation*/,
```
armiol, line 34: Please add a period and a space at the end.

base/src/main/java/io/spine/io/Resource.java
```diff
@@ -51,27 +49,30 @@
...
-        this.customLoader = customLoader;
+        this.classLoader = classLoader;
     }
 
     /**
      * Creates a new resource reference.
      *
      * @param path
      *         the path to the resource file, relative to the classpath
+     * @deprecated Using the {@code ClassLoader} of this class may lead to unexpected behaviour.
```
armiol, line 69: This statement needs more details. Or a link to some place with more details.

armiol (commented):
@dmdashenkov LGTM in general. Please see my comments.

dmdashenkov:
@armiol, PTAL again.

armiol (approved):

PR #567, opened 2020-09-21
[2.x-jdk8] Back-merge latest changes from `master`

armiol:
This changeset matches the `2.x-jdk8-master` branch even with the most recent changes made to `master` branches in `config` and `base` repos.

As long as Checkstyle was re-enabled, a few classes were additionally documented to avoid the missing description of the generic parameters. Thus, the Checkstyle checks now pass.

The own version of the framework in this branch is set to `2.0.0-jdk8-SNAPSHOT.3`.


dmdashenkov (approved):
@armiol, LGTM.

base/src/main/java/io/spine/query/package-info.java
```diff
@@ -304,7 +304,7 @@
  *
  * <pre>
  * // The column presumably defined as a custom column for querying.
- * class ArchivedColumn extends CustomColumn<EntityWithLifecycle, Boolean> {
+ * class ArchivedColumn extends{@literal CustomColumn<EntityWithLifecycle, Boolean>} {
```
alexander-yevsyukov, line 307: @armiol a space after “extends” would be nice.

base/src/main/java/io/spine/string/FnStringifier.java
```diff
@@ -26,6 +26,9 @@
 
 /**
  * Abstract base for stringifiers that convert values using function objects.
+ *
+ * @param <T>
+ *         type of stringified objects
```
alexander-yevsyukov, line 31: @armiol, don't we need “the” before ”type” here and in other generic type descriptions in this PR?

PR #572, opened 2020-10-09
Patch imports in `types.dart`

dmdashenkov:
In this PR we allow the Gradle plugin to patch imports in any Dart file, not just in `.pb.dart` ones.

In order to collect info about known types, we generate `types.dart` file in every package. Previously, our Gradle plugin did not resolve relative path imports in those files. Now it does. Thanks to that, we're able to use Dart Protobuf classes from `spine_client` instead of generating them in place each time.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,128 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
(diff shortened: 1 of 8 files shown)


dmdashenkov:
@armiol, PTAL.

dmdashenkov:
Codecov collapse is a mystery.

armiol (approved):

PR #573, opened 2020-10-13
Allow short relative paths in imports

dmdashenkov:
This PR fixes a crash in `FileReference.directory()`. Currently, the method throws an exception if the file reference has no directory, i.e. it's a simple file name (possibly, with the "`./`" prefix). This PR fixes that crash.

### DIFF
--- a/tools/tool-base/src/main/java/io/spine/code/fs/DirectoryReference.java
+++ b/tools/tool-base/src/main/java/io/spine/code/fs/DirectoryReference.java
@@ -24,8 +24,6 @@ import com.google.common.base.Splitter;
 import com.google.common.collect.ImmutableList;
 import io.spine.value.StringTypeValue;
 
-import java.util.List;
-
 import static io.spine.util.Preconditions2.checkNotEmptyOrBlank;
 
 /**
@@ -37,6 +35,8 @@ import static io.spine.util.Preconditions2.checkNotEmptyOrBlank;
 public final class DirectoryReference extends StringTypeValue {
 
     private static final long serialVersionUID = 0L;
+
+    private static final DirectoryReference CURRENT = new DirectoryReference("");
 
     private DirectoryReference(String value) {
         super(value);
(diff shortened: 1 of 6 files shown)


dmdashenkov:
@armiol, PTAL.

tools/tool-base/src/test/java/io/spine/code/fs/js/FileReferenceTest.java
```diff
@@ -63,4 +64,13 @@ void directorySkipRelative() {
         assertEquals("foo/bar", fileReference.directory()
                                              .value());
     }
+
+    @Test
+    @DisplayName("obtain the empty directory path")
+    void emptyDirectory() {
+        FileReference file = FileReference.of("./neighbour.txt");
```
armiol, line 71: Honestly, I would expect that the `/` symbol here would be a system-specific path delimiter. I don't know, but it seems to me that some Windows versions would gladly interpret `.\neighbour.txt` as a file in the current folder.

tools/tool-base/src/main/java/io/spine/code/fs/DirectoryReference.java
```diff
@@ -54,10 +54,17 @@ public static DirectoryReference of(String value) {
         return new DirectoryReference(value);
     }
 
+    /**
+     * Obtains the reference to the current directory.
+     */
+    public static DirectoryReference currentDir() {
```
armiol, line 60: Similar to my comment below, I would expect that the reference to the current dir would be either `"."` or `"." + File.separatorChar`.

armiol (commented):
@dmdashenkov please see my comments.

tools/tool-base/src/test/java/io/spine/code/fs/js/FileReferenceTest.java
dmdashenkov: As this is an *imported* file reference, we only use this `/` type of slashes.

armiol (approved):

PR #575, opened 2020-10-15
Structure the version file.

alexander-yevsyukov:


alexander-yevsyukov:
According to `master`, this change was already applied somewhere else.

PR #576, opened 2020-10-15
Repackage the Guava `Beta` dependencies of `base:main` source set

alexander-yevsyukov:
...as well as their transitive `Beta` dependencies.


alexander-yevsyukov:
@armiol, what do we do with this?  It's being hanging around for quite a while.

I would say it would be better to manipulate `.class` files changing imports instead of doing that manually.  We can, obviously, continue the manual path and watch for changes in the Beta status of Guava API. It's not too much work, but it's still additional manual work.

alexander-yevsyukov:
We always use the latest one, and enforce it across the dependencies. So, for the framework development and current users it's not an issue. As confirming vocally with @armiol, let's put this off until we have a real issue with conflicts caused by Guava version.

PR #579, opened 2020-10-16
Extract `spine-tools-api` artifact

### LINKED ISSUE #577: Extract `io.spine.tools` package from the `spine-base` artifact
The aim of the task is to extract the `tools`-only APIs from the `spine-base` which is part of the application runtime.

It is the desired change to split compile-time-only and runtime dependencies. Compile-time-only dependencies should not leak into the runtime.

yuri-sergiichuk:
This PR closes #577.

In the PR I have extracted the content of the `io.spine.tools.protoc` package into a separate published `spine-tools-api` artifact.

## Potentially breaking *dependency* change

The new artifact `io.spine.tools:spine-tools-api` should be added as a dependency by the developers who are extending the code generation functionality using their own `MethodFactory` or `NestedClassFactory` implementations.

From the other perspectives, the codebase is fully compatible with the previous approach and no further actions are required.


yuri-sergiichuk:
@alexander-yevsyukov PTAL.

alexander-yevsyukov (approved):
LGTM

PR #580, opened 2020-10-19
Update config in v2-jdk8

alexander-yevsyukov:
This PR:
 1.  updates `config` files
 2. fixes the warning generated when building Javadocs because of `#` symbol in the `@see` Javadoc link.


alexander-yevsyukov:
@armiol, after some further thought, I don't that removing the hash part of the link is a good idea. Changing ErrorProne to the latest version (2.4.0) did not remove the issue. Let's discuss the best course of action here. I think we should minimise warnings in the console as much as possible.

armiol (approved):

armiol:
Closing in favor of #582.

PR #581, opened 2020-10-20
Ensure that `base/master` always looks at `config/master` branch

armiol:
This changeset updates the `pull` script in the root of this repository so that it switches the `config` to its `master` branch.

As there is another `2.x-jdk8-master` branch in this repo, this change is a precautionary measure for those working with both framework versions.


alexander-yevsyukov (approved):
LGTM

Issue #186, opened 2018-08-30
Add `SingletonClassTest`


alexander-yevsyukov:
Similarly to existing `UtilityClassTest`, we need a test utility that would allow testing a singleton class.

## Construction

The constructor should accept a method reference for obtaining the instance of the singleton.
Later this supplier would be used in test methods.

## Serialization
If the class implements `Serializable`, then the test could look like this:

```java
    @Test
    @DisplayName("be serializable")
    void serialize() {
        SerializableTester.reserializeAndAssert(singletonSupplier.get());
    }
```   

But what if it is not? Probably, we can have `SerializableSingletonClassTest` that is descendant of `SingletonClassTest`.

## Returning same instance
This is easy:

```java
    @Test
    @DisplayName("return the same instance")
    void sameInstance() {
        assertSame(singletonSupplier.get(), singletonSupplier.get());
    }
```

alexander-yevsyukov:
[Suggested](https://github.com/SpineEventEngine/core-java/pull/832#discussion_r214091492) by @dmdashenkov.

alexander-yevsyukov:
Once implemented, please use this arrangement for `EmptyClassTest` under `core-java`.

PR #583, opened 2020-10-23
Class tests

alexander-yevsyukov:
This PR improves utilities for testing classes:
  * Common abstract base `ClassTest` was created.
  * `UtilityClassTest` is derived from `ClassTest`.
  * `UtilityClassTest.getUtilityClass()` deprecated in favor of `ClassTest.subject()`.
  * `SingletonClassTest` was introduced (#186).

### DIFF
--- a/testlib/src/main/java/io/spine/testing/ClassTest.java
+++ b/testlib/src/main/java/io/spine/testing/ClassTest.java
@@ -1,0 +1,132 @@
+/*
+ * Copyright 2020, TeamDev. All rights reserved.
+ *
+ * Redistribution and use in source and/or binary forms, with or without
+ * modification, must retain the above copyright notice and the following
+ * disclaimer.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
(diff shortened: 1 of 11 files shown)


testlib/src/main/java/io/spine/testing/ClassTest.java
```diff
@@ -0,0 +1,132 @@
...
+/**
+ * Abstract base for test suites that test a class (e.g. static methods) rather than an object.
+ *
+ * @param <C>
+ *         the class under the tests
+ */
+@SuppressWarnings("UnstableApiUsage") // NPE testing utils from Guava.
+public abstract class ClassTest<C> {
+
+    private final Class<C> subject;
```
dmdashenkov, line 43: The `Class` itself is not a subject per se, rather the instance and/or static methods are.

dmdashenkov: I suggest renaming.

dmdashenkov (approved):
LGTM with a single comment to address.

testlib/src/main/java/io/spine/testing/ClassTest.java
alexander-yevsyukov: When we test a utility class, it's the class that's under consideration. This is what in the language we speak. When it comes to singleton, it's an class which produces only one object. Also, the name was given in line with what we have in Google Truth.

alexander-yevsyukov: @armiol, I've requested your review hoping you may suggest a better name for class under tests.

alexander-yevsyukov:
It looks like we're failing because of [this issue](https://github.com/gradle/gradle/issues/14242).

testlib/src/main/java/io/spine/testing/ClassTest.java
armiol: @alexander-yevsyukov @dmdashenkov I was ready to suggest something like `StaticApiTest`. While it's a bad name in general, it defines the purpose of derived tests more or less precisely. 
But now, as I see, we also test the class-under-testing for being `final` and for having private and parameter-less ctor, I don't think it's a good name either.

`ClassTest` feels a bit generic to me. By its name, it should be applicable to testing any class, while it's not quite.

So let me think a couple of dozen minutes more.

armiol: @alexander-yevsyukov @dmdashenkov as to `subject`, this variable name looks fine to me. It's tightly related to the name of the `ClassTest`. So, if we keep `ClassTest`, we should keep `subject` for sure, because that's what we test.

armiol: @alexander-yevsyukov @dmdashenkov to satisfy each and every, could we use `private final Class<C> input;`?

armiol (approved):
@alexander-yevsyukov @dmdashenkov I have browsed through the source code of Truth and through various vocabularies and wikipedias in an attempt to find something better.
However, in CS there is a handful of words defining something-under-test. In particular,

* `instance` — but we do NOT test an instance, so that's a no-go,
* `sample` — which correlates with `instance`, so again, not really a match,
* `subject` — which to me sounds OK, and is also  incorporated by Truth,

...and more generic terms, such as
* `input`  — which we could use,
* `actual` — I don't think that's a match, as there is no `expected`.

As to the name of `ClassTest`, I am not able to come up with anything better. But given it's a base type for `SingletonTest` and `UtilityClassTest`, and thus it isn't going to be used directly very often, it looks good to me.

PR #587, opened 2020-11-10
Allow usage of `(distinct)` option on enums

### LINKED ISSUE #437: Add message casters for Protobuf descriptors
The `(distinct)` option applied on enum values fails to actually convert an enum value to `Any` because of the constraint iterates over `EnumValueDescriptor` instead of actual enum values.

In order to support consistent behavior across all possible types, we should consider adding appropriate message casters for such a situation.

yuri-sergiichuk:
This PR resolves #437.

Handling of enum values is considered a unique case while it's the only type that returns `EnumValueDescriptor` over the actual Protobuf message or value.
To be able to fill up the constraint violation from the enum, we first grab its Proto representation (`EnumValueDescriptorProto`) and then use it as the violating value.

The other change done in the PR is the improvement of the types conversion functionality — I've extracted inner classes of the `TypeConverter` utility into the package to make it easier to read and manage.


alexander-yevsyukov:
@yuri-sergiichuk please tell about refactorings done in the description of the PR. It would have been much easier to understand what's going on if such a description were included.

alexander-yevsyukov (changes requested):
Please see my comments. I understand that existing code was pushed to the package level. And my comments related to the state that it used to be 3 years or so. But, as we're making this code more visible, let's make it better.

Also, I could not find a test that checks that declaring `(distrinct)` on enum values is possible and indeed causes validation issues when violated. Please point me.

base/src/main/java/io/spine/validate/MessageValidator.java
```diff
@@ -201,8 +207,11 @@ public void visitRequiredOneof(IsRequiredConstraint constraint) {
             ConstraintViolation violation = ConstraintViolation
                     .newBuilder()
                     .setMsgFormat(constraint.errorMessage(message.context()))
-                    .setFieldPath(Field.named(constraint.oneofName().value()).path())
-                    .setTypeName(targetType.name().value())
+                    .setFieldPath(Field.named(constraint.oneofName()
+                                                        .value())
+                                       .path())
```
dmdashenkov, line 212: Please realign this fragment. You could either extract a variable to preserve the consistency with automatic formatting or simply put `constraint.oneofName().value()` into a single line.

I assume there's no way to match the type of the param to the type of the return value.

version.gradle.kts
```diff
@@ -34,7 +34,7 @@
 /**
  * The version of this library.
  */
-val base = "1.6.7"
+val base = "1.6.8"
```
dmdashenkov, line 37: Shouldn't the library version be updated? And shouldn't our CI catch that it's not updated? Please investigate why the build is "green" while the published version has not changed.

dmdashenkov (changes requested):

version.gradle.kts
yuri-sergiichuk: The PR was created and before the `1.6.8` version was merged in the master. So I guess everything is fine here. I'll bump the version up.

image

base/src/main/java/io/spine/protobuf/EnumCaster.java
```diff
@@ -0,0 +1,91 @@
...
+            int number = input.getNumber();
+            return convertByNumber(number);
+        } else {
+            return convertByName(name);
+        }
+    }
+
+    private Enum<? extends ProtocolMessageEnum> convertByNumber(int number) {
+        Enum<? extends ProtocolMessageEnum>[] constants = type.getEnumConstants();
+        for (Enum<? extends ProtocolMessageEnum> constant : constants) {
```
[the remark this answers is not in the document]
yuri-sergiichuk: No, `for` loop is just cleaner in this particular place. Having type casts to `ProtocolMessageEnum` and back to `Enum<? extends ProtocolMessageEnum` within a stream is a pain and is just harder to read.

yuri-sergiichuk:
> Also, I could not find a test that checks that declaring `(distrinct)` on enum values is possible and indeed causes validation issues when violated. Please point me.
@alexander-yevsyukov It's the `DistinctTest`. It has nested classes for situations when the constraint is violated and is not violated. So the test case was there before this PR with tests disabled for enums.

yuri-sergiichuk:
@alexander-yevsyukov @dmdashenkov PTAL again.

base/src/main/java/io/spine/protobuf/ProtoConverter.java
```diff
@@ -21,40 +21,53 @@
...
 /**
- * The {@link Function} performing the described type conversion.
+ * Performs two-way conversion of a {@linkplain Message Protobuf Message} to its
+ * Java counterpart and back.
+ *
+ * <p>The inheritors implement the actual conversion to a specific Message and Object.
  *
  * @param <M>
- *         type of message to convert
+ *         Protobuf Message to convert
```
alexander-yevsyukov, line 37: It's not the message, but its type. How about “the type of a message to convert"?

base/src/main/java/io/spine/validate/MessageValidator.java
```diff
@@ -149,22 +156,22 @@ public void visitDistinct(DistinctConstraint constraint) {
...
+        FieldValue value = message.valueOf(field);
         Optional<FieldDeclaration> declaration = withField(message, constraint);
-        checkState(declaration.isPresent(),
-                   "Field `%s` noted in `(goes).with` option is not found.",
-                   constraint.optionValue()
-                             .getWith());
+        String withFieldName = constraint.optionValue().getWith();
+        checkState(
+                declaration.isPresent(),
+                "Field `%s` noted in `(goes).with` option is not found.",
```
alexander-yevsyukov, line 165: Please replace “noted” with specified. And please start with “The” before “field”.

base/src/main/java/io/spine/validate/MessageValidator.java
```diff
@@ -207,11 +214,8 @@ public void visitRequiredOneof(IsRequiredConstraint constraint) {
             ConstraintViolation violation = ConstraintViolation
                     .newBuilder()
                     .setMsgFormat(constraint.errorMessage(message.context()))
-                    .setFieldPath(Field.named(constraint.oneofName()
-                                                        .value())
-                                       .path())
-                    .setTypeName(targetType.name()
-                                           .value())
+                    .setFieldPath(Field.named(constraint.oneofName().value()).path())
```
alexander-yevsyukov: Please extract field path creation into a variable. It's too complicated now.

base/src/main/java/io/spine/validate/MessageValidator.java
```diff
@@ -254,10 +257,8 @@ public void visitCustom(CustomConstraint constraint) {
...
     private static void checkTypeConsistency(Range<ComparableNumber> range, FieldValue value) {
         if (range.hasLowerBound() && range.hasUpperBound()) {
-            NumberText upper = range.upperEndpoint()
-                                    .toText();
-            NumberText lower = range.lowerEndpoint()
-                                    .toText();
+            NumberText upper = range.upperEndpoint().toText();
+            NumberText lower = range.lowerEndpoint().toText();
             if (!upper.isOfSameType(lower)) {
                 String errorMessage = "Boundaries have inconsistent types: lower %s, upper %s";
```
alexander-yevsyukov, line 263: Please backtick type names and finish the message with period. Also, please have “is” after “lower” and “upper”.

alexander-yevsyukov (commented):
LGTM, with few minor requests on diag. messages. Please address before merging.

yuri-sergiichuk:
@alexander-yevsyukov PTAL again.

alexander-yevsyukov (approved):
LGTM

yuri-sergiichuk:
@dmdashenkov PTAL.

version.gradle.kts
dmdashenkov: Sorry, my bad. I thought I was looking at the `core-java`, for some reason.

dmdashenkov (approved):

PR #589, opened 2020-11-13
`PubCache` API

dmdashenkov:
In this PR we create a `PubCache` utility which obtains the path to the local Pub cache. `PubCache` is used by the Proto Dart plugin and downstream by the Bootstrap plugin.

### DIFF
--- a/tools/proto-dart-plugin/src/main/java/io/spine/generate/dart/PubCache.java
+++ b/tools/proto-dart-plugin/src/main/java/io/spine/generate/dart/PubCache.java
@@ -1,0 +1,97 @@
+/*
+ * Copyright 2020, TeamDev. All rights reserved.
+ *
+ * Redistribution and use in source and/or binary forms, with or without
+ * modification, must retain the above copyright notice and the following
+ * disclaimer.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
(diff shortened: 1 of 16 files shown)


dmdashenkov:
@armiol, PTAL.

tools/proto-dart-plugin/src/main/java/io/spine/generate/dart/PubCache.java
```diff
@@ -0,0 +1,97 @@
...
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package io.spine.generate.dart;
```
armiol, line 21: I would consider moving this class to something like `io.spine.dart`, as it seems more generic than just a utility for the codegen.

armiol (approved):
@dmdashenkov LGTM with a suggestion.

PR #594, opened 2020-11-25
Improve environment

alexander-yevsyukov:
This PR improves the API of working with the `Environment`, making it consistent with how `ServerEnvironment` works in `core-java`.

Notable changes:
  * `CustomEnvironmentType` was introduced primarily for helping understand the concept of a custom type, and avoid ambiguity when speaking about “mutual exclusivity”. Previously, all the types were descending from `EnvironmentType` and two of them being `Tests` and `Production` were always present, and, thus, one of them could not be mutually exclusive with a custom one. Yes, they were standard ones, but it still raised the question, and because of that...
  * `StandardEnvironmentType` was introduced. It's package-private with the primary purpose of groping `Tests` and `Production` types together.
 * Registering a type now accepts `Class<? extends CustomEnvironmentType>`, not an instance of a type. This plays nicer with `ServerEnvironment.when(Class)`. As we mimic `enum`s let's be operating with classes, not instances.
 * Deprecated methods in `Environment` were removed.
 * The publicly exposed constant with a list of unit testing packages were hidden because of a security warning. The list is now returned by a method, with its Javadoc enumerating the returned values.


armiol (approved):

armiol:
@alexander-yevsyukov LGMT as soon as the builds pass.

Also, I remembered why we had `.register(instance)`, not `.register(type)`. That's because we wanted the developer to be able to feed some dependencies (such as clients to Google Cloud or AWS APIs) into the environment types.

However, now I think that was an over-complication as there was no such cases at the moment. Still, we will probably face such a need some time in future. Once we do, we'll design a solution for it.

base/src/main/java/io/spine/base/CustomEnvironmentType.java
```diff
@@ -0,0 +1,76 @@
...
+ * In this case the environment detection functionality iterates over all known types, starting
+ * with those registered by the framework user:
+ *
+ * <pre>
+ *
+ * public final class Application {
+ *
+ *     static {
+ *         Environment.instance()
+ *                    .register(new Staging())
```
dmitrykuzmin, line 36: I believe these examples now should be `.register(Staging.class)` and so on.

alexander-yevsyukov: Well spotted!

PR #603, opened 2020-12-24
Sort out testing utilities

alexander-yevsyukov:
This PR:
   * Consolidates assertions under the `Assertions` class, moving the seasoned ones from the `Tests` utility class.
   * Renames the `Tests` class into `Testing` to avoid the name clash with the environment type.
   * Moves the `nullRef()` method into `TestValues`.


testlib/src/main/java/io/spine/testing/Assertions.java
```diff
@@ -72,6 +90,163 @@ public static UnknownTypeException assertUnknownType(Executable e) {
...
+     * using Reflection.
+     *
+     * <p>Typically this method is used to add a constructor of a utility class into
+     * the covered code.
+     *
+     * <p>Example:
+     * <pre>
+     * public class MyUtilityShould
+     *     ...
+     *     {@literal @}Test
```
armiol, line 122: As long as `{@literal @}` adds an extra space when rendered, could you please move this line one space back?

It should be like this:
```java
      {@literal @}Test
       public void haveUtilityCtor() {
```

testlib/src/test/java/io/spine/testing/TestValuesTest.java
```diff
@@ -38,15 +39,21 @@
         super(TestValues.class);
     }
 
+    @Test
+    @DisplayName("provide null reference method")
```
armiol, line 43: Let's quote `null`:
```java
@DisplayName("provide `null` reference method")
```

armiol (approved):
@alexander-yevsyukov LGTM with a few minor comments to address.

PR #605, opened 2020-12-28
Fix Pub cache on Windows

dmdashenkov:
This PR allows our Dart-related tools to work with custom Pub cache location on Windows.

Previously, the only place we were looking for the cache by default was the `%LOCALAPPDATA%` dir. Some versions of Pub use `%APPDATA%` dir instead. Now we support both without extra configuration.


dmdashenkov:
@armiol, PTAL.

armiol (approved):

PR #610, opened 2021-01-06
Update copyright to 2021 (part 2)`

dmdashenkov:
In a previous PR, we've updated the bulk of files to have "2021" in their copyright notice. However, due to a technical error, some files have escaped. In this PR we update the remainder of the files.

Also, two files, `testlib/build.gradle.kts` and `tools/smoke-tests/gradlew` had invalid notices. The former listed the copyright notice twice and the latter used to have TeamDev notice instead of the original Gradle one.


dmdashenkov:
@yuri-sergiichuk, @armiol, PTAL.

armiol (approved):

yuri-sergiichuk (approved):

PR #617, opened 2021-03-16
Phase out `jcenter()`

dmdashenkov:
We're phasing out the usage of the JCenter Maven repository, replacing it with the conventional Maven Central. JCenter is announced to be shut down soon and is already experiencing deteriorated performance.


dmdashenkov:
@armiol, PTAL.

armiol (approved):

PR #620, opened 2021-03-23
Remove `BuiltMessage`

alexander-yevsyukov:
This PR removes `BuiltMessage` interface and associated code-generation changes recently introduced.

Other changes:
  *  `SignalMessage` interface was introduced. Signal message interfaces now extend it
  * `EventMessage` and `CommandMessage` were annotated with `@GeneratedMixin` (to be in-line with `EntityState`).
  * `ValidatingBuilder` moved back to `io.spine.validate` package.
  * Published modules now expose popular dependencies using `api` level, to reduce configuration efforts when used.


build.gradle.kts
```diff
@@ -45,36 +46,46 @@ buildscript {
 @Suppress("RemoveRedundantQualifierName") // Cannot use imports here.
 plugins {
     `java-library`
+    kotlin("jvm") version "1.4.30"
     idea
-    id("com.google.protobuf") version io.spine.gradle.internal.Deps.build.protobuf.gradlePluginVersion
-    id("net.ltgt.errorprone") version io.spine.gradle.internal.Deps.build.errorProne.gradlePluginVersion
+    @Suppress("RemoveRedundantQualifierName") // Cannot use imports here.
+    io.spine.gradle.internal.Deps.build.apply {
```
dmdashenkov, line 52: I'm surprised this works. `plugins {}` DSL in very limited. This is possibly because `apply` is an `inline` method. Good to know!

dmdashenkov (approved):
LGTM.

PR #621, opened 2021-03-23
Move `testlib` under `io.spine.tools` Maven group

alexander-yevsyukov:
The `spine-testlib` artifact used to be under `io.spine` group which was confusing because the group serves for the production artifacts of the framework.

This PR moves `testlib` under `io.spine.tools` group.


yuri-sergiichuk (approved):
@alexander-yevsyukov I'd personally prefer having a separate `io.spine.test` group instead of abusing the `tools`.

alexander-yevsyukov:
> ... abusing the `tools`

@yuri-sergiichuk, there's a fine line between these words when it comes to testing and/or tooling. What if we want to add benchmarking to the testing libraries, and some bits of CLI or UI? Would it be still just `tests`? What if we want to add some IDE integration bits?

yuri-sergiichuk:
Yeah, that's a matter of what we're willing to have there.
I'd say that kinda can have all non-production artifacts in `tools`, but it worries me a bit. 

I'd say it may be better to have separate groups like: `test`, `benchmark`, `ide`, etc.

PR #622, opened 2021-03-25
Launch GitHub Actions build on every push

dmdashenkov:
In this PR we set up the GitHub `build` action to run on any push, not just PRs to `master`.

This will allow the action to run when we merge the PR to `master`, which is required so that the artifacts are now published. Since an earlier PR, we're publishing artifacts from GitHub Actions, not from Travis.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,9 +1,6 @@
 name: Build
 
-on:
-  pull_request:
-    branches:
-      - master
+on: push
 
 jobs:
   build:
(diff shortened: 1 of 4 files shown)


yuri-sergiichuk (approved):
@dmdashenkov LGMT. Why the version is `unbumped`?

dmdashenkov:
@yuri-sergiichuk, as I discovered that the previous PR didn't publish any artifacts, I reverted the version increment, as it's unnecessary.

PR #623, opened 2021-03-25
Extra credentials

dmdashenkov:
A previous attempt at publishing artifacts on GitHub Actions failed due to lack of extra credentials, namely an SSH key for publishing generated doc to spine.io.

This PR adds the required extra credential files.


yuri-sergiichuk:
@dmdashenkov can we use GitHub [secrets](https://docs.github.com/en/actions/reference/encrypted-secrets) instead of doing our own PGP encryption/decryption?

dmdashenkov:
@yuri-sergiichuk, we're going exactly that. The credentials are encrypted via `gpg`. The passphrase is stored in an org-wide secret. This is done:
 1. To make sure that even if the credentials file is too big, we still can use it.
 2. To make all credentials work in the same way, for clarity.

yuri-sergiichuk (approved):
I don't like the approach of "unifying" here. I bet in 90% of cases we'll manage to work with secrets that fit, but if that's smth that was discussed earlier, than OK.

E.g. the ssh key itself is ~2kb. And the limit is 64kb per secret.

PR #624, opened 2021-03-26
Publishing troubles

dmdashenkov:
In this PR we fix a number of issues related to moving Maven publishing to the GitHub Actions.

The version is bumped to `2.0.0-SNAPSHOT.10`, skipping a few versions, which have been published already with a different rate of success during the build process debug.


.github/workflows/build.yml
```diff
@@ -45,16 +45,21 @@ jobs:
...
+
       - name: Decrypt GCS credentials
-        run: ./config/scripts/decrypt.sh "GCS_CREDENTIALS_KEY" ./gcs-auth-key.json.gpg ./gcs-auth-key.json
+        run: ./config/scripts/decrypt.sh "$GCS_CREDENTIALS_KEY" ./gcs-auth-key.json.gpg ./gcs-auth-key.json
         env:
           GCS_CREDENTIALS_KEY: ${{ secrets.GCS_CREDENTIALS_KEY }}
 
       - name: Publish artifacts to Maven
-        run: ./gradlew assemble publish --stacktrace
+        run: chmod +x ./config/scripts/register-ssh-key.sh & ./gradlew assemble publish --stacktrace
```
yuri-sergiichuk, line 61: `&` makes the commands run simultaneously. `&&` executes the first one and then the second one. Even though `chmod` will always run faster, logically, I believe the intention was to show that the `gradle publish` requires an executable flag on the first script.
```suggestion
        run: chmod +x ./config/scripts/register-ssh-key.sh && ./gradlew assemble publish --stacktrace
```

.github/workflows/build.yml
```diff
@@ -45,16 +45,21 @@ jobs:
           CLOUDREPO_CREDENTIALS_KEY: ${{ secrets.CLOUDREPO_CREDENTIALS_KEY }}
 
       - name: Decrypt Git SSH credentials
-        run: ./config/scripts/decrypt.sh "GIT_CREDENTIALS_KEY" ./deploy_key_rsa.gpg ./deploy_key_rsa
+        run: ./config/scripts/decrypt.sh "$GIT_CREDENTIALS_KEY" ./deploy_key_rsa.gpg ./deploy_key_rsa
         env:
           GIT_CREDENTIALS_KEY: ${{ secrets.GIT_CREDENTIALS_KEY }}
 
+        # Make sure the SSH key is not "too visible". SSH agent will not accept it otherwise.
+      - run: chmod 400 ./deploy_key_rsa
```
yuri-sergiichuk, line 53: Maybe we can just have a `name` here as well?

yuri-sergiichuk (approved):
@dmdashenkov overall LGTM. Please consider two comments below.

PR #632, opened 2021-04-21
Improve naming and structure in code generation

alexander-yevsyukov:
This PR, its changes and notes serve for other PRs related to improving code generation and presumably other parts in `base`.  It is _not_ going to be merged in.

### Set 3
- [x] Rename the properties of the Gradle `Extension` are renamed to avoid the confusing `target` infix.
- [x] Improve the names in the model check extension. I.e. `spineCheckServerity` should be `defaultSeverity`. `useValidatingBuilder` should be `useValidatingBuilderSeverity`. The outer name should be `modelChecks` (instead of `spineErrorProneChecks`.
- [x] Mark code-related types as `@Immutable` wherever possible.

### Step 4
- [ ] Convert `Extension` classes to use Gradle Properties API.
   This PR converted most of the static methods of the `Extension` class to instance ones, but the properties API looks asa  better way.
- [ ] Unify `io.spine.js.gradle.Extension` with one for Java, and presumably for Dart.

### Set 5
 - [ ] Move `io.spine.tools.java.code.testing.annotation` from `mc-java/test/java` to `plugin-testlib/main/java`. Due to some strange Gradle problems it does not work at the time of writing. More investigations are needed.
 - [ ] Clarify the matter of `spine-protoc.gradle` resource file (see `tools/mc-java/src/main/resources/`).
        What do we do about:
           * Repositories, 
           * The version of the compiler, 
           * Versions of the dependencies
           in this build script stored in the resources?
 - [ ] Обревизовать `plugin-base` на предмет кода, который напрямую определяет Spine Model Compiler. Например, `ModelCompilerTaskName`.


alexander-yevsyukov:
Strangely `./gradlew clean buildAll` run locally does not run tests for `proto-dart-plugin` and `proto-js-plugin`. I'm looking into it.

PR #634, opened 2021-05-03
New package names for `mc-dart`

alexander-yevsyukov:
This PR improves the name of the types implementing Model Compiler for Dart.

* The Dart Gradle plugin now resides under `io.spine.tools.mc.dart.gradle` package.
* `DefaultDartProject` moved under `io.spine.tools.dart.fs` package (from `io.spine.code.fs.dart`).
* `DirectoryPattern` and `ExternalModule` moved from `io.spine.code.structure` package to `io.spine.tools` (thus becoming parts of `tool-base` artifact`, instead of `plugin-base`).
* `ProtoDartPlugin` class was renamed to `McDartPlugin`.
* `OsFamily` enumeration was introduced to avoid dependency on Ant runtime _just_ for checking that the current operating system is Windows.
* Resolving of imports was refactored introducing `SourceLine` class responsible for the resolution.
* Some other class names were improved.

### DIFF
--- a/tools/mc-dart/src/test/java/io/spine/tools/mc/dart/gradle/McDartPluginTest.java
+++ b/tools/mc-dart/src/test/java/io/spine/tools/mc/dart/gradle/McDartPluginTest.java
@@ -24,7 +24,7 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-package io.spine.dart.gradle;
+package io.spine.tools.mc.dart.gradle;
 
 import com.google.errorprone.annotations.CanIgnoreReturnValue;
 import io.spine.tools.gradle.TaskName;
@@ -45,7 +45,7 @@ import static io.spine.tools.gradle.ProtoDartTaskName.copyTestGeneratedDart;
 import static io.spine.tools.gradle.ProtoDartTaskName.resolveImports;
 
 @DisplayName("`ProtoDartPlugin` should")
-class ProtoDartPluginTest {
+class McDartPluginTest {
 
     private Project project;
 
@@ -53,7 +53,7 @@ class ProtoDartPluginTest {
     void setUp(@TempDir File dir) {
         project = ProjectBuilder
                 .builder()
-                .withName(ProtoDartPluginTest.class.getName())
+                .withName(McDartPluginTest.class.getName())
                 .withProjectDir(dir)
                 .build();
         project.apply(action -> action.plugin("java"));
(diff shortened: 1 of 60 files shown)


tools/mc-dart/src/test/java/io/spine/tools/mc/dart/gradle/McDartPluginTest.java
```diff
@@ -45,15 +45,15 @@
 import static io.spine.tools.gradle.ProtoDartTaskName.resolveImports;
 
 @DisplayName("`ProtoDartPlugin` should")
```
armiol, line 47: Let's adjust the display name as well.

tools/tool-base/src/main/java/io/spine/code/fs/FileReference.java
```diff
@@ -50,15 +51,18 @@ private FileReference(String value) {
...
+    }
+
+    /**
+     * Creates a new instance using the passed file path.
      */
-    public static FileReference of(String value) {
-        checkNotEmptyOrBlank(value);
-        return new FileReference(value);
+    public static FileReference of(Path path) {
+        return of(path.toString());
```
armiol, line 65: Let's check `path` for `null` beforehand.

tools/tool-base/src/main/java/io/spine/tools/OsFamily.java
```diff
@@ -0,0 +1,113 @@
...
+ * Detects current operating system properties.
+ *
+ * <p>Based on {@code org.apache.tools.ant.taskdefs.condition.Os}.
+ */
+@SuppressWarnings("AccessOfSystemProperties") // to get current OS props.
+public enum OsFamily {
+
+    Windows,
+
+    Win9x {
```
armiol, line 41: Considering this entry, and other entries such as OS/2, this whole enum looks really weird to me.

I would trim it to Windows, Unix and macOS.

tools/tool-base/src/main/java/io/spine/tools/dart/code/SourceLine.java
```diff
@@ -0,0 +1,150 @@
...
+import java.nio.file.Path;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static java.lang.String.format;
+import static java.util.regex.Pattern.compile;
+
+/**
+ * Resolves import statements in a Dart file.
```
armiol, line 45: To me, this is a "line of code in a Dart file". In particular, it resolves imports, if any. 

It still may have other features in future — but with this description we put a bit too strict of boundaries.

armiol (approved):
@alexander-yevsyukov LGTM from my point of view. Please see my comments though.

PR #635, opened 2021-05-04
Improve `tool-base` packages

alexander-yevsyukov:
This PR improves the names of the packages in the `tool-base` artifact:
  * Language-dependent parts were placed under the packages named after the pattern `io.spine.tools.<language>`.
  * Parts that are language-neutral are placed under `io.spine.tools.code` or `io.spine.tools.fs`.

The `io.spine.tools.protoc` package, which currently belongs to the `tool-base` module was not touched.
This package does not really related to `protoc`. Most of its classes are more about Gradle, and for a Java project in particular. So, to reduce the number of changes, it was left as is  and will be changed along with other changes of the `mc-java`.


tools/tool-base/src/main/java/io/spine/tools/java/code/UuidMethodFactory.java
```diff
@@ -31,7 +31,6 @@
 import com.squareup.javapoet.MethodSpec;
 import com.squareup.javapoet.ParameterSpec;
 import com.squareup.javapoet.TypeName;
-import io.spine.code.gen.java.JavaPoetName;
```
armiol, line 34: That's interesting. I mean it somehow lasted through all these years, and almost escaped inevitable death.

So close :)

armiol (approved):
@alexander-yevsyukov LGTM.

tools/tool-base/src/main/java/io/spine/tools/java/code/UuidMethodFactory.java
alexander-yevsyukov: Checkstyle helped.

PR #644, opened 2021-06-02
More Query DSL

armiol:
This changeset enhances the Query DSL introduced to `base` in scope of Spine 2.0.

Highlights:

* Query predicate now supports "unlimited" depth of nested children.
* Predicates may now be converted to their disjunctive normal form — which greatly simplifies further querying of NoSQL storages such as Google Datastore.
* It is now possible to extend a `RecordQuery` instance with more column conditions. That's a syntax sugar for a long procedure of creating a new `Builder` from an existing `Query`, modifying its internals manually, and building it again:

```java

RecordQuery<...> query = ...;

RecordQuery<...> extended = query.and(record -> record.where(Columns.sky).is("blue")
                                                      .where(Columns.grass).is("green"));

```


base/src/main/java/io/spine/query/QueryPredicate.java
```diff
@@ -181,17 +352,134 @@ private Builder(LogicalOperator operator) {
...
+
+        private boolean hasNoParams() {
+            return parameters.isEmpty() && customParameters.isEmpty();
+        }
+
+        /**
+         * Appends this predicate to its parent.
+         *
+         * <p>In case the predicate has only one part (i.e. a single parameter, just one custom
+         * parameter or a child predicate), only this item is appended to the parent.
```
dmdashenkov, line 452:
```suggestion
         * parameter, or a child predicate), only this item is appended to the parent.
```

base/src/main/java/io/spine/query/RecordPredicates.java
```diff
@@ -0,0 +1,45 @@
...
+ * to some existing {@link RecordQuery}.
+ *
+ * @param <I>
+ *         the type of identifiers of queried records
+ * @param <R>
+ *         the type of queried records
+ */
+public interface RecordPredicates<I, R extends Message>
+        extends UnaryOperator<RecordQueryBuilder<I, R>> {
+
```
dmdashenkov, line 44: Please remove this empty line.

base/src/main/java/io/spine/query/RecordQuery.java
```diff
@@ -84,4 +86,47 @@
...
+    }
+
+    /**
+     * Appends a series of records column predicates to this query treating them in disjunction
+     * with those predicates which are already set for querying.
+     *
+     * <p>This method only processes the column predicates. Additional identifier conditions,
+     * field masks, sorting, or limit are ignored.
+     */
+    public final RecordQuery<I, R> either(RecordPredicates<I, R> predicates) {
```
dmdashenkov, line 109: Let's use `and` and `or`. Or maybe `all` and `either`, but not mix the two options.

base/src/main/java/io/spine/query/TransformToDnf.java
```diff
@@ -0,0 +1,175 @@
...
+     * to the grandchild nodes.
+     */
+    private static <R> Expression<R, ?> flatten(Expression<R, ?> expression) {
+        List<Expression<R, ?>> children = expression.children();
+        if (children.isEmpty()) {
+            return expression;
+        }
+        Expression<R, ?> flattened;
+        if (expression.operator() == AND) {
+
```
dmdashenkov, line 119: Please remove this empty line.

dmdashenkov (approved):
LGTM with a few minor issues.

base/src/main/java/io/spine/query/RecordPredicates.java
armiol: I'd though so. That's the IDEA formatting we have.

base/src/main/java/io/spine/query/Distribution.java
```diff
@@ -0,0 +1,307 @@
...
+ */
+
+package io.spine.query;
+
+import java.util.List;
+
+import static io.spine.query.LogicalOperator.AND;
+
+/**
+ * Utilities helping to apply the Distributive law of Boolean algebra to {@link Expression}s.
```
alexander-yevsyukov, line 34: I think that “Law” and “Algebra” can be capitalized in this context. Please consider.

base/src/main/java/io/spine/query/Distribution.java
```diff
@@ -0,0 +1,307 @@
...
+                AndExpression<R> childAnd = childBuilder.addCustomParam(firstCustom)
+                                                        .build();
+                result.addExpression(childAnd);
+            }
+        }
+    }
+
+    @SuppressWarnings("MethodWithMultipleLoops")    /* Avoiding multiple similar methods. */
+    private static <R>
+    void cartesianSimpleParams(List<SubjectParameter<R, ?, ?>> paramsOfFirst,
```
alexander-yevsyukov, line 267: This and previous method look similar. The methods itself are complex too. Please refactor. 

Consider having a helper class that does the Cartesian thing. As a side effect, you'd be able to extract nested `for` loops into methods.

base/src/test/java/io/spine/query/RecordQueryBuilderTest.java
```diff
@@ -154,22 +155,41 @@ void byEitherColumn() {
...
+            ImmutableList<QueryPredicate<Manufacturer>> children = rootPredicate.children();
+            assertThat(children.size()).isEqualTo(1);
+            QueryPredicate<Manufacturer> either = children.get(0);
+            assertThat(either.operator()).isEqualTo(OR);
+            assertThat(either.customParameters()).hasSize(0);
+            ImmutableList<SubjectParameter<Manufacturer, ?, ?>> params = either.parameters();
+            assertThat(params).hasSize(2);
+
+            assertHasParamValue(params, isin, EQUALS, isinValue);
+            assertHasParamValue(params, is_traded, EQUALS, stocksAreTraded);
```
alexander-yevsyukov, line 175: This test has several assertions. It looks like there are three aspects that we test. Can we have it as a nested class with sections of this method transformed into methods?

alexander-yevsyukov (changes requested):
Almost LGTM, please see my comments.

base/src/main/java/io/spine/query/RecordQuery.java
armiol: Semantically, this `Query.either(..)` is a shortcut to `Query.toBuilder().either(..).build()`. And while these `either()`s are methods of two different types — the former belongs to `Query`, the latter belongs to `QueryBulder` — it would be nice to use the same word.

Therefore, as discussed vocally, we will keep `either` here.

base/src/test/java/io/spine/query/RecordQueryBuilderTest.java
armiol: @alexander-yevsyukov These are already nested classes, which test different aspects :) Besides, these assertions test basically a single thing, which is an exact number and types of resulting parameters joined with an expected operator. E.g. "We expect an outcome to be disjunction of two parameters with certain values" . Not sure how to split it.

Therefore, I would prefer not to make it deeper.

armiol:
@alexander-yevsyukov @dmdashenkov most comments are addressed by the changes in code. The remainder is either discussed vocally, or have my replies.

Please have a look.

alexander-yevsyukov (approved):
LGTM

PR #651, opened 2021-06-24
Rework Model Compiler code generation config API

dmdashenkov:
# Changes to the Model Compiler API

In this PR we change the Gradle API of the Model Compiler (`mc-java`). The change is related to the configuration which targets code generation actions.

In particular, the Java code generation configurations are:
 1. Joined under the `java { }` block.
 2. Grouped by use case into `events { }`, `rejections { }`, `commands { }`, `entities { }`, `messages(..)`, and `validation { }`.

Each of the configuration groups allows tweaking all the related code generation properties.

The configuration is preset with the default values used in Spine, so that, in the simplest scenario, the users don't have to configure anything.

For users of Model Compiler who configure it in the "default" way (see [`model-compiler.gradle`](https://github.com/SpineEventEngine/config/blob/master/buildSrc/src/main/groovy/model-compiler.gradle)) the migration to the new configuration API consists of deleting the whole configuration and moving on.

For users who customize their code generation in other ways, it might be helpful to look into the `io.spine.tools.mc.java.codegen.JavaCodegenConfig` docs for a better understanding of the new API.

## Example
```kotlin
modelCompiler {

    java {

        entities {
            options.add("my_option") 
            markAs("com.acme.MyEntityState")
        }

        events {
            markFieldsAs("com.acme.EventField")
        }
    }
}
```

For the full set of configuration parameters, see `io.spine.tools.mc.java.codegen.JavaCodegenConfig`.

## Breaking changes

Some `modelCompiler` extension fields are deleted, namely `interfaces`, `methods`, `entityQueries`, `fields`, `generateValidatingBuilders`, and `generateValidation`. The equivalent settings can now be made through the `java { }` DSL.


dmdashenkov:

tests/entity-queries/build.gradle.kts
```diff
@@ -24,33 +24,10 @@
...
-        generate(true)
-    }
-
-    interfaces {
-        mark(messages().inFiles(suffix("commands.proto")), asType("io.spine.base.CommandMessage"))
-        mark(messages().inFiles(suffix("events.proto")), asType("io.spine.base.EventMessage"))
-        mark(messages().inFiles(suffix("rejections.proto")), asType("io.spine.base.RejectionMessage"))
-        mark(messages().uuid(), asType("io.spine.base.UuidValue"))
-        mark(messages().entityState(), asType("io.spine.base.EntityState"))
+    codegen {
```
alexander-yevsyukov, line 28: As discussed vocally, let's rename `codegen` to `java`. This way it would be obvious to which language this configuration applies.

tests/entity-queries/build.gradle.kts
```diff
@@ -24,33 +24,10 @@
...
+        forMessage("spine.tools.column.ProjectName") {
```
alexander-yevsyukov, line 29: As discussed vocally, since we're going to rename `config` to `java`, let's avoid the `for` prefix in `forMessages`, `forEntities`, `forEvents` and the like.

base/src/main/java/io/spine/query/QueryPredicate.java
```diff
@@ -427,18 +427,18 @@ private void optimizeForOnlyChild() {
...
-                if(differentChildrenCount == 0) {
-                    for (QueryPredicate<R> child : children) {
-                        copyParams(child, this);
-                        copyCustomParams(child, this);
-                        copyChildren(child, this);
-                    }
-                    children.clear();
+            long differentChildrenCount =
+                    children.stream()
+                            .filter(c -> c.operator != operator())
```
alexander-yevsyukov, line 432: Do we need to call the `operator()` method for each child? Can we extract the result of the call into a variable instead?

base/src/main/java/io/spine/type/MessageType.java
```diff
@@ -391,5 +393,23 @@ public boolean isUuidValue() {
...
+     *
+     * <p>The option must be known at runtime. An uninterpreted option cannot be found this way.
+     *
+     * @param optionName
+     *         name of the option
+     * @return {@code true} if this type is marked with an option with the given name,
+     *         {@code false} otherwise
+     */
+    public boolean hasOption(String optionName) {
+        MessageOptions options = descriptor().getOptions();
```
alexander-yevsyukov, line 410: Please check for nullity and emptiness of the `optionName`.

base/src/main/java/io/spine/type/MessageType.java
```diff
@@ -391,5 +393,23 @@ public boolean isUuidValue() {
...
+        Set<FieldDescriptor> presentOptions = options.getAllFields().keySet();
+        return presentOptions.stream()
+                             .anyMatch(op -> op.getName().equals(optionName));
```
alexander-yevsyukov, line 413: Assuming the previous request for this method, please flip equals. It would read a bit simpler.

tools/mc-java-protoc/src/main/java/io/spine/tools/mc/java/protoc/EntityMatcher.java
```diff
@@ -0,0 +1,66 @@
...
+    @Override
+    public boolean test(MessageType type) {
+        List<ProtoOption> options = entities.getOptionList();
+        boolean match = options.stream()
+                               .map(OptionMatcher::new)
+                               .anyMatch(matcher -> matcher.test(type));
+        if (match) {
+            return true;
+        }
+        List<FilePattern> patterns = entities.getPatternList();
```
alexander-yevsyukov, line 61: Please extract two methods, one for checking agains options, and another for checking against patterns.

tools/mc-java-protoc/src/main/java/io/spine/tools/mc/java/protoc/field/FieldGen.java
```diff
@@ -73,27 +71,61 @@ private FieldGen(ImmutableList<CodeGenerationTask> tasks) {
...
+                         )).collect(toImmutableList());
+    }
+
+    private static Optional<GenerateFieldsByPattern> taskFor(ForMessages forMessages) {
+        GenerateFields generateFields = forMessages.getGenerateFields();
+        if (!generateFields.hasSuperclass()) {
+            return Optional.empty();
+        }
+        Pattern pattern = forMessages.getPattern();
+        GenerateFieldsByPattern task = new GenerateFieldsByPattern(
```
alexander-yevsyukov, line 121: We have this `new GeneratedFieldsByPattern()` thing twice in this file. Please consider extracting a method which makes it easier to read and understand.

tools/mc-java-protoc/src/main/java/io/spine/tools/mc/java/protoc/field/FieldGenerationTask.java
```diff
@@ -27,34 +27,38 @@
...
 
-    private final ClassName fieldSupertype;
+    private final JavaClassName fieldSupertype;
     private final FieldFactory factory;
 
-    FieldGenerationTask(ClassName fieldSupertype, FieldFactory factory) {
+    FieldGenerationTask(JavaClassName fieldSupertype, FieldFactory factory) {
         this.fieldSupertype = fieldSupertype;
         this.factory = factory;
+        checkNotEmptyOrBlank(fieldSupertype.getCanonical());
```
alexander-yevsyukov, line 52: Isn't it a bit (two lines) too late to check for a precondition? Why don't we do when the method starts?

tools/mc-java-protoc/src/main/java/io/spine/tools/mc/java/protoc/message/GenerateNestedClasses.java
```diff
@@ -28,28 +28,33 @@
...
 
+import java.util.function.Predicate;
+
 import static com.google.common.base.Preconditions.checkNotNull;
 import static io.spine.util.Preconditions2.checkNotDefaultArg;
 
 /**
- * Generates nested classes for the supplied type based on
- * {@link ConfigByPattern pattern configuration}.
+ * Generates nested classes for the supplied type based on a
```
alexander-yevsyukov, line 45: Please have the article next to the following word.

tools/mc-java-protoc/src/main/java/io/spine/tools/mc/java/protoc/method/MethodGen.java
```diff
@@ -70,13 +71,23 @@ public static MethodGen instance(SpineProtocConfig spineProtocConfig) {
...
+                    .map(name -> new GenerateUuidMethods(classLoader, name))
+                    .forEach(tasks::add);
         }
-        for (ConfigByPattern byPattern : config.getFactoryByPatternList()) {
-            tasks.add(new GenerateMethods(classLoader, byPattern));
+        for (ForMessages forMessages : spineProtocConfig.getMessagesList()) {
+            Pattern pattern = forMessages.getPattern();
+            forMessages.getGenerateMethodsList()
+                       .stream()
+                       .map(generate -> new GenerateMethods(
```
alexander-yevsyukov, line 87: The snippet `generate -> new GenerateMethods(...)` looks as a popular thing in this update. I commented on this already in relation to one file, but we're having a number of similar looking blocks, where we tend to tell similar things. We need to generalize it somehow.

tools/mc-java/src/main/java/io/spine/tools/mc/java/codegen/MessagesConfig.java
```diff
@@ -0,0 +1,137 @@
...
+        return result.build();
+    }
+
+    private Set<GenerateMethods> generateMethods() {
+        return methodFactories.get()
+                              .stream()
+                              .map(name -> MethodFactoryName.newBuilder()
+                                      .setClassName(className(name))
+                                      .build())
+                              .map(name -> GenerateMethods.newBuilder()
```
alexander-yevsyukov, line 120: Please extract a method which does the `name` -> `MethodFactoryName` -> `GenerateMethods` conversion. This way we'd have only one `map`, and it would be easier to understand.

tools/mc-java/src/main/java/io/spine/tools/mc/java/codegen/MessagesConfig.java
```diff
@@ -0,0 +1,137 @@
...
+                              .collect(toSet());
+    }
+
+    private Set<GenerateNestedClasses> generateNestedClasses() {
+        return nestedClassFactories.get()
+                                   .stream()
+                                   .map(name -> NestedClassFactoryName.newBuilder()
+                                           .setClassName(className(name))
+                                           .build())
+                                   .map(name -> GenerateNestedClasses.newBuilder()
```
alexander-yevsyukov, line 132: Please simplify the `name` -> `GenerateNestedClasses` conversion by extracting it to a method.

tools/mc-java/src/test/kotlin/io/spine/tools/mc/java/codegen/CodegenTest.kt
```diff
@@ -0,0 +1,414 @@
...
+import io.spine.tools.protoc.Pattern
+import io.spine.tools.protoc.ProtoTypeName
+import io.spine.tools.protoc.TypePattern
+import org.gradle.testfixtures.ProjectBuilder
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.DisplayName
+import org.junit.jupiter.api.Nested
+import org.junit.jupiter.api.Test
+
+class `'codegen { }' block should` {
```
alexander-yevsyukov, line 55: Please don't forget to update this test suite name too.

alexander-yevsyukov (changes requested):
Please see my comments.

base/src/main/java/io/spine/query/QueryPredicate.java
armiol: @alexander-yevsyukov 
This is actually a plain getter returning a constant value. So `javac` is going to replace it with the value anyway. 
I was thinking of shortening the method rather than of explicit optimisations, such as extracting into a variable.

@dmdashenkov thanks for formatting this piece!

tools/mc-java-protoc/src/main/java/io/spine/tools/mc/java/protoc/CodeGenerator.java
```diff
@@ -222,7 +222,7 @@ private void checkCompilerVersion(CodeGeneratorRequest request) {
 
     private static File concatContent(File left, File right) {
         return left.toBuilder()
-                   .setContent(left.getContent() + right.getContent())
+                   .setContent(left.getContent() + System.lineSeparator() + right.getContent())
```
armiol, line 225: While I realise this is a correct version, it as well looks a breaking change to me. 

Please consider renaming the method, or at least documenting this behaviour.

tools/mc-java-protoc/src/main/java/io/spine/tools/mc/java/protoc/field/FieldGen.java
```diff
@@ -73,27 +71,61 @@ private FieldGen(ImmutableList<CodeGenerationTask> tasks) {
...
     public static FieldGen instance(SpineProtocConfig spineProtocConfig) {
         checkNotNull(spineProtocConfig);
-        AddFields config = spineProtocConfig.getAddFields();
-
         ImmutableList.Builder<CodeGenerationTask> tasks = ImmutableList.builder();
-        EntityStateConfig entityStateConfig = config.getEntityStateConfig();
-        if (isNotDefault(entityStateConfig)) {
-            tasks.add(new GenerateEntityStateFields(entityStateConfig, factory));
+
+        if (spineProtocConfig.hasEntities()) {
```
armiol, line 78: I would consider extracting these `if`s into methods right away, since I'd expect their number grows in the nearest future.

tools/mc-java-protoc/src/main/java/io/spine/tools/mc/java/protoc/message/ImplementByPattern.java
```diff
@@ -42,13 +46,18 @@
...
+        checkNotDefaultArg(pattern);
+        this.matcher = new FilePatternMatcher(pattern);
+    }
 
-    ImplementByPattern(ConfigByPattern config) {
-        super(config.getValue());
-        FilePattern filePattern = config.getPattern();
-        checkNotDefaultArg(filePattern);
-        this.matcher = new FilePatternMatcher(filePattern);
+    ImplementByPattern(JavaClassName config, Pattern pattern) {
```
armiol, line 57: To a naked eye, this ctor seems odd, as the class-level description only mentions `FilePattern`s, not `Pattern`s.

Please consider documenting both ctors.

tools/mc-java-protoc/src/main/java/io/spine/tools/mc/java/protoc/message/ImplementInterface.java
```diff
@@ -30,20 +30,18 @@
...
 import io.spine.tools.mc.java.protoc.CompilerOutput;
+import io.spine.tools.protoc.JavaClassName;
 import io.spine.type.MessageType;
 
-import static io.spine.util.Preconditions2.checkNotEmptyOrBlank;
-
 /**
  * An abstract base for the interface code generation tasks.
  */
 public abstract class ImplementInterface implements CodeGenerationTask {
```
armiol, line 39: Do we expect it to have custom or user-defined descendants? If so, I would describe that in the documentation and mark it `@SPI`. 

However, to me, this class should safely become package-private for now.

tools/mc-java-protoc/src/test/java/io/spine/tools/mc/java/protoc/given/CodeGeneratorRequestGiven.java
```diff
@@ -78,63 +82,83 @@ private CodeGeneratorRequestGiven() {
...
+     */
+    public static GenerateMethods generateMethods(Class<?> factory) {
+        MethodFactoryName factoryName = methodFactory(factory);
+        return GenerateMethods
+                .newBuilder()
+                .setFactory(factoryName)
+                .build();
+    }
+
+    public static MethodFactoryName methodFactory(Class<?> cls) {
```
armiol, line 123: As long as we started documenting `public static` API level in this type, I'd propose to do it here as well.

tools/mc-java-protoc/src/test/java/io/spine/tools/mc/java/protoc/given/CodeGeneratorRequestGiven.java
```diff
@@ -78,63 +82,83 @@ private CodeGeneratorRequestGiven() {
...
-                .setAddNestedClasses(nestedClasses.asProtocConfig())
+                .setFile(filePattern)
                 .build();
-        try (FileOutputStream fos = new FileOutputStream(configPath.toFile())) {
+    }
+
+    /**
+     * Writes the given Protoc config into the given file and obtains the Protoc plugin argument.
+     *
+     * @return the path to the serialized config, encoded in base 64
```
armiol, line 153: "Base64"

tools/mc-java/src/main/java/io/spine/tools/mc/java/codegen/Config.java
```diff
@@ -24,19 +24,20 @@
...
+import com.google.protobuf.Message;
 
 /**
- * A selector which signalizes that the configuration should be applied to all messages that
- * represent an entity state.
+ * A Gradle configuration of a certain aspect of Model Compiler code generation.
  *
- * @see Interfaces#mark(IsEntityState, ClassName)
+ * @param <P>
+ *         Protobuf type which serializes this configuration
```
armiol, line 35: I would be more precise here. How about this?
```
Protobuf type reflecting a snapshot of this configuration.
```

tools/mc-java/src/main/java/io/spine/tools/mc/java/codegen/ConfigWithFields.java
```diff
@@ -0,0 +1,84 @@
...
+import static io.spine.tools.mc.java.codegen.Names.className;
+
+/**
+ * A configuration which includes field generation.
+ *
+ * <p>Model Compiler generates type-safe API for filtering messages by fields in queries
+ * and subscriptions.
+ *
+ * @param <P>
+ *         Protobuf type which serializes this configuration
```
armiol, line 44: Basically, my comment (see above) applies to this and similar cases as well.

tools/mc-java/src/main/java/io/spine/tools/mc/java/codegen/ConfigWithFields.java
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    final void convention(@Nullable Class<?> fieldSuperclass) {
+        if (fieldSuperclass != null) {
+            markFieldsAs.convention(fieldSuperclass.getCanonicalName());
+        }
+    }
+
+    /**
+     * Provides a superclass for the field type.
```
armiol, line 63: I would say "Equips the field type with a superclass." This way, it's closer to "generation", when "Provides" seems a bit "look-up"-y to me.

tools/mc-java/src/main/java/io/spine/tools/mc/java/codegen/ConfigWithInterfaces.java
```diff
@@ -0,0 +1,86 @@
...
+    ConfigWithInterfaces(Project p) {
+        super();
+        this.interfaceNames = p.getObjects()
+                               .setProperty(String.class);
+    }
+
+    /**
+     * Configures the associated messages to implement a Java interface with the given name.
+     *
+     * <p>The interface must exist. The declaration of the interface will not be generated.
```
armiol, line 56: "If missing, the declaration of the interface will not be generated."

tools/mc-java/src/main/java/io/spine/tools/mc/java/codegen/EntityConfig.java
```diff
@@ -0,0 +1,138 @@
...
+import static java.util.stream.Collectors.toList;
+
+/**
+ * Configuration for entity state types' code generation.
+ *
+ * <p id="disclaimer">Note. This configuration allows to change the Protobuf types which are
+ * recognized as entities by changing the {@code options} and/or submitting a custom
+ * {@linkplain #includeFiles file pattern}. These are superuser options. Changing the default values
+ * may lead to runtime errors or unexpected behaviour in the Spine Event Engine framework.
+ * Proceed by caution.
```
armiol, line 52: Did you mean "with"?

tools/mc-java/src/main/java/io/spine/tools/mc/java/codegen/MessagesConfig.java
```diff
@@ -0,0 +1,137 @@
...
+        interfaceNames().convention(ImmutableSet.of());
+        methodFactories.convention(ImmutableSet.of());
+        nestedClassFactories.convention(ImmutableSet.of());
+    }
+
+    /**
+     * Specifies a {@link io.spine.tools.protoc.MethodFactory} to generate methods for
+     * the message classes.
+     *
+     * <p>Calling this method multiple times will add provide factories for code generation.
```
armiol, line 79: "provided"

tools/mc-java/src/main/java/io/spine/tools/mc/java/codegen/MessagesConfig.java
```diff
@@ -0,0 +1,137 @@
...
+     */
+    public void generateMethodsWith(String factoryClassName) {
+        methodFactories.add(factoryClassName);
+    }
+
+    /**
+     * Specifies a {@link io.spine.tools.protoc.NestedClassFactory} to generate nested classes
+     * inside the message classes.
+     *
+     * <p>Calling this method multiple times will add provide factories for code generation.
```
armiol, line 92: Same as above.

tools/mc-java/src/main/java/io/spine/tools/mc/java/codegen/Names.java
```diff
@@ -24,24 +24,37 @@
...
+     * Prevents the utility class instantiation.
+     */
+    private Names() {
+    }
 
-    private TestClassLoader() {
+    /**
+     * Obtains a {@code JavaClassName} with the given value.
+     */
+    public static JavaClassName className(String name) {
```
armiol, line 47: Even considering this type is `@Internal`, I would check the arguments before working with them. It's here and in the method below.

tools/mc-java/src/main/java/io/spine/tools/mc/java/codegen/SignalConfig.java
```diff
@@ -0,0 +1,76 @@
...
+ *
+ * <p>The configuration applied to events does not automatically apply to rejections.
+ */
+public final class SignalConfig extends MessageGroupConfig<ForSignals> {
+
+    SignalConfig(Project p) {
+        super(p);
+    }
+
+    void convention(MessageFile file, Class<? extends SignalMessage> interfaceClass) {
```
armiol, line 50: These undocumented methods are strangely package-private. Which probably means they are made for the whole package to use those.

Please consider documenting those.

tools/mc-java/src/main/java/io/spine/tools/mc/java/codegen/UuidConfig.java
```diff
@@ -0,0 +1,93 @@
...
+
+    private final SetProperty<String> methodFactories;
+
+    UuidConfig(Project p) {
+        super(p);
+        methodFactories = p.getObjects()
+                           .setProperty(String.class);
+    }
+
+    void convention(Class<? extends MethodFactory> methodFactory,
```
armiol, line 57: Same as above. I am not sure, but  _may_ they be a part of some common contract? Probably, not (due to the difference in their argument list), but still, please check.

armiol (changes requested):
@dmdashenkov please see my comments.

tools/mc-java-protoc/src/main/java/io/spine/tools/mc/java/protoc/EntityMatcher.java
dmdashenkov: I've refactored this class into a more functional style. Please have a look. There's definitely less duplicate now. I want to see how readable is it to other people.

tools/mc-java-protoc/src/main/java/io/spine/tools/mc/java/protoc/field/FieldGen.java
dmdashenkov: It's a trick. Those are two different overloads of the constructor. One accepts `FilePattern` and the other — `Pattern`.

I've changed the naming a bit so that it's more obvious.

tools/mc-java-protoc/src/main/java/io/spine/tools/mc/java/protoc/method/MethodGen.java
dmdashenkov: Those are all different types though they do look similar.

I find it handy to name them in such a cohesive way.

Let's discuss this in person if necessary.

dmdashenkov:
@alexander-yevsyukov, @armiol, PTAL again.

alexander-yevsyukov:
@dmdashenkov, please update the PR description with new names for the Gradle DSL.

tools/mc-java-protoc/src/main/java/io/spine/tools/mc/java/protoc/field/FieldGen.java
```diff
@@ -73,27 +71,85 @@ private FieldGen(ImmutableList<CodeGenerationTask> tasks) {
...
-    @SuppressWarnings("MethodWithMultipleLoops") // Required to configure code generation tasks.
     public static FieldGen instance(SpineProtocConfig spineProtocConfig) {
         checkNotNull(spineProtocConfig);
-        AddFields config = spineProtocConfig.getAddFields();
-
         ImmutableList.Builder<CodeGenerationTask> tasks = ImmutableList.builder();
-        EntityStateConfig entityStateConfig = config.getEntityStateConfig();
-        if (isNotDefault(entityStateConfig)) {
-            tasks.add(new GenerateEntityStateFields(entityStateConfig, factory));
+        addFromEntities(tasks, spineProtocConfig);
```
alexander-yevsyukov, line 77: Please introduce a helper class which would accept `spineProtocConfig` on constructor, and create `tasks` in this constructor. Under this class you'll be be able to gather de-staticized versions of the below methods. As instance methods they won't need such lengthy parameters.

tools/mc-java/src/main/java/io/spine/tools/mc/java/codegen/MessagesConfig.java
```diff
@@ -0,0 +1,148 @@
...
+                .setFactory(factoryName)
+                .build();
+        return config;
+    }
+
+    private Set<GenerateNestedClasses> generateNestedClasses() {
+        return nestedClassFactories.get()
+                                   .stream()
+                                   .map(MessagesConfig::nestedClassFactoryConfig)
+                                   .collect(toSet());
```
alexander-yevsyukov, line 136: Can we extend `SetProperty` so that we can do the thing `get().stream().map().collect()` in the derived class once and then use it all over the places just passing the arguments of the `map()` call?

alexander-yevsyukov (changes requested):
I understand my tediousness (aka «занудство») is not in the spirit of Friday evening, but I have two more minor refactoring requests. Please see my comments and questions... :)

dmdashenkov:

tools/plugin-base/src/main/kotlin/io/spine/tools/gradle/UsefulSetProperty.kt
```diff
@@ -0,0 +1,71 @@
...
+     *          the project which provides an object factory for creating Gradle properties
+     * @param cls
+     *          Java class of the elements of the set
+     */
+    public constructor(project: Project, cls: Class<T>) : this(project, cls.kotlin)
+
+    /**
+     * Obtains the value of this property and applies the given mapping function to each element.
+     */
+    public fun <O> transform(mapper: (T) -> O): ImmutableSet<O> =
```
alexander-yevsyukov, line 66: Please think of a different letter for the generic parameter. `<0>` looks  a bit misleading. Please also document the generic parameter.

tools/plugin-base/src/main/kotlin/io/spine/tools/gradle/UsefulSetProperty.kt
```diff
@@ -0,0 +1,71 @@
...
+
+import com.google.common.collect.ImmutableSet
+import kotlin.reflect.KClass
+import org.gradle.api.Project
+import org.gradle.api.provider.SetProperty
+
+/**
+ * A Gradle `SetProperty` which has useful utility methods.
+ */
+public class UsefulSetProperty<T : Any>
```
alexander-yevsyukov, line 37: The `Useful` prefix does not add much meaning. If it can be just a set of extension functions, instead of inheritance, as I suggested in my previous comments, so be it.

alexander-yevsyukov (changes requested):
Almost, LGTM. Please see my comments.

alexander-yevsyukov (approved):
Assuming vocally discussed changes, LGTM.

armiol (approved):

PR #662, opened 2021-08-16
Isolate GCAR publishing

dmdashenkov:
in this PR we disable the Google Cloud Artifact Registry Maven repo publishing.

It has proven to be faulty with big artifacts (e.g. fat JARs).  We suspect that the problem has to do with the lifespan of the access token, issued by Google with the help of the Gradle plugin (which we use as a library — `com.google.cloud.artifactregistry`).
[This change](https://github.com/GoogleCloudPlatform/artifact-registry-maven-tools/pull/53) ought to fix the issue, once it is released. Meanwhile, we simply skip the GCAR Maven repo.

Once the mentioned change is applied, we should update our dependencies and check again. To enable publishing, simply restore the commented YAML code in `.github/workflows/publish.yml`.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -35,15 +35,19 @@ jobs:
         env:
           GCS_CREDENTIALS_KEY: ${{ secrets.GCS_CREDENTIALS_KEY }}
 
-      - name: Decrypt GCAR credentials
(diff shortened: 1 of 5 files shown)


dmdashenkov:
@armiol, PTAL.

armiol (approved):

PR #668, opened 2021-08-25
Unify extension property names

alexander-yevsyukov:
This PR:

  * Unifies the name of properties of Gradle plugin extensions we have for different languages supported by Spine Model Compiler.
  * Eliminates `GradleExtension` class in favor of Kotlin-based `Project` extensions that do the same job, and do not require having the class hierarchy of extensions just for that. 

In addition to this, language-neutral Model Compiler plugin ("io.spine.tools:spine-mc") is introduced. Currently the plugin only creates an extension which comes under a temporary name `modelCompiler2`, and does nothing more.

Following PRs would extend the plugin, and make language-related plugins use it for common language-neutral properties.


tools/mc-java/build.gradle.kts
```diff
@@ -36,6 +36,7 @@ var protocPluginDependency: Dependency? = null
 val spineVersion: String by extra
 
 dependencies {
+    //api(project(":mc"))
```
dmdashenkov, line 39: Please resolve this commented code line.

tools/mc/src/main/kotlin/io/spine/tools/mc/gradle/McExtension.kt
```diff
@@ -0,0 +1,83 @@
...
+
+        testDescriptorSetFile = of.fileProperty()
+            .convention(file(project.defaultTestDescriptors))
+    }
+
+    private fun register() {
+        project.extensions.add(javaClass, name, this)
+    }
+
+    companion object Companion : Logging {
```
dmdashenkov, line 73: Why do we need to name this `companion object`?
Also, `Companion` is the default JVM name for it anyway.

dmdashenkov (changes requested):
Please see my comments.

dmdashenkov (approved):

Issue #670, opened 2021-08-26
Make builds faster


alexander-yevsyukov:
The build of `base` is currently about 20 minutes or so. It hinders the development much.

The biggest set of “contributors” to the build time is test Gradle projects that are created and executed. It is expected that if we do that using in-memory file system, things would work faster.

## Failed Approach: custom `FileSystem` inside Java

One of the ways that I already tried is to substitute an in-memory `FileSystem` for creating `TempDir`. The options were [Google Jimfs](https://github.com/google/jimfs) and [Memory File System](https://github.com/marschall/memoryfilesystem).

I even tried implementing switching of disk vs. memory, but the attempt failed because of the following.

1. Implementations of non-default `FileSystem`'s do not support conversion `Path.toFile()`. All of them. They throw `UnsupportedOperationException`. It's a part of the contract of `Path`.

2. The `GradleRunner` that we use for creating test Gradle project has only methods which accept `File` instances. This API does not accept `Path` instances. The method we use is `GradleRunner.withProjectDir(File)`.

## Suggested Approach: virtual in-memory drive (initially under macOS)

A possible solution would be to:

1. Create a virtual in-memory drive using Java `Process` API and [command line utilities of macOS](https://eshop.macsales.com/blog/46348-how-to-create-and-use-a-ram-disk-with-your-mac-warnings-included/).
2. Switch the value of Java System property `"java.io.tmpdir"` to a sub-directory on the created drive.
3. Arrange [JUnit 5 extension](https://github.com/marschall/memoryfilesystem#next-steps-junit-5) which does the trick and arranges all the necessary clean-ups.

alexander-yevsyukov:
Addressed for this repository by extracting `tool-base` into a separate repository.

PR #671, opened 2021-08-27
Reduce build time

alexander-yevsyukov:
This PR partially addresses #670. On local tests, the build time reduced from 18 to 13 minutes. It's not much, but still saves some life... 

The build time was reduced by reducing the number of test Gradle projects created and compiled. So, what we did on `@BeforeEach`, we do now `@BeforeAll`.

Another improvement of this PR is fixing the way we treat temporary directories created by tests. They were not deleted, and accumulated under the root temporary directory. The `TempDir` instances now are created under the `io.spine.testing` directory, which is deleted on JVM exit. This way it's easy to see «кто насрал», even if the cleanup fails.

More work on making our treatment on temporary disk space needs to be done (https://github.com/SpineEventEngine/config/issues/240).


tools/javadoc-style/src/main/java/io/spine/tools/javadoc/style/LineFormatting.java
```diff
@@ -49,15 +49,18 @@
...
+                Splitter.on(separator)
+                        .splitToList(text);
+        List<String> formattedLines =
+                lines.stream()
+                     .map(this::formatLine)
+                     .collect(toList());
+        String result =
+                Joiner.on(separator)
+                      .join(formattedLines);
+        return result;
```
yuri-sergiichuk, line 63: `Splitter` and `Joiner` instances are thread-safe and it is usually recommended to cache them as there's no need to create them each time the method is called. Especially taking into account the static nature of the `separator` here.
Also, `Splitter` has a convenient `splitToStream` method which we can use instead of creating a list first and then converting it to a stream and back to the list.

```
        String separator = System.lineSeparator(); 
        String result = Splitter.on(separator)
                .splitToStream(text)
                .map(this::formatLine)
                .collect(Collectors.joining(separator));
```

tools/javadoc-style/src/test/java/io/spine/tools/javadoc/style/LineFormattingTest.java
```diff
@@ -32,26 +32,33 @@
...
+
+        assertThat(formattedLines)
+                .isEqualTo(expectedLines);
     }
 
-    private static class ALineFormatting extends LineFormatting {
+    /**
+     * A stub formatting which simply returns the passed line.
+     */
+    private static class NothingFormatting extends LineFormatting {
```
yuri-sergiichuk, line 61: `NoOpFormatting`?

tools/mc-java/src/test/java/io/spine/tools/mc/java/gradle/AnnotationProcessorConfigurationTest.java
```diff
@@ -26,55 +26,46 @@
...
-    void setUp() {
-        project = StubProject.createFor(getClass()).get();
-        projectConfigs = project.getConfigurations();
-        preprocessorConfig = projectConfigs.getByName(annotationProcessor.value());
+    @BeforeAll
+    static void createProject() {
+        project = StubProject.createFor(AnnotationProcessorConfigurationTest.class).get();
+        configurations = project.getConfigurations();
+        preprocessorConfig = configurations.getByName(annotationProcessor.value());
     }
```
yuri-sergiichuk, line 59: Does this approach lead to any side effects? As far as I understand, the project is now created only once and a series of tests are run on the same project. Even though it improves the performance, I'd probably double-check the code to be side-effect-free, or maybe add some notes on similar usages.

tools/mc-java/src/test/java/io/spine/tools/mc/java/gradle/AnnotationProcessorConfigurationTest.java
```diff
@@ -26,55 +26,46 @@
...
-    }
-
-    @Test
-    @DisplayName("create and return annotation processor config if it does not exist")
+    @DisplayName("create annotation processor config if it does not exist")
     void createAndReturnAnnotationProcessorConfigIfItDoesNotExist() {
-        projectConfigs.remove(preprocessorConfig);
-        assertNull(projectConfigs.findByName(annotationProcessor.value()));
+        configurations.remove(preprocessorConfig);
+        assertNull(configurations.findByName(annotationProcessor.value()));
```
yuri-sergiichuk, line 65: Let's replace `assertNull` with `Truth.assertThat().isNull`. We're usually migrating the code to Truth assertions while touching up some tests.

tools/mc-java/src/test/java/io/spine/tools/mc/java/gradle/AnnotatorPluginTest.java
```diff
@@ -99,49 +105,49 @@ void createTempDir() {
         @Test
         @DisplayName("if file option is true")
         void ifFileOptionSet() throws IOException {
-            checkNestedTypesAnnotations(INTERNAL_ALL, true);
+            checkNestedTypesAnnotations(INTERNAL_ALL.fileName(), true);
```
yuri-sergiichuk, line 108: Maybe we can just have an override which does this `fileName()` call internally?

tools/mc-java/src/test/java/io/spine/tools/mc/java/gradle/McJavaExtensionChecksTest.java
```diff
@@ -0,0 +1,78 @@
...
+import org.junit.jupiter.api.Test;
+
+import java.io.File;
+
+import static io.spine.tools.mc.java.gradle.McJavaExtension.getSpineCheckSeverity;
+import static io.spine.tools.mc.java.gradle.given.ModelCompilerTestEnv.MC_JAVA_GRADLE_PLUGIN_ID;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNull;
+
+@DisplayName("`McJavaExtension` for checks return")
```
yuri-sergiichuk, line 43:
```suggestion
@DisplayName("`McJavaExtension` for checks should return")
```

tools/mc-java/src/test/java/io/spine/tools/mc/java/gradle/McJavaExtensionChecksTest.java
```diff
@@ -0,0 +1,78 @@
...
+        project.getPluginManager()
+               .apply(MC_JAVA_GRADLE_PLUGIN_ID);
+    }
+
+    @Test
+    @DisplayName("severity, if set")
+    void specifiedValue() {
+        spineProtobuf().defaultCheckSeverity = Severity.ERROR;
+        Severity actualSeverity = getSpineCheckSeverity(project);
+        assertEquals(spineProtobuf().defaultCheckSeverity, actualSeverity);
```
yuri-sergiichuk, line 64: Let's use `Truth` over `Assertions`.

tools/mc-java/src/test/java/io/spine/tools/mc/java/gradle/McJavaExtensionDirsTest.java
```diff
@@ -0,0 +1,154 @@
...
+        private List<String> actualDirs() {
+            return McJavaExtension.getDirsToClean(project);
+        }
+    }
+
+    private void assertNotEmptyAndIsInProjectDir(String path) {
+        assertFalse(path.trim()
+                        .isEmpty());
+        assertTrue(path.startsWith(project.getProjectDir()
+                                          .getAbsolutePath()));
```
yuri-sergiichuk, line 147: We can replace this `true` assertion with a better one from `Truth`:
```
assertThat(path)
    .startsWith(project.getProjectDir().toString())
```

In case of a failure it'll immediately provide the different and not just tell that smth is not `true`.

tools/mc-java/src/test/java/io/spine/tools/mc/java/gradle/McJavaExtensionDirsTest.java
```diff
@@ -0,0 +1,154 @@
...
+        }
+
+        private List<String> actualDirs() {
+            return McJavaExtension.getDirsToClean(project);
+        }
+    }
+
+    private void assertNotEmptyAndIsInProjectDir(String path) {
+        assertFalse(path.trim()
+                        .isEmpty());
```
yuri-sergiichuk, line 145: Same here. Let's use:
```
assertThat(path.trim())
    .isNotEmpty();
```

yuri-sergiichuk (approved):
@alexander-yevsyukov overall LGTM. Please consider a couple of suggestions and comments below prior to merging the PR.

My only concern with using static (per test suite and not per test case) projects is possible issues with working with a non-fresh project. If that's not smth that may anyhow mess up the tests, I'm with both hands-on merging the PR.

base/src/main/java/io/spine/io/Files2.java
```diff
@@ -267,4 +271,27 @@ public static File toAbsolute(String path) {
...
+     * @see Runtime#addShutdownHook(Thread)
+     */
+    public static void deleteRecursivelyOnShutdownHook(Path directory) {
+        checkNotNull(directory);
+        Runtime runtime = Runtime.getRuntime();
+        runtime.addShutdownHook(new Thread(() -> deleteRecursively(directory)));
+    }
+
+    private static void deleteRecursively(Path directory) {
+        boolean success = FilesKt.deleteRecursively(directory.toFile());
```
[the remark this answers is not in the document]
alexander-yevsyukov: We have Kotlin dependency already. Making a simple call to Kotlin is easy. It does what we need, and it's just one call.

Also, we want more Kotlin, hoping to have a multi-platform backend in v3.

testlib/src/main/java/io/spine/testing/TempDir.java
```diff
@@ -43,10 +48,48 @@
...
+        PackageName packageName = PackageName.of(TempDir.class);
+        Path baseDir = Paths.get(tmpDir, packageName.toString());
+        if (!Files.exists(baseDir)) {
+            try {
+                Files.createDirectories(baseDir);
+            } catch (IOException e) {
+                throw newIllegalStateException(e, "Unable to create `%s`.", baseDir);
+            }
+        }
+        return baseDir;
```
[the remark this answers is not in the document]
alexander-yevsyukov: I tried what you suggest. It brings more code. I've extracted a method instead.

tools/mc-java/src/test/java/io/spine/tools/mc/java/gradle/AnnotationProcessorConfigurationTest.java
alexander-yevsyukov: I understand your concerns. There were mine too, and I've stumbled upon a couple of such side effects. I handled them, and I'm going to continue in another PR.

I don't see any other ways for improving the build performance when working with Gradle things.

PR #672, opened 2021-08-30
Improve Javadoc Style plugin

alexander-yevsyukov:
This PR separates the “meat” of Javadoc style improvement of Gradle-related things by sub-packages.  This is good per se, but it will also allow to reduce time on testing Gradle-related things.


tools/javadoc-style/src/main/java/io/spine/tools/javadoc/style/formatting/JavadocStyler.java
```diff
@@ -43,29 +47,50 @@
...
+    @VisibleForTesting
+    JavadocStyler(Formatting... actions) {
         this.actions = ImmutableList.copyOf(actions);
     }
 
+    /**
+     * Improves the style for all Java files in the specified directory, including sub-directories.
+     */
+    public static void applyFormattingAt(Path directory) {
+        if (!Files.exists(directory)) {
```
armiol, line 72: `Files.exists(path)` has its first parameter annotated as `@NonNull`. So, let's check it for `null` before passing on.

armiol (approved):
@alexander-yevsyukov LGTM in general. Please see my comment.

Also, please consider bumping the library version.

PR #674, opened 2021-08-31
Do not trigger the updates of dependant repositories

armiol:
Previously, upon each successful build we attempted to update all repositories dependant onto `base` artifacts. In particular, we wanted to increment their version and put a freshly built dependency onto `base` libs into their dependencies.

However, this mechanism does not work properly. Therefore, PR switches it off.


armiol:
@alexander-yevsyukov @dmdashenkov PTAL.

dmdashenkov (approved):

alexander-yevsyukov (approved):
LGTM

PR #675, opened 2021-09-03
Comparable messages

dmdashenkov:
In this PR we introduce the `compare_by` option. It allows users to specify the natural ordering strategy for messages, based on the values of their fields.

Example:
```proto
message LocalDate {
    option (compare_by) = {
        field: "year" field: "month" field: "day"
    };

    int32 year = 1;
    Month month = 2;
    uint32 day = 3;
}
```

Such declaration indicates that the value must be compared by the fields `year`, `month`, and `day` in that order.
Recursive comparisons are also supported:

```proto
message LocalDateTime {
    option (compare_by) = {
        field: "date" field: "time"
    };

    LocalDate date = 1;
    LocalTime time = 2;
}
```

Here, both `LocalDate` and `LocalTime` must be marked with the `(compare_by)` option as well as `LocalDateTime`. 

This feature is applicable to any Protobuf message.

**Note.** This PR merely provides a specification for the feature. The implementation, based on ProtoData, will come in a future PR. In order to avoid a circular dependency, the implementation will not declare a dependency on `:base`, and the new option must come from a transitive dependency via ProtoData.

### DIFF
--- a/base/src/main/proto/spine/options.proto
+++ b/base/src/main/proto/spine/options.proto
@@ -858,7 +858,7 @@ message CompareByOption {
     // The allowed field types are:
     //  - any number type;
     //  - `bool` (false is less than true);
-    //  - `string` (alphabetical order);
+    //  - `string` (in the order of respective Unicode values);
     //  - enumerations (following the order of numbers associated with each constant);
     //  - messages marked with `(compare_by)`.
     //
--- a/base/src/main/proto/spine/options.proto
+++ b/base/src/main/proto/spine/options.proto
@@ -451,9 +451,9 @@ extend google.protobuf.MessageOptions {
 
     // Specifies the natural ordering strategy for this type.
     //
-    // Code generators should generate language-specific comparisons based on the fields paths.
+    // Code generators should generate language-specific comparisons based on the field paths.
     //
-    // Runtime comparators may use reflection API to compare field values sequentially.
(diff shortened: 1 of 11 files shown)


dmdashenkov:
@armiol, PTAL.

base/src/main/proto/spine/options.proto
```diff
@@ -836,3 +844,42 @@ message IsOption {
...
     string java_type = 2;
 }
+
+// Defines the way to compare two messages of the same type to one another.
+//
+// Comparisons can be used to sort values.
+//
+// See the `(compare_by)` option.
+//
+message CompareOption {
```
armiol, line 854: As discussed, let's call this `CompareByOption`, to leave `Compare...` unengaged for future use.

base/src/main/proto/spine/options.proto
```diff
@@ -836,3 +844,42 @@ message IsOption {
...
+
+    // Field paths used for comparisons.
+    //
+    // The allowed field types are:
+    //  - any number type;
+    //  - `bool` (false is less than true);
+    //  - `string` (alphabetical order);
```
armiol, line 861: "in the order of respective Unicode values"

armiol (approved):
@dmdashenkov LGTM with some discussion results.

PR #677, opened 2021-09-08
Return of the GCAR

dmdashenkov:
In this PR we enable publishing artifacts to the Google Cloud Artifact Registry.

Previously, there was an issue with uploading large artifacts. It seems to be gone now.

The version is bumped to `2.0.0-SNAPSHOT.55`, since `SNAPSHOT.54` was "eaten up" while testing the publishing on CI.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,55 +1,53 @@
 name: Publish
 
-on:
-  push:
-    branches: [master]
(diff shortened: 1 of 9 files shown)


dmdashenkov:
@alexander-yevsyukov, PTAL.

alexander-yevsyukov (approved):
LGTM

PR #678, opened 2021-09-09
`Comparable`  messages code generation

dmdashenkov:


tools/comparables-java/src/main/kotlin/io/spine/tools/comparables/ComparablesJavaRenderer.kt
```diff
@@ -0,0 +1,160 @@
...
+    private fun comparison(selfName: PoetClassName, type: ComparableType): List<String> {
+        val other = Literal("other")
+        val naturalOrder = Literal("naturalOrder")
+        val body = CodeBlock
+            .builder()
+            .addStatement("\$T.requireNonNull(\$L)", Objects::class.java, other)
+            .addStatement("return \$L.compare(this, \$L)", naturalOrder, other)
+            .build()
+        val method = MethodSpec
+            .methodBuilder("compareTo")
```
alexander-yevsyukov, line 100: Can we have `methodBuilder()` here and `builder()` above coming on the previous line?

Please also consider adding `methodBuilder()` to the list of builder methods in IDEA settings.

tools/comparables/src/main/proto/spine/tools/comparable_type.proto
```diff
@@ -0,0 +1,33 @@
...
+
+option (type_url_prefix) = "type.spine.io";
+option java_package = "io.spine.tools.comparables";
+option java_outer_classname = "ComparableTypeProto";
+option java_multiple_files = true;
+
+import "spine/base/field_path.proto";
+import "spine/protodata/ast.proto";
+
+// A view on a  message type that can be ordered.
```
alexander-yevsyukov, line 15: Please remove extra space after `a`.

dmdashenkov:
The build in `base` is complicated enough. Applying this change brings even more complexity. For example, right now, there are two different errors on two different OSes.

I'm moving these changes to a [new repository](https://github.com/SpineEventEngine/code-generation/pull/1) — `code-generation`. The Comparables will return in a PR that applies the `code-generation` to `base`.

PR #680, opened 2021-09-21
Publish base Model Compiler artifact

alexander-yevsyukov:
This PR renames the base Spine Model Compiler Gradle plugin to be `model-compiler` rather than just `mc`, and fixes its publishing.


yuri-sergiichuk (approved):
@alexander-yevsyukov LGTM.

A question though: which one should we use in the projects that are built on top of Spine v2?

PR #687, opened 2021-10-20
I/O package improvements

alexander-yevsyukov:
This PR:
  * Splits `Files2` utilities into those named after corresponding activities (i.e. `Copy`, `Ensure`, Delete`).
  * Removes `DirectoryCleaner` in favor of recursive directory deletion used in other parts of the code.
  * Updates to latest `config`, hence the number of change files.

### DIFF
--- a/base/src/main/java/io/spine/io/Ensure.java
+++ b/base/src/main/java/io/spine/io/Ensure.java
@@ -66,25 +66,23 @@ public final class Ensure {
      *
      * @param file
      *         a file to check
-     * @return {@code true} if the file did not exist and was successfully created;
-     *         {@code false} if the file already existed
+     * @return the given instance
      * @throws IllegalArgumentException
      *         if the given file is a directory
      * @throws IllegalStateException
-     *         in case of any I/O exceptions that may occur while creating the parent directories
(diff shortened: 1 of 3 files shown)


base/src/main/java/io/spine/io/Copy.java
```diff
@@ -0,0 +1,119 @@
...
+public final class Copy {
+
+    /** Prevents instantiation of this utility class. */
+    private Copy() {
+    }
+
+    /**
+     * Copies a whole directory and its contents into another directory.
+     *
+     * <p>Both paths must point to an existing directory.
```
armiol, line 56: "to existing directories".

base/src/main/java/io/spine/io/Copy.java
```diff
@@ -0,0 +1,119 @@
...
+     *         the new parent directory
+     */
+    public static void copyDir(Path dir, Path target) throws IOException {
+        copyDir(dir, target, path -> true);
+    }
+
+    /**
+     * Copies the directory and its contents matching the passed predicate into another directory.
+     *
+     * <p>Both paths must point to an existing directory.
```
armiol, line 75: Same as above.

PR #695, opened 2021-11-09
Document validation error message format in `options.proto`

dmdashenkov:
As a part of the validation library renovation, we change the format of error messages presented with the constraint violations.

This means replacing the `%s` placeholders with case-specific tokens such as `{value}` and `{other}`.

We also deprecate the old `msg_format` fields and replace them with `error_msg` in order not to confuse the users with the word "format".

### DIFF
--- a/base/src/main/java/io/spine/validate/option/GoesConstraint.java
+++ b/base/src/main/java/io/spine/validate/option/GoesConstraint.java
@@ -47,8 +47,7 @@ public final class GoesConstraint extends FieldConstraint<GoesOption> {
     }
 
     @Override
-    @SuppressWarnings("deprecation")
-        // The old validation will not migrate to the new error messages.
+    @SuppressWarnings("deprecation") /* Old validation won't migrate to the new error messages. */
     public String errorMessage(FieldContext field) {
         GoesOption option = optionValue();
         return ViolationText.errorMessage(option, option.getMsgFormat());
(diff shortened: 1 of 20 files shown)


dmdashenkov:
@armiol, PTAL.

base/src/main/java/io/spine/validate/option/GoesConstraint.java
```diff
@@ -47,6 +47,8 @@
     }
 
     @Override
+    @SuppressWarnings("deprecation")
```
armiol, line 50: Can we make it fit a single line?
```
@SuppressWarnings("deprecation") /* Old validation won't migrate to the new error messages. */
```

base/src/main/proto/spine/options.proto
```diff
@@ -548,7 +553,13 @@ message IfMissingOption {
...
 
     // A user-defined validation error format message.
-    string msg_format = 1;
+    //
+    // Use `error_msg` instead.
+    //
+    string msg_format = 1 [deprecated = true];
+
+    // A user-defined error message.
+    string error_msg = 2;
```
armiol, line 562: Please say something about the available placeholders (i.e. `other`, `value`) which you mention in the example below.

armiol (approved):
@dmdashenkov LGTM with a few comments to address.

base/src/main/proto/spine/options.proto
dmdashenkov: We mention the placeholders where it makes sense. There is, however, no reason to add those tokens to `IfInvalid` and `IfMissing` error messages.

PR #696, opened 2021-11-23
Deprecate `FsObject.directory()` method

alexander-yevsyukov:
This PR:
  * Deprecates the `FsObject.directory()` method which turned to have a confusing name. The `parent()` method was introduced instead.
  * Removes redundant settings in the build scripts.
  * Bumps `config`.

### DIFF
--- a/base/src/main/java/io/spine/code/fs/FsObject.java
+++ b/base/src/main/java/io/spine/code/fs/FsObject.java
@@ -26,6 +26,9 @@
 
 package io.spine.code.fs;
 
+import com.google.errorprone.annotations.InlineMe;
+import org.checkerframework.checker.nullness.qual.Nullable;
+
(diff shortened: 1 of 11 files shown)


base/src/main/java/io/spine/code/fs/FsObject.java
```diff
@@ -52,11 +56,21 @@ public final Path path() {
 
     /**
      * Obtains the directory to which this object belongs.
+     *
+     * @deprecated please use {@link #parent()}.
      */
-    public final Path directory() {
-        return checkNotNull(path.getParent());
+    @Deprecated
+    @InlineMe(replacement = "this.parent()")
```
dmdashenkov, line 63: Cool, now we have one for Java too!

dmdashenkov (approved):
LGTM.

PR #697, opened 2021-11-26
Resource directory and its copying

alexander-yevsyukov:
This PR improves the way we copy directories and resources. In particular:
  * In addition to methods that copy a directory, `Copy` utilities are enhanced with `copyContent()` methods that copy the content of the directory (so that the directory itself is not copied, only what's inside the directory).
  * `ResourceDirectory` class is introduced. It allows to copy a directory under resources to a file system.
 
Minor changes:
  * Some warnings associated with Java 11 compiler were addressed.
  * Tests associated with the code changed in this PR were migrated to Kotlin.


alexander-yevsyukov:
An interesting issue were with Windows `URL` to `Path` conversion. Here's [what I found at StackOverflow](https://stackoverflow.com/questions/6164448/convert-url-to-normal-windows-filename-java#:~:text=The%20current%20recommendation%20(with%20JDK%201.7%2B)%20is%20to%20convert%20URL%20%E2%86%92%20URI%20%E2%86%92%20Path.%20So%20to%20convert%20a%20URL%20to%20File%2C%20you%20would%20say%20Paths.get(url.toURI()).toFile().%20If%20you%20can%E2%80%99t%20use%20JDK%201.7%20yet%2C%20I%20would%20recommend%20new%20File(URI.getSchemeSpecificPart())).

base/src/main/java/io/spine/io/ResourceObject.java
```diff
@@ -0,0 +1,114 @@
...
+
+import java.io.IOException;
+import java.net.URL;
+import java.util.Enumeration;
+
+import static com.google.common.base.Preconditions.checkNotNull;
+import static io.spine.util.Exceptions.newIllegalStateException;
+import static java.lang.String.format;
+
+abstract class ResourceObject {
```
armiol, line 39: Please document this type.

armiol (approved):
@alexander-yevsyukov LGTM except for one file which is missing documentation.

PR #698, opened 2021-12-08
Migrate to JDK 11 and Java 11 syntax [part 1]

armiol:
This changeset is the first part of the migration to Java 11.

* switches the project to JDK 11 and Java 11 in source/target level for Java files;
* uses `var`s instead of explicit types — this is done for the part of the source files;
* removes the Error Prone plugin from the "unit" test sources — in order to avoid the fatal compiler error;
* removes the unused code.

In scope of this changeset, a lot of "unused" endpoints were discovered. See #699 on this matter.

The library version is set to `2.0.0-SNAPSHOT.78`.


buildSrc/src/main/kotlin/io/spine/internal/gradle/javascript/JsContext.kt
```diff
@@ -0,0 +1,61 @@
...
+    /**
+     * Executes `npm` command in a separate process.
+     *
+     * This [File] is used as a working directory.
+     */
+    fun File.npm(vararg args: String) = project.exec {
+
+        workingDir(this@npm)
+        commandLine(npmExecutable)
+        args(*args)
```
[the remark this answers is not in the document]
armiol: @sonatype-lift ignore

base/src/main/java/io/spine/reflect/PackageGraph.java
```diff
@@ -116,39 +115,36 @@ public static PackageGraph containing(String packagePrefix) {
...
-            Optional<PackageInfo> directParent =
-                    graph.nodes()
-                         .stream()
-                         .filter((node) -> IsDirectParent.of(current)
-                                                         .test(node.getValue()))
-                         .findFirst();
-            PackageInfo newNode = PackageInfo.of(current);
+            final var current = first;
+            var directParent = graph.nodes().stream()
+                    .filter((node) -> IsDirectParent.of(current)
```
dmdashenkov, line 144: Please extract `IsDirectParent.of(current)` from the lambda. Also, there is something strange going on with the renaming `first` to `current`. Maybe, we could get rid of that as well.

alexander-yevsyukov (approved):
LGTM

base/src/main/java/io/spine/string/MapQuoter.java
```diff
@@ -45,11 +44,11 @@
...
         checkNotNull(stringToQuote);
-        Matcher matcher = compile(QUOTE_PATTERN).matcher(stringToQuote);
-        String unslashed = matcher.find() ?
-                           matcher.replaceAll(BACKSLASH + matcher.group()) :
-                           stringToQuote;
-        String result = QUOTE_CHAR + unslashed + QUOTE_CHAR;
+        var matcher = compile(QUOTE_PATTERN).matcher(stringToQuote);
+        var unslashed = matcher.find() ?
+                        matcher.replaceAll(BACKSLASH + matcher.group()) :
+                        stringToQuote;
```
dmdashenkov, line 50: Don't we usually put the `?` and the `:` on the following lines?
I remember that because it does not work in Groovy, we had to use the other format (the one used in this snippet).

dmdashenkov (approved):
LGTM with a couple of comments.
Also, what did you do to make GitHub show `buildSrc/` changes at the end rather than at the start? This makes browsing the files in GitHub waaay easier.

PR #703, opened 2021-12-29
Add `Glob` data class

alexander-yevsyukov:
This PR adds support for glob patterns, adding the `Glob` data class the code of which was initially introduced in `ProtoData`.

This PR improves the original code by:
  * Making it `data class`.
  * Allowing leading dots when creating instances for filtering files by extensions.
  * Disallowing instances with empty patterns.

Latest `config` was also applied.

### DIFF
--- a/base/src/main/kotlin/io/spine/io/Glob.kt
+++ b/base/src/main/kotlin/io/spine/io/Glob.kt
@@ -48,7 +48,7 @@ public data class Glob(val pattern: String) {
     public companion object {
 
         /**
-         * A pattern which matched any file.
+         * A pattern which matches any file.
          */
         public val any: Glob = Glob("**")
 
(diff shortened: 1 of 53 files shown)


base/src/main/kotlin/io/spine/io/Glob.kt
```diff
@@ -0,0 +1,73 @@
...
+        require(pattern.isNotEmpty())
+    }
+
+    private val matcher: PathMatcher = FileSystems.getDefault()
+        .getPathMatcher("glob:$pattern")
+
+    public companion object {
+
+        /**
+         * A pattern which matched any file.
```
armiol, line 51: I think, it's "matches".

armiol (approved):
@alexander-yevsyukov LGTM with a comment.

PR #707, opened 2022-01-10
Allow more than one extension per `Glob`

alexander-yevsyukov:
This PR extends the `Glob` class to allow having more than one file extension per instance (using the `**.{ext1,ext2}` syntax. 

Also the class got the `extensionLowerAndUpper()` method which allows having both lower- and uppercase versions of extensions for handling unusual cases of seasoned source code files coming with uppercase names.

### DIFF
--- a/base/src/main/kotlin/io/spine/io/Glob.kt
+++ b/base/src/main/kotlin/io/spine/io/Glob.kt
@@ -106,8 +106,8 @@ public data class Glob(val pattern: String) {
             if (!iterator().hasNext()) {
                 return listOf()
             }
-            val values: MutableSet<String> = mutableSetOf()
-            val result: MutableList<String> = mutableListOf()
+            val values = mutableSetOf<String>()
+            val result = mutableListOf<String>()
             for (seq in this) {
                 val dotless = seq.withoutLeadingDot()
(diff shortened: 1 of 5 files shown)


base/src/main/kotlin/io/spine/io/Glob.kt
```diff
@@ -53,21 +57,78 @@ public data class Glob(val pattern: String) {
...
+         * @param allowUpperCase
+         *         if `true` each entry of the returned list would have lower- and uppercase
+         *         version of the sequence. Otherwise, the sequences would be used as is.
+         */
+        private fun Iterable<CharSequence>.withCaseOptions(allowUpperCase: Boolean): List<String> {
+            if (!iterator().hasNext()) {
+                return listOf()
+            }
+            val values: MutableSet<String> = mutableSetOf()
+            val result: MutableList<String> = mutableListOf()
```
dmdashenkov, line 110:
```suggestion
            val values = mutableSetOf<String>()
            val result = mutableListOf<String>()
```

dmdashenkov (approved):
LGTM with a comment.

armiol (approved):

PR #709, opened 2022-01-17
Demote debug level for diag. messages when switching `Environment` type

alexander-yevsyukov:
It used to be `INFO`, but it's not very informative and is a bit noisy. This PR demotes it to the `_debug()` level, which is just `FINE`.


armiol (approved):
LGTM.

PR #717, opened 2022-04-23
Bump Kotlin -> `1.6.21`

alexander-yevsyukov:
This PR applies latest `config` and bumps some of the dependencies (including Kotlin to latest `1.6.21`), which required some tuning to Gradle build files.

### The problem
The versions of `kotlin-gradle-plugin` after `1.6.10` (i.e. `1.6.20` and `1.6.21` applied by this PR) somehow differ from `1.6.10` in a way that affect the way tasks `generateTestProto` and `processTestResources` used to work. This may deal with the parallel processing introduced in the plugin in `1.6.20`. Anyway, the change resulted in not copying the file `build/descriptors/test/known_types_test.desc` into program resources. That caused tests that deal with proto types fail because types defined under `test/proto/` were not loaded to `KnownTypes`.  

This problem did not appear for the `main` proto types, though, which is a bit strange. Another strange thing is that running tests from IDEA (even when configured to use Gradle for running tests) worked OK. Only console execution failed 35 tests under the `base` module.

### The solution
1. Make `processResources` tasks explicitly depend on `generateProto` tasks.
2. Extract the configuration of `generateProto` tasks into an extension function `GenerateProtoTask.setup` (see `ProtoTaskExtensions.kt` for details).
3. Use the task setup in the build scripts of the modules like this:

```kotlin
protobuf {
    generateProtoTasks {
        for (task in all()) {
            task.setup(generatedDir)
        }
    }
}
```

It should be said that trying to do this under the `submodules` section of the parent project did not work, resulting in the failure of those 35 tests.

### Other changes
 * Make `testlib` module use the same “trick” for proto tasks. Even though this module does not have production proto types, it does have such types for tests. So, applying the same setup makes sense and eliminates the console warning too.
 * Force configurations for Kotlin dependencies under `buildSrc/build.gradle.kts`. It still may result in warnings in the console. It is expected that Gradle 7.5 should fix this issue.
 * Use `-opt-in` instead of `-Xopt-in`, addressing the Kotlin compiler warning.
 * Bump Protobuf -> `3.20.1`.

Changes to dependencies and scripts introduced in this PR will be applied to `config` shortly.


armiol (approved):
LGTM.

PR #720, opened 2022-08-21
Extend definition of `internal` and `internal_all` options

alexander-yevsyukov:
This PR updates the documentation for `internal` and `internal_all` options to reflect the extended usage scenario.

Until this PR these options were meant to identify public types that are not meant to be used by the users of the framework. 

In the light of the request to have signals “private” to a bounded context it was decided to extend the meaning of the already existing options. For more information on this decision please see this [design draft document](https://docs.google.com/document/d/1IBlnIBzeVft7btrYD61h4HckD86aDRRTIv4AZ3wOd0s/edit?pli=1#).

After this PR, being `internal` for a message type would also mean that it cannot be sent to a bounded context, which marks its as `internal`. Also it would mean that such bounded context cannot send it outside.

These limitations would be implemented after this PR is complete.

Other changes:
  * Gradle bumped to 7.5.1.
  * Latest `config` was applied.
  * Nasty deprecation warnings on calling `isAccessible()` were addressed.


armiol (approved):
@alexander-yevsyukov LGTM.

PR #721, opened 2022-08-24
More on unpulished language

alexander-yevsyukov:
This PR continues extending usage of `internal_type` proto option and `@Internal` Java annotation for marking non-published language of a bounded context.

* Documentation of the `@Internal` annotation was updated accordingly.
* The retention policy of the annotation was changed from `SOURCE` to `RUNTIME` because we're going to check it then.
* `UnpublishedLanguageException` was introduced to be thrown when the user attempts to use a non-published message outside of the context.
* Several Kotlin-based extensions for `Message` types and utility functions related to published language were introduced.
* `StubMessage` class was introduced to simplify creation of stub message types in tests.

### DIFF
--- a/base/src/main/java/io/spine/annotation/Internal.java
+++ b/base/src/main/java/io/spine/annotation/Internal.java
@@ -65,7 +65,7 @@ import java.lang.annotation.Target;
  * @see SPI
  */
 @Internal
-@Retention(RetentionPolicy.SOURCE)
+@Retention(RetentionPolicy.RUNTIME)
 @Target({
         ElementType.ANNOTATION_TYPE,
         ElementType.CONSTRUCTOR,
--- a/base/src/main/java/io/spine/annotation/Internal.java
+++ b/base/src/main/java/io/spine/annotation/Internal.java
@@ -34,7 +34,8 @@ import java.lang.annotation.Target;
 
 /**
  * Annotates a program element (class, method, package, etc.) which is not a part of a public API,
- * and thus should not be used.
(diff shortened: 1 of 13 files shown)


base/src/main/java/io/spine/annotation/Internal.java
```diff
@@ -33,8 +33,28 @@
...
+ * not be used programmers using the framework.
+ *
+ * <p>If the annotation is used for a type it may mean one of the following:
+ * <ol>
+ *     <li>This type is internal to the Spine Event Engine framework.
+ *
+ *     <li><p>The type is internal to a bounded context, artifact of which exposes the type to
+ *     the outside world (presumably for historical reasons).</p>
+ *
+ *     <p>The type with this annotation can be used only inside the bounded context
```
armiol, line 51: I think some of these tags should be list items, not paragraphs. At least, they are formatted in this way. 

Please review.

armiol (approved):
@alexander-yevsyukov LGTM except for a single comment. Please have a look.

base/src/main/java/io/spine/annotation/Internal.java
alexander-yevsyukov: @armiol, how about now?

armiol (approved):
LGTM.

base/src/main/java/io/spine/annotation/Internal.java
armiol: Looks more consistent to me. Thanks!

PR #722, opened 2022-08-28
Fix Javadoc and tests for `GenericTypeIndex`

alexander-yevsyukov:
This PR:
  * Fixes Javadoc of `GenericTypeIndex` which previously misused the word `extends` where `implements` should be said.
  * Extends the Javadoc with the code example of using enum members for obtaining generic type arguments.
  * Fixes the placement of `Types` utility class.
  * Adds test for enum-based type access.


armiol (approved):
@alexander-yevsyukov LGTM.

PR #723, opened 2022-08-29
Convert collections to `Any`

armiol:
This changeset introduces an internal tool, which converts `Map`s and `List`s to `Any`.

Such an addition is required to pack the instances of these collections into `ValidationError`, when validating a `Message`. Previously, a similar functionality was included into `validation-runtime` library. However, the code generated by the validation library already has the dependency onto `base` (due to an extensive usage of Spine Proto options and utilities such as `AnyPacker`), so it's nice to have less dependencies from the validation-generated codebase.

Similar functionality will be removed from the `validation` library in the following PRs.


armiol:
@alexander-yevsyukov PTAL.

alexander-yevsyukov (approved):
LGTM

PR #724, opened 2022-09-16
Open a constructor of `UnpublishedLanguageException` which accepts a type name

alexander-yevsyukov:
This PR makes the constructor of `UnpublishedLanguageException` accepting a type name `public` so that users of the class don't have to create default instances of a message (knowing the type) just to create an exception.


armiol (approved):
LGTM.

PR #726, opened 2022-09-23
Fix validation check

alexander-yevsyukov:
This PR restores the `Validate.checkValid()` method returning `void` (to fix the broken API) and deprecates it in favor of the one returning the passed message.


armiol (approved):
@alexander-yevsyukov LGTM.

PR #727, opened 2022-09-23
Extract validation step 1

alexander-yevsyukov:
This PR extract the code of `io.spine.validate` package into a separate `base-validate` submodule and artifact.

I order to perform the separation and keep the same level of functionality, an `OptionsProvider` interface was "borrowed" from ProtoData, implementing the population of an `ExtensionRegistry` via the Service Loader mechanism.

Another [pending change](https://github.com/SpineEventEngine/tool-base/pull/24) is modification of `MoreKnownTypes` class with loading `ExternalConstraints` when more types are added. Previously it was done by `KnownTypes.extendWith()` method (which is a part of `base`), which is no longer possible because of separation of `validate` module (which now depends on `base`).


base/src/main/proto/spine/base/error.proto
```diff
@@ -50,10 +49,13 @@ message Error {
...
     int32 code = 2;
 
-    // An error indicating that a message did not pass validation.
-    validate.ValidationError validation_error = 3;
+    reserved "validation_error";
+    reserved 3; // It used to be the index of the deprecated `validation_error` field.
 
-    reserved 4 to 9;
+    // The message of the error wrapped into `Any`, if available.
+    google.protobuf.Any error = 4;
```
dmdashenkov, line 56: How about `details`? I'd avoid the duplication of `Error.error`.

base/src/main/proto/spine/base/error.proto
```diff
@@ -50,10 +49,13 @@ message Error {
     // An integer code resulting from a native code call, or a number from a Proto enum value.
     int32 code = 2;
 
-    // An error indicating that a message did not pass validation.
-    validate.ValidationError validation_error = 3;
```
dmdashenkov, line 54: Let's start by deprecating the field yet leaving it, so that we don't break any existing code. 

Please add the `[deprecated = true]` field option.

base/src/main/java/io/spine/base/ErrorMessage.java
```diff
@@ -0,0 +1,44 @@
...
+import com.google.protobuf.Message;
+
+/**
+ * Common interface for exceptions that can convert to a message instance for
+ * later inclusion into an {@link Error}.
+ *
+ * @param <M>
+ *         the type of the error message provided by the exception
+ */
+public interface ErrorMessage<M extends Message> {
```
dmdashenkov, line 38: I suggest `ErrorWithMessage` and `ErrorMessage.asMessage()`.

An error message sounds like a text message of an exception.

dmdashenkov (commented):
LGTM with a couple of suggestions.

base/src/main/java/io/spine/option/BaseOptionsProvider.java
```diff
@@ -0,0 +1,46 @@
...
+
+import com.google.auto.service.AutoService;
+import com.google.protobuf.ExtensionRegistry;
+
+/**
+ * Register options generated for the {@code spine/options.proto} file.
+ */
+@SuppressWarnings("unused") // loaded by AutoService.
+@AutoService(OptionsProvider.class)
+public class BaseOptionsProvider implements OptionsProvider {
```
armiol, line 37: Looks `final` to me.

validate/src/main/java/io/spine/validate/option/RangedConstraint.java
```diff
@@ -82,8 +82,8 @@ private static void verifyType(FieldDeclaration field, Range<ComparableNumber> r
     }
 
     private static void checkInteger(ComparableNumber number, FieldDeclaration field) {
-        checkState(number.isInteger(),
-                   "An integer bound expected for field `%s`, but got `%s`.", field, number);
+        Preconditions.checkState(number.isInteger(),
```
armiol, line 85: Please see if we could keep this static import.

I think this might be IDEA "taking care" of us after the file movement. And it looks like it may not be just here :(

validate/src/main/java/io/spine/validate/option/ValidatingOptionsProvider.java
```diff
@@ -0,0 +1,44 @@
...
+
+package io.spine.validate.option;
+
+import com.google.auto.service.AutoService;
+import com.google.protobuf.ExtensionRegistry;
+import io.spine.option.OptionsProvider;
+
+@SuppressWarnings("unused") // loaded by AutoService
+@AutoService(OptionsProvider.class)
+public class ValidatingOptionsProvider implements OptionsProvider {
```
armiol, line 35: Again, looks `final`.

armiol (changes requested):
@alexander-yevsyukov please see my comments in addition to those left by @dmdashenkov.

base/src/main/proto/spine/base/error.proto
alexander-yevsyukov: This change was done to allow moving `ValidationError` into another submodule. If we make it just deprecated, we won't be able to move the type.

armiol (approved):

PR #730, opened 2022-09-28
Include `type.proto` descriptor into the distribution

alexander-yevsyukov:
This PR makes the descriptor of `google/protobuf/type.proto` to be included into the distribution.


alexander-yevsyukov:
@armiol, I'll call you about this PR and current anomaly in `core-java` that relates to it.

armiol (approved):
LGTM.

However, let's discuss this in person later.

Issue #729, opened 2022-09-27
Make the `Tests` environment code reusable


dmytro-kashcheiev:
Currently, it's impossible to reuse code from the `Tests` environment. 

Use case:
I want to have several custom environments for example `IntegrationTests` and `UnitTests`. Both of these environments in addition to their own code will perform checks identical to the checks in the `Tests` environment.

Issue #728, opened 2022-09-27
Add the possibility to disable default `Production` and `Tests` environments


dmytro-kashcheiev:
Usecase: 
Several custom environments. I want to fail fast when none of _MY_ custom environments is suitable for the current runtime. In the current Spine implementation, I will get the default `Production` or `Tests` environment, which I would not like.

- Add the possibility to disable default `Production` and `Tests` environments.
- In addition, it would be great to have a custom exception handler for scenarios when non of the environments match the current runtime. So I can send error notifications via some external system to inform developers about misconfiguration.

alexander-yevsyukov:
@dmytro-kashcheiev, how about registering a `CustomEnvironmentType` (before all the other) that would do the job of the exception handling and then throw an exception, so that the execution would stop?

dmytro-kashcheiev:
@alexander-yevsyukov 

Currently, all `CustomEnvironmentTypes` are registered before the default one. 

For example, in my current case, I use the environment variables to determine if this is production or staging. If the environment variable is "some-random-value" I want Environment API to throw an exception. 

I don't want to cover ALL possible values and have something like a `default` environment. I used to have a `default` environment before. it was `LocalEnvironment` but when I misconfigured production I get `Local` version of the app instead of the `Production` one, this was confusing.

With the current API if I have something like: 
```
.register(MyProduction) // For `some-random-value` this will return `enabled=false`. 
.register(MyStaging) // For `some-random-value` this will return `enabled=false`.
```
The Spine will consider such a situation as `io.spine.base.Production`. 


I can achieve the required fail-fast behavior with the current API. I should do the following: 
```
.register(WrongEnvironment) // And here we should inform developer about missconfiguration. 
.register(MyProduction) // For `some-random-value` this will return `false`. 
.register(MyStaging) // For `some-random-value` this will return `false`.
```
But this looks like a hack.

alexander-yevsyukov:
Resolved by #733.

PR #733, opened 2022-09-28
Better environments

alexander-yevsyukov:
This PR improves working with custom environments:
  * `Environment.autoDetect()` method was added allowing to clear currently detected type. This way we don't have to `reset()` and refill the `Environment` with custom types for later auto-detection when the environment changes.
  * `Tests.type()` and `Tests.enabled()` were made public (addressing #729).
  * `Production` environment type was renamed to `DefaultMode` to better reflect the reality of this case.
  * `Environment` was extended with the ability to configure callbacks when a corresponding environment type is detected.
  * Spek and Kotest were added to the list of known testing frameworks.

By adding the detection callback, this PR resolves the root cause of #728.


alexander-yevsyukov:
@dmytro-kashcheiev, please have a look. Also, please see my comment in #728 as to disabling standard types.

@armiol, FYI and review.

armiol:
@alexander-yevsyukov let's discuss vocally later today.

alexander-yevsyukov:
[On the new name](https://www.ted.com/talks/manoush_zomorodi_how_boredom_can_lead_to_your_most_brilliant_ideas) for the `Production` type.

alexander-yevsyukov:
@armiol, @dmytro-kashcheiev, PTAL.

base/src/main/java/io/spine/environment/EnvironmentType.java
```diff
@@ -45,6 +53,28 @@ public abstract class EnvironmentType {
      */
     protected abstract boolean enabled();
 
+    /**
+     * Installs the callback to be called when this environment type is
+     * {@linkplain #enabled() detected.}
+     */
+    final void onDetected(@Nullable Consumer<T> callback) {
```
armiol, line 60: Looks like this one should be `public`. Otherwise it is inaccessible for the framework users.

armiol (changes requested):
@alexander-yevsyukov LGTM except for a single comment.

base/src/main/java/io/spine/environment/EnvironmentType.java
alexander-yevsyukov: This method is intended for the framework users:
```java
public <T extends EnvironmentType<T>>
    void whenDetected(Class<T> cls, @Nullable Consumer<T> callback)
```
An instance method would require having an instance, which is the opposite to the "protocol" we have with working with `Environment`: we work via `Class` references instead.

armiol (approved):
LGTM. 

Sorry, I've missed that public piece.

PR #744, opened 2022-10-07
`Text` utility class

alexander-yevsyukov:
This PR adds the utility class `Text` for working with text lines and migrates `KnownTypes` to it.

An instance of `Text` represents immutable list of text lines separated by `System.lineSeparator()`. It is meant to replace usage of `ImmutableList<String>` in our code.


base/src/main/java/io/spine/util/Lines.java
```diff
@@ -0,0 +1,87 @@
...
+    public static String join(String[] lines) {
+        checkNotNull(lines);
+        return joiner().join(lines);
+    }
+
+    /**
+     * Obtains the {@link Joiner} on {@link System#lineSeparator()}.
+     */
+    public static Joiner joiner() {
+        return Joiner.on(lineSeparator());
```
armiol, line 78: Please consider caching some bits. `joiner` is stateful, so it's not a candidate. But `lineSeparator()` looks like one.

armiol (approved):
LGTM. Please see my comment on performance. Given the potentially big number of invocations, it seems important.

alexander-yevsyukov:
@armiol, having more look at what we do with test, I think we need `Text` utility class which does a bit more than what `Lines` do. I'll work more on this PR to introduce the utility container.

base/src/main/java/io/spine/util/Text.java
```diff
@@ -0,0 +1,206 @@
...
+        return Objects.equals(lines, strings.lines);
+    }
+
+    @Override
+    public int hashCode() {
+        return lines.hashCode();
+    }
+
+    /*
+     * Static utilities. Please prefer instance methods of {@code Text}.
```
armiol, line 160: Please format the section according to Spine guidelines. 

Or move this recommendation to some other place, which will be reflected in the rendered Javadocs.

armiol (approved):
@alexander-yevsyukov LGTM, but please see my comment prior to merging.

PR #746, opened 2022-10-08
Remember `Text` value

alexander-yevsyukov:
This PR teaches the `Text` class to remember its value so that users do not have to merge lines to work with the text as a whole thing.

### DIFF
--- a/base/src/test/kotlin/io/spine/util/TextTest.kt
+++ b/base/src/test/kotlin/io/spine/util/TextTest.kt
@@ -75,4 +75,12 @@ internal class TextTest {
         assertThrows<IllegalArgumentException> {  Text.of("un", "${nl}o") }
         assertThrows<IllegalArgumentException> {  Text(listOf("dos", "tres${nl}")) }
     }
+
+    @Test
+    fun `always return the same value`() {
+        val text = Text.of("donna", "be", "la", "mare")
+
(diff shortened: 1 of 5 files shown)


base/src/test/kotlin/io/spine/util/TextTest.kt
```diff
@@ -75,4 +75,12 @@ internal class TextTest {
         assertThrows<IllegalArgumentException> {  Text.of("un", "${nl}o") }
         assertThrows<IllegalArgumentException> {  Text(listOf("dos", "tres${nl}")) }
     }
+
+    @Test
+    fun `always return the same value`() {
+        val text = Text.of("donna", "be", "la", "mare")
+
+        val value = text.value()
+        assertThat(value).isSameInstanceAs(text.value())
```
armiol, line 84: I think this test will pass even if we return a new instance every time, since `String` itself has a pool of re-usable instances.

armiol (approved):
LGTM. See my comment, although I don't suggest anything, just commenting.

base/src/test/kotlin/io/spine/util/TextTest.kt
alexander-yevsyukov: Then let's keep it just a requirement. Don't want to involve mocking just for this.

PR #753, opened 2022-10-26
More Kotlin extensions

alexander-yevsyukov:
This PR:
  * Adds `VIEW` alias for `PROJECTION` in the `Enum.Kind` enumeration. This is to allow writing `option (entity.kind)=VIEW` which plays nicer with `View` types in ProtoData (which is hopefully would be added to `core-java` soon).
  * Moves general purpose Kotlin extensions from ProtoData.

### DIFF
--- a/base/src/main/proto/spine/options.proto
+++ b/base/src/main/proto/spine/options.proto
@@ -825,19 +825,25 @@ message EntityOption {
 
     // A type of an entity for state of which the message is defined.
     enum Kind {
+        option allow_alias = true;
+        
         // Reserved for errors.
         KIND_UNKNOWN = 0;
 
         // The message is an aggregate state.
         AGGREGATE = 1;
 
-        // The message is a state of a projection.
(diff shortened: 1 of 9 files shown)


alexander-yevsyukov:
@armiol, if you can have a look at this PR soon, I'd be able to use this version in that PR in ProtoData already.

base/src/main/proto/spine/options.proto
```diff
@@ -825,19 +825,25 @@ message EntityOption {
...
 
         // The message is an aggregate state.
         AGGREGATE = 1;
 
-        // The message is a state of a projection.
+        // The message is a state of a projection (same as "view").
         PROJECTION = 2;
 
+        // The message is a state of a view (same as "projection").
+        VIEW = 2;
```
armiol, line 840: This is an interesting feature! Didn't know about it.

armiol (approved):
LGTM.

PR #754, opened 2022-10-26
Perform Kotlin code analysis during the build

yevhenii-nadtochii:
This PR moves Detekt code analyzing from GitHub workflow to Gradle `detekt` task. This task is bound to `check` by default, thus it will be executed with `./gradlew build`.

It addresses [config #412](https://github.com/SpineEventEngine/config/issues/412).

In particular, this changeset does the following:

1. Drops `detekt-analysis` GitHub workflow.
2. Pulls the latest `config` with implemented [`detekt-code-analysis`](https://github.com/SpineEventEngine/config/pull/420) script-plugin.
3. Applies that plugin to `base` and `testlib` modules.

Please note, due to a [bug](https://github.com/gradle/gradle/issues/22532) in Gradle, plugins are applied with string ID instead of referencing to dependency-objects.

Detekt reported two finding during the build:

1. Unused exception in `catch` block. I've renamed it to `ignored`.
2. Too many functions in `TruthExtensions.kt`. I've suppressed this one with the baseline file.


pom.xml
```diff
@@ -169,6 +169,11 @@ all modules and does not describe the project structure per-subproject.
     <artifactId>checkstyle</artifactId>
     <version>10.3.4</version>
   </dependency>
+  <dependency>
+    <groupId>io.gitlab.arturbosch.detekt</groupId>
+    <artifactId>detekt-cli</artifactId>
+    <version>1.21.0</version>
```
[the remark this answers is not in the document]
yevhenii-nadtochii: @sonatype-lift ignore

I see that the latest `1.22.0-RC2` release candidate have a patched version of `snakeyaml`, which is reported here.

yevhenii-nadtochii:
@alexander-yevsyukov @armiol Please take a look.

Detekt is set up here without `buildSrc`.

base/build.gradle.kts
```diff
@@ -56,3 +61,9 @@ protobuf {
 tasks {
     excludeGoogleProtoFromArtifacts()
 }
+
+detekt {
+    buildUponDefaultConfig = true
+    baseline = file("config/quality/detekt-baseline.xml")
```
alexander-yevsyukov, line 67: What happens if these files are not there?

I'm asking so that we can use script plugins (e.g. from `build-logic`).

base/config/quality/detekt-config.yml
```diff
@@ -0,0 +1,9 @@
+naming:
```
alexander-yevsyukov, line 1: This looks like settings that are applicable to all subprojects. So, they should probably go under `./config/quality`. I mean be not under submodule, but under `config` of the project root.

base/build.gradle.kts
alexander-yevsyukov: OK, let's now assume we have `buildSrc` (so that we don't introduce too many changes with `build-logic`).

Can we have these general things done there? Together with checking for the presence of configuration files, please.

pom.xml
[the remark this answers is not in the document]
yevhenii-nadtochii: ignore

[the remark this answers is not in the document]
yevhenii-nadtochii: ignore

yevhenii-nadtochii:
@armiol @alexander-yevsyukov PTAL.

armiol (approved):

buildSrc/src/main/kotlin/io/spine/internal/dependency/Pmd.kt
```diff
@@ -26,7 +26,6 @@
 
 package io.spine.internal.dependency
 
-// https://pmd.github.io/
```
alexander-yevsyukov, line 29: Please restore this. Newer `config` fixes this acceidental removal.

alexander-yevsyukov (approved):
LGTM, with one minor request.

PR #767, opened 2022-11-14
Add `Subject.instanceOf<T>` extension

alexander-yevsyukov:
This PR adds extension fun to `Subject` so that it's possible to write:

```kotlin
assertThat(Time.currentTime()).isInstanceOf<Timestamp>()`
```
... instead of 

```kotlin
assertThat(Time.currentTime()).isInstanceOf(Timestamp::class.java)`
```
The latter syntax is, of course, stays available for Java.


yevhenii-nadtochii (approved):
LGTM.

PR #768, opened 2022-11-18
Kotest assertions and descriptor extensions

alexander-yevsyukov:
This PR:
  * Carries over extensions to `FileDescriptor` introduced in ProtoData.
  * Adds Kotest assertions to the API of `testlib` and employs these assertions in some tests.


base/src/main/kotlin/io/spine/protobuf/DescriptorExtensions.kt
```diff
@@ -0,0 +1,75 @@
...
+@file:JvmName("DescriptorExtensions")
+
+package io.spine.protobuf
+
+import com.google.protobuf.Descriptors.FileDescriptor
+import com.google.protobuf.ExtensionRegistry
+import io.spine.code.java.ClassName
+
+/**
+ * Obtains a [binary][ClassName.binaryName] name of the outer Java class associated with this file.
```
yevhenii-nadtochii, line 36: We can say just `Returns` for properties to hide their `fun` nature. This verb is "lighter" to me. Or even more, just what it is without a verb: `A [binary][ClassName.binaryName] name of the outer Java class associated with this file.`

testlib/build.gradle.kts
```diff
@@ -28,34 +28,39 @@ import com.google.protobuf.gradle.generateProtoTasks
...
-    api(Guava.testLib)
-    api(kotlin("test-junit5"))
+    (Protobuf.libs
+            + JUnit.api
+            + Truth.libs
+            + Guava.testLib
+            + kotlin("test-junit5")
+            + Kotest.assertions).forEach {
+        api(it)
+    }
```
yevhenii-nadtochii, line 54: As an alternative, we could have an extension upon `dependencies` block to say: `api(Protobuf.libs)`.

yevhenii-nadtochii (approved):
LGTM. Kotest assertions look pretty. 

I've left a couple of style suggestions.

testlib/build.gradle.kts
alexander-yevsyukov: The thing here is that we deal with the ”soup” of `String` and `List<String>` here, which is nicely handled by `plus()` operator. 

We can probably extend `api` to handle things, but I would leave it to other PR _and_ when we have more cases like this.

Issue #699, opened 2021-12-09
Review the public API and currently unused types


armiol:
After splitting `base` into several repositories, some of the types and some of the methods became unused. Also, many endpoints are over-exposed — typically, being `public` instead of package-private.

During the migration to Java 11, the "unused" warnings were suppressed with 
```java
@SuppressWarnings("unused") /* Part of the public API. */
```
The over-exposed methods were left intact.

When addressing this issue, one should review the "unused" and over-exposed pieces. And either remove/hide them, or cover with the corresponding tests.

PR #771, opened 2022-12-08
Public API review, bump Protobuf

alexander-yevsyukov:
This PR:
  1. Bumps Protobuf to `v3.21.11`, which finally fixes Kotlin API modifiers.
   
  2. Partially addresses #699, by 
     - moving some of the types to the sub-projects which are the only users of the type. These types are `PrimitiveType`, `EntityIdField`.
     - covering API marked with @SuppressWarning("unused") with tests.
     The package `io.spine.query` was not included into this procedure because this package is a subject for further, more thorough, analysis.
     - Removing the code (e.g. `FieldTypesProto.trimTypeName(FieldDescriptorProto)`, which is not used by any of the subprojects.

#### Other notable changes:
  * `SourceFile` constructor was promoted to `protected` so that subprojects can extend the type too.
  
  * `Durations2.ZERO` was deprecated in favor of `com.google.protobuf.Durations.ZERO`.
  
  * `io.spine.string.Registrar` was deprecated in favor of new methods of `StringifierRegistry` called `register()`. The methods obtain either one `Stringifier` or vararg, which eliminates the need for the `Registrar` class.
  
  * The initialization of `StringifierRegistry` was simplified using the new registration API.
  
  * Added Kotlin extension properties for `Class`: `className`, `nameOfPackage`, `simplyNamed: SimpleClassName`.
  
  * Added Kotlin extension functions for `NullPointerTester` allowing to use generic parameter instead of `T::class.java`.


base/src/test/java/io/spine/io/EnsureSpec.java
```diff
@@ -47,9 +47,9 @@
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 @DisplayName("`Ensure` utilities class should")
-class EnsureTest extends UtilityClassTest<Ensure> {
+class EnsureSpec extends UtilityClassTest<Ensure> {
```
armiol, line 50: Let's not rename all of our Java tests into "...Spec". Otherwise, it should be done for each and every test class in the framework. Since it's not a Kotlin test type, I don't see any reason for such a massive change.

base/src/test/java/io/spine/io/GlobSpec.java
```diff
@@ -37,7 +37,7 @@
 import static com.google.common.truth.Truth.assertWithMessage;
 
 @DisplayName("`Glob` Java API should expose")
-class GlobTest {
+class GlobSpec {
```
armiol, line 40: Same as above.

armiol (changes requested):
@alexander-yevsyukov Mostly LGTM. 

Please see my comments. 

Also, I am hoping for the best resolution for the deleted (moved) files which previously were a part of our public API.

base/src/test/java/io/spine/io/GlobSpec.java
alexander-yevsyukov: Converted both to Kotlin. I see your point, but even in Java the `Spec` suffix makes sense because it makes tests a specification document, thus changing attitude of junior and mid-level developers to this code.

armiol (approved):

alexander-yevsyukov:
> Also, I am hoping for the best resolution for the deleted (moved) files which previously were a part of our public API.

I'm going to address these API changes along with applying new Protobuf and Protobuf Gradle Plugin version. I've updated some of the subprojects already, as you may have seen from pushed commits. But changing `mc-java` is not easy, because it depends on ProtoData, which uses older version of Protobuf Gradle Plugin extension and fails on applying the plugin to the project. You know this kind of things...

PR #772, opened 2022-12-15
Restore Java API test for `Glob`

alexander-yevsyukov:
This PR restores Java-based test for testing the API exposed by the Kotlin `Glob` class to Java. 

The tests were converted to Kotlin by mistake which was revealed when existing Kotlin-based tests (for the Kotlin class) and those converted from Java appeared in the same package. As a result, the Java API tests were restored and the test suite class was renamed to `GlobJavaApiSpec`. The existing Kotlin-based tests were improved to use Kotest assertions where it was easy to do. The test suite `GlobSpec` still uses Truth because of the need for custom assertion messages (which are not yet mastered by the PR author).


armiol (approved):
@alexander-yevsyukov LGTM.

PR #776, opened 2023-01-26
Improve Kotlin extensions for working with types and annotations

alexander-yevsyukov:
This PR:  
  * Migrates Java utility class `Messages` to Kotlin extensions.
  
  * Separates extensions by files using the notation `<TypeName>Exts.kt`.
  
  * Moves `ApiOption` class from `mc-java`, fixing the issues with explicitly and implicitly defined API options.
  
  * The method `ValueHolder.value()` made `public` to avoid overriding just for opening access to the contained value. The class also got `getValue()` method for compatibility with the Kotlin property syntax.
  
  * `Message.isInternal()` now takes into account not only `@Internal` annotation of the class, but also the values of the `(internal_type)` option and the `(internal_all)` in the declaring file.


base/src/main/java/io/spine/value/ValueHolder.java
```diff
@@ -47,8 +47,19 @@ protected ValueHolder(T value) {
...
+     * Returns the stored value.
+     *
+     * <p>Overriding methods may perform additional type conversion, if needed.
+     */
+    public T value() {
+        return getValue();
+    }
+
+    /**
+     * Obtains the value passed to the folder during construction.
```
armiol, line 60: Which "folder"? Is it "holder"?

base/src/test/proto/spine/given/type/fields.proto
```diff
@@ -0,0 +1,50 @@
...
+
+import "spine/options.proto";
+import "google/protobuf/wrappers.proto";
+
+option (type_url_prefix) = "type.spine.io";
+option java_package="io.spine.given.type";
+option java_multiple_files = true;
+
+// A bunch of fields with different API level options and without options.
+message Filderation {
```
armiol, line 38: Given the name, I would also expect enumerations and `oneof`s :)

armiol (approved):
LGTM with some comments to consider.

base/src/test/proto/spine/given/type/fields.proto
alexander-yevsyukov: It's the next level of expansion ;)

PR #777, opened 2023-02-17
Preserving system line separators

alexander-yevsyukov:
This PR:
  * Adds extensions functions for `String` which work similarly to `trimIndent()` and `prependIndent()` from Kotlin, but unlike them preserve system line separators.
  * Add the `Indent` class for working with indentation. The new class is going to replace its Java namesake from `tool-base`.


armiol (approved):
LGTM.

PR #778, opened 2023-02-18
Detection of OS family

alexander-yevsyukov:
This PR brings `OsFamily` enum which was a part of `tool-base` into this code tree. The ability to detect current OS is needed only for tools, but for libraries that may need to deal with system specifics. For example, tests of the Text library already need to be able current OS because line separators are OS-specific.

It is expected that `tool-base` would migrate to `OsFamily` becauses it fixes an issue with detecting Unix (non-macOS) systems.

### DIFF
--- a/base/src/main/java/io/spine/environment/OsFamily.java
+++ b/base/src/main/java/io/spine/environment/OsFamily.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2022, TeamDev. All rights reserved.
+ * Copyright 2023, TeamDev. All rights reserved.
  *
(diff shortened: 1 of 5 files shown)


base/src/main/java/io/spine/environment/OsFamily.java
```diff
@@ -0,0 +1,131 @@
+/*
+ * Copyright 2022, TeamDev. All rights reserved.
```
armiol, line 2: While we still do that, let's use the most recent year number :)

armiol (approved):
LGTM with a minor comment to address.

PR #780, opened 2023-02-26
Bump Protobuf, its plugin and other dependencies

alexander-yevsyukov:
This PR bumps:
  * Protobuf -> [`3.22.0`](https://github.com/protocolbuffers/protobuf/releases/tag/v22.0).  There are many improvements related to code generation for for Java and Kotlin.
  * Protobuf Gradle Plugin -> [`0.9.2`](https://github.com/google/protobuf-gradle-plugin/releases/tag/v0.9.2).
  * Removes `compiler/plugin.proto` from the list of standard files available in descriptors bundled with `spine-base` because `protobuf-java` no longer contains this file in the distribution. Please see the comment in `GoogleTypesTest.kt` for details.

PR #781, opened 2023-02-26
Test overriding of `ValueHolder.value()` method

alexander-yevsyukov:
This PR adds test for `ValueHolder.value()` not being final. This is needed because some descendants in `core-java` override this method for casting return type values.

Also, it fixes the type [reported](https://github.com/SpineEventEngine/base/pull/780#pullrequestreview-1314716328) for the previous PR.


armiol (approved):
LGTM, and thanks for "to be" :)

PR #784, opened 2023-03-10
Remove `com.google` classes

alexander-yevsyukov:
This PR improves the Protobuf code generation arrangement, thus avoiding appearance of `com.google` classes in the distributed binaries that sneaked in after migration to Protobuf Gradle Plugin v0.9.2.

`ProtoTaskExtensions.kt` was reworked to address the changes introduced in Protobuf Gradle Plugin. 
Dependencies between task outputs and inputs were adjusted to avoid duplicated classes.

The directory `$buildDir/generated/source/proto` which is used by the Protobuf Gradle Plugin was cut off the task dependencies in favor of `$projectDir/generated` directory.

The migration to the `module` script plugin similarly to other projects was not successful because it caused `ConcurrentModificationException` when applying the `idea` plugin. Here's a piece of stacktrace related to this problem:

```
Caused by: java.util.ConcurrentModificationException
        at org.gradle.api.internal.file.DefaultSourceDirectorySet.getSourceTrees(DefaultSourceDirectorySet.java:266)
        at org.gradle.api.internal.file.DefaultSourceDirectorySet.getSourceTrees(DefaultSourceDirectorySet.java:269)
        at org.gradle.api.internal.file.DefaultSourceDirectorySet.getSrcDirTrees(DefaultSourceDirectorySet.java:256)
        at org.gradle.api.internal.file.DefaultSourceDirectorySet.getSrcDirs(DefaultSourceDirectorySet.java:114)
        at org.gradle.plugins.ide.idea.IdeaPlugin$23.call(IdeaPlugin.java:390)
        at org.gradle.plugins.ide.idea.IdeaPlugin$23.call(IdeaPlugin.java:386)
        at org.gradle.util.internal.GUtil.uncheckedCall(GUtil.java:437)
```

Further investigations regarding this issue would be conducted when applying the code generation approach from this PR in ProtoData.


armiol (approved):
@alexander-yevsyukov LGTM with some minor fixes discussed vocally.

PR #788, opened 2023-04-17
Prepare separation of `testlib`

alexander-yevsyukov:
This PR eliminates the dependencies of the `testlib` on other modules of the `base` project. This is preparational step for further extraction of Spine Testlib into a dedicated repository.


testlib/src/main/java/io/spine/testing/logging/mute/MutingLoggerTap.java
```diff
@@ -163,7 +163,7 @@ synchronized long streamSize() {
         try {
             stream.flush();
         } catch (IOException e) {
-            throw newIllegalStateException(e, "Error flushing `%s`.", stream);
+            throw Testing.newIllegalStateException(e, "Error flushing `%s`.", stream);
```
armiol, line 166: Let's just throw ISE by hand instead of using this "copy".

testlib/src/main/java/io/spine/testing/Assertions.java
```diff
@@ -77,12 +76,23 @@ public static IllegalStateException assertIllegalState(Executable e) {
...
-     * Asserts that running the passed executable causes {@link UnknownTypeException}.
+     * Asserts that running the passed executable causes {@code io.spine.type.UnknownTypeException}.
+     *
+     * @implNote The exception class belongs to the {@code io.spine.base:spine-base} artifact.
+     *           In order to avoid the dependency on {@code spine-base} from {@code spine-testlib}
+     *           We obtain the class of the exception via its name.
      */
     @CanIgnoreReturnValue
-    public static UnknownTypeException assertUnknownType(Executable e) {
+    public static Throwable assertUnknownType(Executable e) {
```
armiol, line 86: In SpineEventEngine organisation, there's just [as much usage scenarios](https://github.com/search?q=org%3ASpineEventEngine%20assertUnknownType&type=code), and all of them are in `base`. 

I suggest we just leave this method in `base`, perhaps, as a Kotlin extension.

testlib/src/main/java/io/spine/testing/Testing.java
```diff
@@ -82,4 +97,125 @@ public static void repeat(int count, Runnable action) {
...
+     * Throws {@code IllegalStateException} with the formatted string and the cause.
+     *
+     * @param cause the cause of the exception
+     * @param format the format string
+     * @param args formatting parameters
+     * @return nothing ever, always throws an exception. The return type is given for convenience.
+     * @throws IllegalStateException always
+     */
+    @CanIgnoreReturnValue
+    public static IllegalStateException newIllegalStateException(Throwable cause,
```
armiol, line 111: I'd like to propose two things.

1. Instead of copying the "throwing" code all the way, we could just throw corresponding exceptions by hand.

We don't have many cases in `testlib`, so I would say it's OK.

The main reason here is that eventually, users will have two JARs in their classpath: `testlib` and `base`. In this case, we'll have some ambiguity on imports.

2. For the rest of methods — for those which don't exist just to throw — mark the `Testing` type `@Internal`.

armiol (changes requested):
@alexander-yevsyukov please see my comments.

armiol (approved):

PR #789, opened 2023-04-19
Depend on separated modules

alexander-yevsyukov:
This PR finalizes the separation of `logging`, `reflect`, and `testlib` modules into dedicated repositories.

It turns out that `base` does not need to depend on `reflect`. Only `kotlin-reflect` dependency is required for `KClass.isFinal` property.

Also, the project was made single-module.


build.gradle.kts
```diff
@@ -69,19 +70,36 @@ spinePublishing {
...
-    JacocoConfig.applyTo(project)
-    LicenseReporter.mergeAllReports(project)
-    PomGenerator.applyTo(project)
+tasks {
+    excludeGoogleProtoFromArtifacts()
 }
 
+LicenseReporter.mergeAllReports(project)
+PomGenerator.applyTo(project)
+
```
armiol, line 105: This line seems redundant.

armiol (changes requested):
@alexander-yevsyukov LGTM in general. Please see a minor comment in this changeset, and I also suggest `buildSrc` contents to be updated once [`config#466`](https://github.com/SpineEventEngine/config/pull/466/files) is reviewed and merged.

armiol (approved):

PR #795, opened 2023-07-11
Merge `java-module` into `kotlin-jvm-module`, bump Guava

alexander-yevsyukov:
This PR:
  * **Eliminates the `java-module` script plugin**, merging its code into `kotlin-jvm-module`. Since our modules use both Java and Kotlin there's no sense in having a separate script plugin just for Java.
  *  Bumps Guava -> `32.1.1-jre`


armiol (changes requested):
@alexander-yevsyukov let's discuss @yevhenii-nadtochii comments in `config`, since they are very much about the same thing.

armiol (approved):

PR #800, opened 2023-08-03
Add a smoke test for `spine-logging

yevhenii-nadtochii:
This PR adds a smoke test for `spine-logging` to make sure that the logging is working in concrete modules. 

`base` module is the first one where this test is added. Further, it will be added to more modules. 

`Spine` dependency object will also be updated in `config` when this PR is approved.

### DIFF
--- a/src/test/kotlin/io/spine/LoggingSmokeTest.kt
+++ b/src/test/kotlin/io/spine/LoggingSmokeTest.kt
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2023, TeamDev. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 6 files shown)


yevhenii-nadtochii:
@alexander-yevsyukov PTAL

src/test/kotlin/io/spine/LoggingSmokeTest.kt
```diff
@@ -0,0 +1,36 @@
...
+package io.spine
+
+import io.spine.logging.AbstractLoggingSmokeTest
+
+/**
+ * Checks that `spine-logging` library works correctly within `base` module.
+ *
+ * The abstract base is provided by `Spine.Logging.smokeTest` dependency.
+ */
+internal class LoggingSmokeTest : AbstractLoggingSmokeTest()
```
alexander-yevsyukov, line 36: Look, this is too abstract (pun intended). We need to test that logging in this project does something useful for this library.

alexander-yevsyukov (changes requested):
Please see my comment.

alexander-yevsyukov (approved):
LGTM. Let's discuss the abstract test matter when @armiol is back.

PR #807, opened 2023-11-24
Update the dependencies from the latest `config`

armiol:
In scope of this changeset, the dependencies were updated to the latest available via `config`.

Additionally, some changes were made, as follows.

* New `dependencies.md` file is now in place instead of `license-report.md` (see [`config#498`](https://github.com/SpineEventEngine/config/issues/498)).
* A version of Kotlin `stdlib-jdk7` was forced, as it is now being transitively picked from some of our dependencies (see [`config#500`](https://github.com/SpineEventEngine/config/pull/500)).
* An explicit dependency onto Spine Reflect is added to allow the code to compile. Probably, earlier it was a transitive dependency of Spine Logging — but now it's not.


armiol:
@alexander-yevsyukov PTAL.

alexander-yevsyukov (approved):
LGTM

PR #813, opened 2024-02-03
Promote `Closeable` inteface from `io.spine.server`

alexander-yevsyukov:
This PR brings the `Closeable` interface which was for years a part of `core-java` on the server side to Spine Base. 

Recent works on code generation in `tool-base` show that we need a custom `Closeable` in modules that do not have `io.spine.server` package in their classpath (and we don't want to bring the dependency on `spine-server` artifact just because of the interface).

It is also expected that the interface would be useful for other usage scenarios that do not assume dependency on `spine-server`.

After copying the interface was converted to Kotlin and supplied with unit tests.

### Dependencies
 * Protobuf -> `3.25.2`

### DIFF
--- a/buildSrc/src/main/kotlin/io/spine/internal/dependency/Protobuf.kt
+++ b/buildSrc/src/main/kotlin/io/spine/internal/dependency/Protobuf.kt
@@ -33,7 +33,7 @@ package io.spine.internal.dependency
 )
 object Protobuf {
     private const val group = "com.google.protobuf"
-    const val version       = "3.25.1"
+    const val version       = "3.25.2"
     /**
      * The Java library containing proto definitions of Google Protobuf.
      */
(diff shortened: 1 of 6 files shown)


armiol (approved):
LGTM, and I really hope we won't die addressing the classpath conflicts in ProtoData. 

You'll probably will need to update `core-java` chain with this change, as ProtoData is built upon both Spine Server, and McJava (and hence, `tool-base`). That's just a wild guess, but we have some good chance to see some weird stuff with two _different_ `Closeable`s in classpath.

alexander-yevsyukov:
> That's just a wild guess, but we have some good chance to see some weird stuff with two _different_ `Closeable`s in classpath.

I'm planning to inherit `io.spine.server.Closable` from `io.spine.io.Closeable` for the time being. So, it should ease the pain.

PR #815, opened 2024-03-30
Bring extensions for `Base64` encoding from McJava

alexander-yevsyukov:
This PR adds extensions for `String` and `Path` that were previously adopted when advancing McJava. Since these extensions are generic enough, we bring them to the `base` project.

`config` was also updated.

### DIFF
--- a/src/main/kotlin/io/spine/string/Strings.kt
+++ b/src/main/kotlin/io/spine/string/Strings.kt
@@ -162,7 +162,7 @@ public fun Iterable<String>.indent(step: Indent, level: Int): String {
 }
 
 /**
- * Converts this string to base64 encoded version using UTF-8 charset.
+ * Converts this string to Base64 encoded version using UTF-8 charset.
  *
  * @see Base64
  */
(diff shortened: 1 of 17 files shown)


src/main/kotlin/io/spine/string/Strings.kt
```diff
@@ -29,6 +29,9 @@
 
 package io.spine.string
 
+import java.util.*
```
armiol, line 32: This is suspicious. Do we use that much from `java.util`?

Please review.

src/main/kotlin/io/spine/string/Strings.kt
```diff
@@ -157,3 +160,25 @@ public fun Iterable<String>.indent(step: Indent, level: Int): String {
         indentation + it
     }
 }
+
+/**
+ * Converts this string to Base64 encoded version using UTF-8 charset.
```
armiol, line 165: "Base64-encoded"

armiol (approved):
@alexander-yevsyukov LGTM with a couple of comments to address prior to merging.

PR #819, opened 2024-05-31
Parsing extensions for `CodeGeneratorRequest` and writing utility

alexander-yevsyukov:
This PR introduces utilities for parsing `CodeGeneratorRequest` having `ExtensionRegistry` initialised with all known custom Protobuf options.

Also the PR introduces `CodeGeneratorRequestWriter` utility class which aims to replace the code used in ProtoData and ProtoTap `protoc` plugins. Currently, the code of the plugins is identical and we want eliminate the duplication.

## Other notable changes
 * `ExtensionRegistryHolder` object was introduced to keep the reference to lazily constructed `ExtensionRegistry` with all known custom Protobuf options.
 * `TypeRegistryHolder` object was introduced to keep the reference to lazily constructed `TypeRegistry` with all Protobuf types gathered by `KnownTypes`.
 * `File.replaceExtension(String)` was introduced in addition to `Path.replaceExtension()`. The reason is that `toPath().replaceExtension("foo").toFile()` looks really awkward and mouthful, especially in tests.


src/main/kotlin/io/spine/code/proto/CodeGeneratorRequestWriter.kt
```diff
@@ -0,0 +1,97 @@
+/*
+ * Copyright 2024, TeamDev. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
```
armiol, line 8: Let's use the fresh copyright header, with `https` in this line.

It's here and in all newly created files.

src/main/kotlin/io/spine/code/proto/CodeGeneratorRequests.kt
```diff
@@ -0,0 +1,60 @@
+/*
+ * Copyright 2024, TeamDev. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
```
armiol, line 8: And here.

src/main/kotlin/io/spine/type/TypeRegistryHolder.kt
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright 2024, TeamDev. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
```
armiol, line 8: And here.

src/test/kotlin/io/spine/code/proto/CodeGeneratorRequestsSpec.kt
```diff
@@ -0,0 +1,83 @@
...
+        protoFile += descr.toProto()
+        fileToGenerate += descr.file.name
+        compilerVersion = version {
+            major = 42
+            minor = 314
+            patch = 271
+        }
+        parameter = parameterValue
+    }
+
```
armiol, line 83: Let's kill this line.

armiol (approved):
@alexander-yevsyukov LGTM with some comments.

PR #820, opened 2024-07-23
Utility method for checking UUID value

alexander-yevsyukov:
This PR extends `UuidValue` interface with a utility static method for checking a string representation of UUID value. The method will be used in new code generation for message types implementing the interface. 

Currently, the code of checking UUID value is repeated for all `UuidValue` types. This PR is the first set of "refactoring" this approach, so that the generated code would simply call `UuidValue.checkValid()`.

### DIFF
--- a/buildSrc/src/main/kotlin/io/spine/internal/dependency/ProtoData.kt
+++ b/buildSrc/src/main/kotlin/io/spine/internal/dependency/ProtoData.kt
@@ -5,7 +5,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- * http://www.apache.org/licenses/LICENSE-2.0
+ * https://www.apache.org/licenses/LICENSE-2.0
  *
  * Redistribution and use in source and/or binary forms, with or without
  * modification, must retain the above copyright notice and the following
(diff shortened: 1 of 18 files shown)


buildSrc/src/main/kotlin/io/spine/internal/dependency/ProtoData.kt
```diff
@@ -65,7 +65,7 @@ object ProtoData {
      * The version of ProtoData dependencies.
      */
     val version: String
-    private const val fallbackVersion = "0.21.3"
+    private const val fallbackVersion = "0.30.0"
```
armiol, line 68: I think we have a much newer version.

Also, I am not sure why we update the dogfooding version to some different value, than this.

Please review.

armiol (approved):
@alexander-yevsyukov LGTM with a comment to review.

buildSrc/src/main/kotlin/io/spine/internal/dependency/ProtoData.kt
alexander-yevsyukov: It came from `master` in `config`. I've updated it to latest ProtoData. It's not used by Base, but let's bump it anyway.

PR #824, opened 2024-09-10
Implement `ComparatorRegistry`

yevhenii-nadtochii:
This PR introduces the `ComparatorRegistry` for maintaining `T::class -> Comparator<T>` mapping of available comparators. New comparators can be registered using the `ComparatorProvider` service provider.

We are going to use this registry when generating Java code for handling `compare_by` option in Proto messages. 

Please note, the registry along **with its custom providers** should be present in both generation and generated message's runtimes. The registry usage is incorporated into the generated code only if we are sure that it has the appropriate comparator, which is checked during the code generation.

### DIFF
--- a/src/main/kotlin/io/spine/compare/ProtoComparators.kt
+++ b/src/main/kotlin/io/spine/compare/ProtoComparators.kt
@@ -33,10 +33,17 @@ import com.google.protobuf.util.Durations
 import com.google.protobuf.util.Timestamps
 
 /**
- * Registers comparators for some of the Protobuf Well-Known messages
- * in the [ComparatorRegistry].
+ * Registers comparators for Protobuf Well-Known messages in the [ComparatorRegistry].
  *
- * The used comparators are provided by Protobuf itself.
+ * The following messages are assigned a comparator:
+ *
+ * 1. [Timestamp].
+ * 2. [Duration].
+ *
+ * The used comparators are provided by Protobuf itself in a scope
+ * of `protobuf-java-util` library.
+ *
+ * See also: [Protobuf Docs | Well-Known Types](https://protobuf.dev/reference/protobuf/google.protobuf/).
  */
 @AutoService(ComparatorProvider::class)
 internal class ProtoComparators : ComparatorProvider {
--- a/src/main/kotlin/io/spine/compare/ProtoComparators.kt
+++ b/src/main/kotlin/io/spine/compare/ProtoComparators.kt
@@ -32,10 +32,16 @@ import com.google.protobuf.Timestamp
 import com.google.protobuf.util.Durations
 import com.google.protobuf.util.Timestamps
 
+/**
+ * Registers comparators for some of the Protobuf Well-Known messages
+ * in the [ComparatorRegistry].
+ *
+ * The used comparators are provided by Protobuf itself.
+ */
 @AutoService(ComparatorProvider::class)
-internal class ProtoWellKnownComparators : ComparatorProvider {
+internal class ProtoComparators : ComparatorProvider {
 
-    override fun provideIn(registry: ComparatorRegistry) = registry.run {
+    override fun registerIn(registry: ComparatorRegistry) = registry.run {
         register<Timestamp>(Timestamps.comparator())
         register<Duration>(Durations.comparator())
     }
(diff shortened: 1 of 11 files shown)


yevhenii-nadtochii:
@alexander-yevsyukov PTAL

src/test/kotlin/io/spine/compare/ComparatorRegistrySpec.kt
```diff
@@ -0,0 +1,82 @@
...
+import org.junit.jupiter.api.Test
+
+@DisplayName("`ComparatorRegistry` should`")
+internal class ComparatorRegistrySpec {
+
+    private val registry = ComparatorRegistry
+
+    @Test
+    fun `load the comparators from the providers`() {
+        registry.contains(Timestamp::class.java).shouldBeTrue()
```
alexander-yevsyukov, line 46: Why don't use `shouldBe true`? It's a bit easier to read, and two chars shorter.

I'm not insisting, just curious.

src/test/kotlin/io/spine/compare/ComparatorRegistrySpec.kt
```diff
@@ -0,0 +1,82 @@
...
+        val comparator = compareBy<Float> { it }
+        shouldThrow<IllegalStateException> { registry.get<Float>() }
+        registry.register<Float>(comparator)
+        registry.get<Float>() shouldBe comparator
+    }
+
+    @Test
+    fun `search for a comparator`() {
+        val comparator = compareBy<StringBuilder> { it.length }
+        registry.find<StringBuilder>().shouldBeNull()
```
alexander-yevsyukov, line 78: One thing to consider is the consistency of assertions. We mostly use the infix form of assertions. It forms some rhythm and texture in the code. The non-infix assertions break the texture.

alexander-yevsyukov (changes requested):
Almost there. Please see my comments.

src/main/kotlin/io/spine/compare/ProtoComparators.kt
```diff
@@ -0,0 +1,48 @@
...
+package io.spine.compare
+
+import com.google.auto.service.AutoService
+import com.google.protobuf.Duration
+import com.google.protobuf.Timestamp
+import com.google.protobuf.util.Durations
+import com.google.protobuf.util.Timestamps
+
+/**
+ * Registers comparators for some of the Protobuf Well-Known messages
```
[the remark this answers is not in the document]
yevhenii-nadtochii: Only these two are provided by Protobuf itself.

src/main/kotlin/io/spine/compare/ProtoComparators.kt
```diff
@@ -0,0 +1,55 @@
...
+package io.spine.compare
+
+import com.google.auto.service.AutoService
+import com.google.protobuf.Duration
+import com.google.protobuf.Timestamp
+import com.google.protobuf.util.Durations
+import com.google.protobuf.util.Timestamps
+
+/**
+ * Registers comparators for Protobuf Well-Known messages in the [ComparatorRegistry].
```
alexander-yevsyukov, line 36: Please change to read:

```
Register comparators provided by `protobuf-java-util` library for `Timestamp` and `Duration` types.
```

First of all, it's shorter. Also, currently the first sentence gives the impression that we register comparators for all the Well-Known types. Of course, it's negated by the following text. But it makes the readers "sum up" the pieces of texts.

src/main/kotlin/io/spine/compare/ProtoComparators.kt
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * The following messages are assigned a comparator:
+ *
+ * 1. [Timestamp].
+ * 2. [Duration].
+ *
+ * The used comparators are provided by Protobuf itself in a scope
+ * of `protobuf-java-util` library.
+ *
+ * See also: [Protobuf Docs | Well-Known Types](https://protobuf.dev/reference/protobuf/google.protobuf/).
```
alexander-yevsyukov, line 46: Please remove the "see also" bit. It's not necessary given the change I reguested above.

alexander-yevsyukov (approved):
LGTM, provided the documentation change comments I've made are addressed.

PR #826, opened 2024-09-13
Add `CharSequence.containsAny()` extension function

alexander-yevsyukov:
This PR adds `CharSequence.containsAny()` extension function so that it's easier to detect several prohibited characters.

### DIFF
--- a/src/main/kotlin/io/spine/string/CharSequenceExts.kt
+++ b/src/main/kotlin/io/spine/string/CharSequenceExts.kt
@@ -35,7 +35,19 @@ public fun CharSequence.contains(s: Separator): Boolean =
     contains(s.value)
 
 /**
- * Obtains all line separators found it this sequence.
+ * Tells if this sequence contains any of the given characters.
+ */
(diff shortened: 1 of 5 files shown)


src/main/kotlin/io/spine/string/CharSequenceExts.kt
```diff
@@ -35,7 +35,19 @@ public fun CharSequence.contains(s: Separator): Boolean =
...
+ * Tells if this sequence contains any of the given characters.
+ */
+public fun CharSequence.containsAny(vararg char: Char): Boolean {
+    char.forEach {
+        if (contains(it)) {
+            return true
+        }
+    }
+    return false
+}
```
Oleg-Melnik: There is a shorter form of the same functionality:
```
    return char.any {
        contains(it)
    }
```

Oleg-Melnik (approved):
@alexander-yevsyukov LGTM with the minor improvement to address prior to merging.

PR #833, opened 2024-10-28
Add `if_set_again` field option

yevhenii-nadtochii:
This PR creates `IfSetAgainOption` to use in conjunction with `set_once_option`. The added option allows specifying the custom error message in case the field's value is re-assigned.

### DIFF
--- a/src/main/proto/spine/options.proto
+++ b/src/main/proto/spine/options.proto
@@ -267,7 +267,10 @@ extend google.protobuf.FieldOptions {
     //
     bool column = 73854;
 
-    // Reserved 73855 to 73890 for future options.
+    // See `IfSetAgainOption`.
+    IfSetAgainOption if_set_again = 73855;
+
+    // Reserved 73856 to 73890 for future options.
 
     // Reserved 73900 for removed `by` option.
 }
@@ -1002,4 +1005,24 @@ message CompareByOption {
     // If true, the default order is reversed. For example, numbers are ordered from the greater to
     // the lower, enums — from the last number value to the 0th value, etc.
     bool descending = 2;
+}
+
+// Defines the error message used if a `set_once` field is set again.
+//
+// Applies only to the fields marked as `set_once`.
+//
+// Example: Using the `(set_once)` option.
+//
+//     message User {
+//         UserId id = 1 [(set_once) = true,
+//                        (if_set_again).error_msg = "The student ID can't be re-assigned."];
+//     }
+//
+message IfSetAgainOption {
+
+    // The default error message for the field.
+    option (default_message) = "The field value can't be re-assigned.";
+
+    // A user-defined error message.
+    string error_msg = 1;
 }
--- a/version.gradle.kts
+++ b/version.gradle.kts
@@ -24,4 +24,4 @@
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-val versionToPublish: String by extra("2.0.0-SNAPSHOT.215")
+val versionToPublish: String by extra("2.0.0-SNAPSHOT.216")


yevhenii-nadtochii:

src/main/proto/spine/options.proto
```diff
@@ -1003,3 +1006,23 @@ message CompareByOption {
...
+//
+//     message User {
+//         UserId id = 1 [(set_once) = true,
+//                        (if_set_again).error_msg = "The student ID can't be re-assigned."];
+//     }
+//
+message IfSetAgainOption {
+
+    // The default error message for the field.
+    option (default_message) = "The field value can't be re-assigned.";
```
alexander-yevsyukov, line 1024: Let's use the full form "cannot" instead of "can't".

It is recommended to use full forms in technical documentation. This is so because full forms are more visible than the short ones. When it comes to serious matters such a small difference in reader's perception can save lives.

src/main/proto/spine/options.proto
```diff
@@ -267,7 +267,10 @@ extend google.protobuf.FieldOptions {
     //
     bool column = 73854;
 
-    // Reserved 73855 to 73890 for future options.
+    // See `IfSetAgainOption`.
```
alexander-yevsyukov, line 270: The documentation should explain what this type is about. And then give additional information.

src/main/proto/spine/options.proto
```diff
@@ -1003,3 +1006,23 @@ message CompareByOption {
...
+// Example: Using the `(set_once)` option.
+//
+//     message User {
+//         UserId id = 1 [(set_once) = true,
+//                        (if_set_again).error_msg = "The student ID can't be re-assigned."];
+//     }
+//
+message IfSetAgainOption {
+
+    // The default error message for the field.
```
alexander-yevsyukov, line 1023: Which field? How about telling something like this

```proto
// The default error message used when the field with the `(set_once)` option is re-assigned.
```

I'm not insisting on this text, but we need to help the user more.

src/main/proto/spine/options.proto
alexander-yevsyukov, line 1024: Why don't we name the field in the message? We should. We also should tell the current value, and the new one. This saves time in debugging.

alexander-yevsyukov (changes requested):
Please see my comments.

yevhenii-nadtochii:
@alexander-yevsyukov Please take another look.

alexander-yevsyukov (changes requested):
Please see my comments.

src/main/proto/spine/options.proto
```diff
@@ -1003,3 +1012,31 @@ message CompareByOption {
...
+    // 2. `{currentValue}` — the current field value.
+    // 3. `{proposedValue}` — the value that was attempted to be set.
+    //
+    // The tokens will be replaced at runtime when the error is constructed.
+    //
+    // Example: Using the `(set_once)` option.
+    //
+    //     message User {
+    //         UserId id = 1 [(set_once) = true,
+    //                        (if_set_again).error_msg = "The student ID field cannot be re-assigned. The current value: `{currentValue}`. The proposed value: `{proposedValue}`."];
```
armiol, line 1038: Let's try to shorten this example message:
```
The field `{fieldName}` already has value `{currentValue}`, and cannot be re-assigned to `{proposedValue}`.
```

armiol (approved):
@yevhenii-nadtochii LGTM with the respect to my suggestion.

src/main/proto/spine/options.proto
alexander-yevsyukov: Can we have it as the default message, please? It tells what we need to tell in a general way.

The example here should tell something about the student ID as a particular case of custom domain message.

alexander-yevsyukov (changes requested):
Almost LGTM. Should we finalise the error message language in a video call?

yevhenii-nadtochii:
@alexander-yevsyukov PTAL

src/main/proto/spine/options.proto
```diff
@@ -1003,3 +1012,31 @@ message CompareByOption {
...
 }
+
+// Defines the error message used if a `set_once` field is set again.
+//
+// Applies only to the fields marked as `set_once`.
+//
+message IfSetAgainOption {
+
+    // The default error message used when the field value is re-assigned.
+    option (default_message) = "The field `{fieldName}` already has value `{currentValue}`, and cannot be re-assigned to `{proposedValue}`.";
```
alexander-yevsyukov, line 1023: It's probably "has the value". Please consult with ChatGPT.

alexander-yevsyukov (approved):
LGTM with one comment.

PR #834, opened 2024-11-01
String extensions for diagnostic output

alexander-yevsyukov:
This PR adds two extension functions for brevity when forming string templates.

The function `simply<T>()` obtains `simpleName` of the class of the generic parameter `T`. It shorter than `T::class.simpleName`.

The function `Message.shortly()` calls another extension function `shortDebugString()`, but, again, is also more convenient in the context of a string template.

## Build changes
 * Gradle was bumped to `8.10.2`.
 * Latest `config` with new dependency objects was applied.


armiol (approved):
@alexander-yevsyukov LGTM, once the builds pass.

PR #836, opened 2024-11-21
Extract creating `DescriptorReference` file

alexander-yevsyukov:
This PR enhances the `DescriptorReference` class encapsulating calculation of the reference file in the given directory.

Also, the related tests were converted to Kotlin, eliminating the code duplication.


armiol (approved):
@alexander-yevsyukov LGTM, with the respect of re-using the existing `Separator` definitions instead of hard-coding the UNIX/Windows separators in the test. Which was discussed vocally.

PR #838, opened 2024-12-02
Update docs to `(set_once)` option

yevhenii-nadtochii:
This PR adds more details to `(set_once)` option specification.

### DIFF
--- a/src/main/proto/spine/options.proto
+++ b/src/main/proto/spine/options.proto
@@ -147,6 +147,17 @@ extend google.protobuf.FieldOptions {
 
     // Indicates that a field can only be set once.
     //
+    // The field is considered unset as long as it has the default value. Setting a value over
+    // the default one is always allowed because it is not considered as "overriding".
+    // An overriding of a non-default value with the same value is also allowed. Please note,
+    // as for now, there is no support for explicit `optional` fields.
+    //
+    // The option can be applied to any singular field type:
+    //
+    // - any message or enum type;
+    // - any number type;
+    // - `bool`, `string` and `bytes`.
+    //
     // A typical use-case would include a value of an ID, which doesn't change over the course of
     // the life of an entity.
     //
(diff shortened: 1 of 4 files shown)


yevhenii-nadtochii:

armiol (changes requested):

alexander-yevsyukov (approved):
LGTM, provided that the comment from @armiol is addressed.

yevhenii-nadtochii:
@armiol PTAL

src/main/proto/spine/options.proto
```diff
@@ -147,8 +147,50 @@ extend google.protobuf.FieldOptions {
 
     // Indicates that a field can only be set once.
     //
-    // A typical use-case would include a value of an ID, which doesn't change over the course of
-    // the life of an entity.
+    // This option allows the target field to accept assignments only if one of the following
+    // conditions is met:
+    //
+    // 1. The current filed value is the default for its type.
```
armiol, line 153: "field".

src/main/proto/spine/options.proto
```diff
@@ -147,8 +147,50 @@ extend google.protobuf.FieldOptions {
...
+    // 2. The current field value equals to the proposed new value.
+    //
+    // The option can be applied to the following singular field types:
+    //
+    // - any message or enum type;
+    // - any numeric type;
+    // - `bool`, `string` and `bytes`.
+    //
+    // Repeated fields, maps, and fields with explicit `optional` cardinality are not supported.
+    // Such declarations can lead to build-time errors. For more information on field cardinality,
```
armiol, line 164: "can" or "will"?

Such an uncertainty isn't good.

src/main/proto/spine/options.proto
```diff
@@ -147,8 +147,50 @@ extend google.protobuf.FieldOptions {
...
+    // 5. Merging specific fields (available for Message fields).
+    //
+    // Unsupported Java use cases include:
+    //
+    // 1. Constructing a message using the `DynamicMessage` class.
+    // 2. Merging data from messages provided by third-party Protobuf implementations.
+    // 3. Clearing a specific field or an entire message.
+    //
+    // For unsupported scenarios, the option performs no validation. It does not throw errors
+    // or prints warnings.
```
armiol, line 190: "does not throw... or print...".

src/main/proto/spine/options.proto
```diff
@@ -159,7 +201,8 @@ extend google.protobuf.FieldOptions {
     // Once set, the `id` field cannot be changed.
     //
     // Use `(if_set_again).error_msg` option to specify a custom error message that will be shown
```
armiol, line 203: I would change this "shown" to something that describes the real world better. Maybe it could be as follows:

"... that will be used for composing the error upon attempting to re-assign the field value."

armiol (changes requested):

PR #842, opened 2024-12-11
Use named placeholder for error messages in `options.proto`

yevhenii-nadtochii:
This PR unifies error messages format for all Spine options. Previously, the default error messages used `printf`-style format, but the custom ones expected placeholders instead. Now, both the custom and the default messages expect a string, which may contain placeholders in an arbitrary order.

A number of placeholders have been increased to allow specifying more detailed errors.

### DIFF
--- a/src/main/proto/spine/options.proto
+++ b/src/main/proto/spine/options.proto
@@ -933,7 +933,7 @@ message GoesOption {
     // 4. `{parent.type}` – the fully qualified name of the field declaring type.
     // 5. `{goes.companion}` – the name of the companion specified in `with`.
     //
-    // The tokens will be replaced at runtime when the error is constructed.
+    // The placeholders will be replaced at runtime when the error is constructed.
     //
     string error_msg = 3;
 }
@@ -1119,7 +1119,7 @@ message IfSetAgainOption {
     // 4. `{field.proposed_value}` – the value, which was attempted to be set for the field.
     // 5. `{parent.type}` – the fully qualified name of the field declaring type.
     //
-    // The tokens will be replaced at runtime when the error is constructed.
+    // The placeholders will be replaced at runtime when the error is constructed.
     //
     // Example: Using the `(set_once)` option.
     //
--- a/src/main/proto/spine/options.proto
+++ b/src/main/proto/spine/options.proto
@@ -135,9 +135,7 @@ extend google.protobuf.FieldOptions {
     //
     //   1. A message field value satisfies the validation constraints defined in the corresponding
     //      message type of the field.
-    //
     //   2. Each value of a map entry satisfies validation constraints.
-    //
     //   3. Each item of a repeated field satisfies validation constraints.
     //
     bool validate = 73821;
@@ -1130,7 +1128,7 @@ message IfSetAgainOption {
 message IfHasDuplicatesOption {
 
     // The default error message.
-    option (default_message) = "The field `{parent.type}.{field.name}` of the type `{field.type}` must not contain duplicates. The found duplicates: `{field.duplicates}`.";
+    option (default_message) = "The field `{parent.type}.{field.name}` of the type `{field.type}` must not contain duplicates. The duplicates found: `{field.duplicates}`.";
 
     // A user-defined error message.
     //
@@ -1139,7 +1137,7 @@ message IfHasDuplicatesOption {
     // 1. `{field.name}` – the field name.
     // 2. `{field.type}` – the fully qualified name of the field type.
     // 3. `{field.value}` – the field value (the whole collection).
-    // 4. `{field.duplicates}` – the found duplicates (elements that occur more than once).
+    // 4. `{field.duplicates}` – the duplicates found (elements that occur more than once).
     // 5. `{parent.type}` – the fully qualified name of the field declaring type.
     //
     // The placeholders will be replaced at runtime when the error is constructed.
@@ -1148,7 +1146,7 @@ message IfHasDuplicatesOption {
     //
     //    message Blizzard {
     //        repeated Snowflake = 1 [(distinct) = true,
-    //                                (if_has_duplicates).error_msg = "Every snowflake must be unique! The found duplicates: `{field.duplicates}`."];
+    //                                (if_has_duplicates).error_msg = "Every snowflake must be unique! The duplicates found: `{field.duplicates}`."];
     //    }
     //
     string error_msg = 1;
@@ -1179,7 +1177,7 @@ message RangeOption {
     //         double angle = 4 [(range).value = "(0.0..180.0)"];
     //     }
     //
-    // NOTE: That definition of ranges must be consistent with the type they constrain.
+    // NOTE: That definition of ranges must be consistent with the type they describe.
     //       An range for an integer field must be defined with integer endpoints.
     //       A range for a floating point field must be defined with decimal separator (`.`),
     //       even if the endpoint value does not have a fractional part.
--- a/src/main/proto/spine/options.proto
+++ b/src/main/proto/spine/options.proto
@@ -1156,12 +1156,17 @@ message IfHasDuplicatesOption {
 }
 
 // The option to indicate that a numeric field is required to have a value which belongs
-// to the specified bounded range. For unbounded ranges, please use `(min)` and `(max) options.
+// to the specified bounded range.
+//
+// For unbounded ranges, please use `(min)` and `(max) options.
+//
 message RangeOption {
 
     // The default error message.
     option (default_message) = "The field `{parent.type}.{field.name}` must be within the following range: `{range.value}`.";
 
+    // The string representation of the range.
+    //
     // The range can be open (not including the endpoint) or closed (including the endpoint) on
     // each side. Open endpoints are indicated using a parenthesis (`(`, `)`). Closed endpoints are
     // indicated using a square bracket (`[`, `]`).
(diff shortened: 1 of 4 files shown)


yevhenii-nadtochii:
@armiol @alexander-aleksandrov PTAL

src/main/proto/spine/options.proto
```diff
@@ -668,19 +676,14 @@ message IfMissingOption {
 //         double value = 1 [(min) = {
 //             value: "0.0",
 //             exclusive: true,
-//             error_msg: "Temperature cannot reach {other}K, but provided {value}."
+//             error_msg: "The temperature cannot reach `{min.threshold}K`, but provided `{field.value}`."
```
armiol, line 679: I would simplify to `{min.value}`.

src/main/proto/spine/options.proto
```diff
@@ -668,19 +676,14 @@ message IfMissingOption {
...
 
-    // The default error message format string.
-    //
-    // The format parameters are:
-    //   1) "or equal to " string (if the `exclusive` parameter is false) or an empty string;
-    //   2) the minimum number.
-    //
-    option (default_message) = "The number must be greater than %s%s.";
+    // The default error message.
+    option (default_message) = "The field `{parent.type}.{field.name}` must be {max.comparison} {min.threshold}.";
```
armiol, line 686: How about `{max.operator}`?

src/main/proto/spine/options.proto
```diff
@@ -694,34 +697,39 @@ message MinOption {
...
     //
-    // May include tokens `{value}`—for the actual value of the field, and `{other}`—for
-    // the threshold value. The tokens will be replaced at runtime when the error is constructed.
+    // 1. `{field.value}` - the field value.
+    // 2. `{field.name}` – the field name.
+    // 3. `{field.type}` – the fully qualified name of the field type.
+    // 4. `{parent.type}` – the fully qualified name of the field declaring type.
+    // 5. `{min.threshold}` – the specified minimum threshold `value`.
+    // 6. `{min.comparison}` – if `exclusive` is set to `true`, this placeholder equals
+    //    to "greater than". Otherwise, "greater than or equal to".
```
armiol, line 710: As discussed today, let's use `>` and `>=`.

armiol (changes requested):
@yevhenii-nadtochii LGTM except for some comments.

yevhenii-nadtochii:
@armiol PTAL

Though, we still have `(distinct)` and `(range)`, for which there are no custom messages, nor the default ones on the level of the option definitions. The default messages for them are still defined as string literals within `validation`.

armiol (approved):
@yevhenii-nadtochii LGTM.

But what if we define the corresponding behaviour for `(distinct)` and `(range)` in this PR? It would be easier to review all the proposed placeholder names together.

Please consider.

src/main/proto/spine/options.proto
```diff
@@ -630,22 +620,14 @@ extend google.protobuf.ServiceOptions {
...
-//    message Holder {
-//        MyMessage field = 1 [(required) = true,
-//                             (if_missing).error_msg = "This field is required."];
-//    }
 //
 message IfMissingOption {
 
     // The default error message.
-    option (default_message) = "A value must be set.";
+    option (default_message) = "The field `{parent.type}.{field.name}` of type `{field.type}` must have a value.";
```
alexander-yevsyukov, line 630: Isn't it "the type"?

src/main/proto/spine/options.proto
```diff
@@ -834,28 +847,39 @@ message PatternOption {
...
-//    }
+// It is applicable only to message fields marked with `(validate)`.
+// Repeated fields are supported.
 //
 message IfInvalidOption {
 
-    // The default error message for the field.
-    option (default_message) = "The message must have valid properties.";
+    // The default error message.
+    option (default_message) = "The field `{parent.type}.{field.path}` of type `{field.type}` is invalid. The field value: `{field.value}`.";
```
alexander-yevsyukov, line 857: Why are we having here `field.path` instead of `field.name`? We refer to the immediate placement of the field in this context.

alexander-yevsyukov: We have the combination of `{parent.type}.{field.name}` above.

src/main/proto/spine/options.proto
```diff
@@ -1072,15 +1099,18 @@ message CompareByOption {
 //
 message IfSetAgainOption {
 
-    // The default error message used when the field value is re-assigned.
-    option (default_message) = "The field `{fieldName}` already has the value `{currentValue}` and cannot be reassigned to `{proposedValue}`.";
+    // The default error message.
+    option (default_message) = "The field `{parent.type}.{field.name}` of `{field.type}` already has the value `{field.value}` and cannot be reassigned to `{field.proposed_value}`.";
```
alexander-yevsyukov, line 1103: Can we have "of the type `{field.type}`" instead of "of `{field.type}`"?

alexander-yevsyukov: Here and in similar case above and below, please.

alexander-yevsyukov (changes requested):
Please see my comments. I'll review it once again after this batch is addressed.

yevhenii-nadtochii:

The two remaining options have got the default and custom messages support:

1. For the boolean `(distinct)` constraint, I've added a companion option `(if_has_duplicates)`.
2. `(range)` option became a message option – `RangeOption`.

src/main/proto/spine/options.proto
```diff
@@ -1093,3 +1122,81 @@ message IfSetAgainOption {
...
 }
+
+// Defines the error message used if a `distinct` field has duplicates.
+//
+// Applies only to the repeated fields marked as `distinct`.
+//
+message IfHasDuplicatesOption {
+
+    // The default error message.
+    option (default_message) = "The field `{parent.type}.{field.name}` of the type `{field.type}` must not contain duplicates. The found duplicates: `{field.duplicates}`.";
```
armiol, line 1133: "The duplicates found:"

src/main/proto/spine/options.proto
```diff
@@ -1093,3 +1122,81 @@ message IfSetAgainOption {
...
+    // 4. `{field.duplicates}` – the found duplicates (elements that occur more than once).
+    // 5. `{parent.type}` – the fully qualified name of the field declaring type.
+    //
+    // The placeholders will be replaced at runtime when the error is constructed.
+    //
+    // Example: Using the `(distinct)` option.
+    //
+    //    message Blizzard {
+    //        repeated Snowflake = 1 [(distinct) = true,
+    //                                (if_has_duplicates).error_msg = "Every snowflake must be unique! The found duplicates: `{field.duplicates}`."];
```
armiol, line 1151: Same as above.

src/main/proto/spine/options.proto
```diff
@@ -1093,3 +1122,81 @@ message IfSetAgainOption {
...
+    option (default_message) = "The field `{parent.type}.{field.name}` of the type `{field.type}` must not contain duplicates. The found duplicates: `{field.duplicates}`.";
+
+    // A user-defined error message.
+    //
+    // The specified message may include the following placeholders:
+    //
+    // 1. `{field.name}` – the field name.
+    // 2. `{field.type}` – the fully qualified name of the field type.
+    // 3. `{field.value}` – the field value (the whole collection).
+    // 4. `{field.duplicates}` – the found duplicates (elements that occur more than once).
```
armiol, line 1142: Same as above.

src/main/proto/spine/options.proto
```diff
@@ -1093,3 +1122,81 @@ message IfSetAgainOption {
     //
     string error_msg = 1;
 }
+
+// Defines the error message used if a `distinct` field has duplicates.
```
armiol, line 1126: It would be good to mention which objects we consider to be duplicates.

This is more challenging than it seems, because some objects, like Entities, have IDs. And users of this API might think that could be some special case: comparing by IDs only. Which, of course, it isn't.

src/main/proto/spine/options.proto
```diff
@@ -1093,3 +1122,81 @@ message IfSetAgainOption {
...
+    // Example: Defining ranges of numeric values.
+    //
+    //     message NumRanges {
+    //         int32 hour = 1 [(range).value = "[0..24)"];
+    //         uint32 minute = 2 [(range).value = "[0..59]"];
+    //         float degree = 3 [(range).value = "[0.0..360.0)"];
+    //         double angle = 4 [(range).value = "(0.0..180.0)"];
+    //     }
+    //
+    // NOTE: That definition of ranges must be consistent with the type they constrain.
```
armiol, line 1182: I would say "... they limit", or "... they describe".

src/main/proto/spine/options.proto
```diff
@@ -834,28 +847,38 @@ message PatternOption {
...
-// Repeated fields are supported.
-//
-// Example: Using the `(if_invalid)` option.
 //
-//     message Holder {
-//         MyMessage field = 1 [(validate) = true,
-//                              (if_invalid).error_msg = "The field is invalid."];
-//    }
+// It is applicable only to message fields marked with `(validate)`.
+// Repeated fields are supported.
```
armiol, line 852: What about `map`s?

armiol (changes requested):
@yevhenii-nadtochii LGTM in general. Please see my comments.

alexander-yevsyukov (approved):
LGTM

yevhenii-nadtochii:
@armiol PTAL

I've updated docs to `(validate)` and `(distinct)`.

armiol (approved):

PR #843, opened 2024-12-13
Properly migrate the field type in `options.proto`

yevhenii-nadtochii:
This PR assigns a new field number to the field, which has changed the type in #842. 

Not following this convention, breaks compilation of ProtoData with `InvalidWireTypeException`. 

I've tested the local publication. ProtoData builds successfully with this change.


yevhenii-nadtochii:

armiol (approved):

PR #847, opened 2025-01-07
Talk about `field.path` in API of validation options

yevhenii-nadtochii:
This PR updates `field.name` notation to `field.path`. 

`(if_invalid)` option was removed. We have no plans to retain nested `ConstraintViolation`s. One error will mean one constraint violation, and it will use the original error message. The error message specified for `(validate)` is no longer needed.

Another aspect is that the options below expect field names to be passed to them, but may report invalid nested fields:

1. `(required_field)`.
3. `(goes_option)`.

If I write there `field.path`, it can be interpreted as "the option supports nested fields here". But they clearly do not support nested fields as input like `(compare_by)`, for example.

### DIFF
--- a/dependencies.md
+++ b/dependencies.md
@@ -1,6 +1,6 @@
 
 
-# Dependencies of `io.spine:spine-base:2.0.0-SNAPSHOT.234`
+# Dependencies of `io.spine:spine-base:2.0.0-SNAPSHOT.235`
 
 ## Runtime
 1.  **Group** : com.google.code.findbugs. **Name** : jsr305. **Version** : 3.0.2.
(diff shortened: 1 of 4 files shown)


armiol:
@yevhenii-nadtochii `parent` in this context means the instance of `Message` which was initially passed to the validation. 

Consider we have `UserAccount`, containing `Address address = 42 [(validate) = true];`, which in turn has `ZipCode zip_code = 17  [(validate) = true];`

So if we pass an instance of `UserAccount`, and the ZIP code in it is invalid, the `parent.type` is still `foo.bar.acme.UserAccount`.

yevhenii-nadtochii:
@armiol I've updated docs to `parent.type`. See the last commit.

armiol (approved):

PR #848, opened 2025-01-08
Restore and deprecate `if_invalid` option

yevhenii-nadtochii:
This PR restores binary compatibility with older versions. 

I can't migrate `spine-core` to the latest versions of `mc-java` and `validation`. And I can't remove usages of `if_invalid` without bumping the mentioned libs because the old versions were deleted from Maven.


yevhenii-nadtochii:

alexander-yevsyukov (approved):
LGTM

PR #853, opened 2025-02-20
Integrate collections conversion into `TypeConverter`

### LINKED ISSUE #851: Extend `ProtoConverter` to support `List` and `Map`
An attempt to convert a list from Java to Proto fails with the following exception:

```
        java.lang.IllegalArgumentException: Could not find a wrapper type for `java.util.Collections$EmptyList`.
            at com.google.common.base.Preconditions.checkArgument(Preconditions.java:218)
            at io.spine.protobuf.PrimitiveConverter.primitiveConverter(PrimitiveConverter.java:112)
            at io.spine.protobuf.PrimitiveConverter.toMessage(PrimitiveConverter.java:93)
            at io.spine.protobuf.ProtoConverter.doBackward(ProtoConverter.java:99)
            at io.spine.protobuf.ProtoConverter.doBackward(ProtoConverter.java:46)
            at com.google.common.base.Converter.correctedDoBackward(Converter.java:212)
            at com.google.common.base.Converter$ReverseConverter.correctedDoForward(Converter.java:341)
            at com.google.common.base.Converter.convert(Converter.java:195)
            at io.spine.protobuf.TypeConverter.toMessage(TypeConverter.java:122)
            at io.spine.protobuf.TypeConverter.toAny(TypeConverter.java:104)
```

It attempts to handle `java.util.Collections$EmptyList` with a `PrimitiveConverter`, and inevitably fails. Please add support for `List` and `Map` types. They come from Protobuf `repeated` and `map` types.

yevhenii-nadtochii:
This PR extends `TypeConverter` to add support of collections conversion to Protobuf `Any`. 

We would like to keep a single entry point for Protobuf `Any` conversions, so that the following expression works well for all common types, including `List` (proto `repeated`) and `Map` (proto `map`):

```kotlin
/**
 * Wraps this [Expression] into Protobuf `Any` using [TypeConverter.toAny] method.
 */
public fun Expression<*>.packToAny(): Expression<ProtoAny> {
    val type = ClassName(TypeConverter::class)
    return type.call("toAny", arguments = listOf(this))
}
```

Otherwise, we would have to keep in mind this split (`TypeConverter` vs `CollectionsConverter`) whenever we generate or test the generated code.

For this, `ListConverter` and `MapConverter` were implemented and integrated into `TypeConverter`. 

Please note, for collections, only one-way conversion is supported, from Kotlin/Java to Protobuf message. These converters throw when asked to convert Protobuf message to `List` or `Map`. Support of such a conversion requires changes to `TypeConverter` API or to the structure of target Protobuf messages.

The signature `TypeConverter.toObject(Any message, Class<T> target)` can cover `List` or `Map` classes, but will not allow getting a type of generic arguments for these collections. One generic for `List` and two for `Map`. As for now, we just don't need this.

The original `CollectionsConverter` was deprecated in favor of `TypeConverter`.

Resolves #851.


yevhenii-nadtochii:

Besides the issue discovered yesterday (with `build` directories in ProtoData), I had two more today:

1. In ProtoData, Gradle didn't react to the version changing under `buildSrc/.../Base.version` until I manually removed `buildSrc/build` directory.
2. In Base, the `generatePom` task was not executed along with `build`. I had to run `./gradlew generatePom` to force its execution. `build` is finalized by `generatePom`. Should always run even if `build` itself is skipped.

This is odd. Maybe something is wrong with Gradle on my machine.

armiol (changes requested):
LGTM in general. Please see my comments.

src/main/kotlin/io/spine/protobuf/ListConverter.kt
```diff
@@ -0,0 +1,50 @@
...
+package io.spine.protobuf
+
+import io.spine.annotation.Internal
+import io.spine.base.ListOfAnys
+import io.spine.base.listOfAnys
+import io.spine.protobuf.TypeConverter.toAny
+
+/**
+ * Converts Kotlin's list of [Any] to [ListOfAnys] proto message.
+ */
```
alexander-yevsyukov, line 35: Please add that backward conversion is not supported in the header of this class, and for other similar classes too.

It's a bit too late to tell about it in the method. People should be warned about the missing capability when they familiarise themselves with the thing.

src/main/kotlin/io/spine/protobuf/ListConverter.kt
```diff
@@ -0,0 +1,50 @@
...
+package io.spine.protobuf
+
+import io.spine.annotation.Internal
+import io.spine.base.ListOfAnys
+import io.spine.base.listOfAnys
+import io.spine.protobuf.TypeConverter.toAny
+
+/**
+ * Converts Kotlin's list of [Any] to [ListOfAnys] proto message.
```
alexander-yevsyukov, line 34: What happens when we attempt to use this class from Java?

alexander-yevsyukov (changes requested):
Please see my comments and questions

yevhenii-nadtochii:

armiol (approved):
@yevhenii-nadtochii LGTM from my side.

src/main/java/io/spine/protobuf/TypeConverter.java
```diff
@@ -56,6 +56,16 @@
...
+ *
+ * <p>Note that the following types can only be converted in one direction,
+ * from {@linkplain Object Java Objects} to {@link Any}:
+ * <ul>
+ *     <li>{@link java.util.List Java List} types via {@link ListConverter}.
+ *     <li>{@link java.util.Map Java Map} types via {@link MapConverter}.
+ * </ul>
+ *
+ * <p>The converter throws {@link UnsupportedOperationException} if asked to perform
+ * an unsupported conversion.
```
alexander-yevsyukov, line 68: Let's talk specifically here. We already mentioned the "unsupported" thing in the name of the exception.

src/main/kotlin/io/spine/protobuf/MapConverter.kt
```diff
@@ -0,0 +1,64 @@
...
+import io.spine.annotation.Internal
+import io.spine.base.MapOfAnys
+import io.spine.base.MapOfAnysKt.entry
+import io.spine.base.mapOfAnys
+import io.spine.protobuf.TypeConverter.toAny
+
+/**
+ * Converts Kotlin's map of [Any] to [MapOfAnys] proto message.
+ *
+ * Note that the backward conversion from [MapOfAnys] to Kotlin's map
```
alexander-yevsyukov, line 38: The "backward conversion" text should be a link to the corresponding function.

src/main/kotlin/io/spine/protobuf/MapConverter.kt
```diff
@@ -0,0 +1,64 @@
...
+package io.spine.protobuf
+
+import io.spine.annotation.Internal
+import io.spine.base.MapOfAnys
+import io.spine.base.MapOfAnysKt.entry
+import io.spine.base.mapOfAnys
+import io.spine.protobuf.TypeConverter.toAny
+
+/**
+ * Converts Kotlin's map of [Any] to [MapOfAnys] proto message.
```
alexander-yevsyukov, line 36: Let's drop the "Kotlin" word in this documentation entirely, and "Java" too. It just adds unnecessary details that people programming in these two languages know already.

alexander-yevsyukov (changes requested):
Please see my comments. Let's have a quick video chat regarding the Kotlin & Java comment. It would be easier to address it via voice.

yevhenii-nadtochii:
@alexander-yevsyukov PTAL

src/main/java/io/spine/protobuf/TypeConverter.java
```diff
@@ -56,6 +58,21 @@
...
+ *
+ * <p>The following collection types can only be converted in one direction,
+ * from {@linkplain Object Java Objects} to {@link Any}:
+ * <ul>
+ *     <li>{@link java.util.List Java List} types via {@link ListConverter}.
+ *     <li>{@link java.util.Map Java Map} types via {@link MapConverter}.
+ * </ul>
+ *
+ * <p>An attempt to convert a collection type that is not listed above, or to convert a supported
+ * collection type back to a Java object, results in a {@link UnsupportedOperationException}.
```
alexander-yevsyukov, line 75: `a` -> `an`

alexander-yevsyukov (approved):
LGTM, with the one doc-related comment before the merge.

PR #859, opened 2025-03-05
Clarify `(pattern)` option and implement `restoreProtobufEscapes()` util

yevhenii-nadtochii:
This PR does the following:

1. Clarifies docs to `(pattern)` option, mentioning which Regex dialect is expected by the option.
2. Implements `restoreProtobufEscapes()` function that allows to restore originally passed users' string literals, so that these literals could be rendered as literals in source files by `validation`.

PR #860, opened 2025-03-12
Clarify docs to `(validate)` option

yevhenii-nadtochii:
This PR adds more details to specification of the `(validate)` option.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -10,7 +10,7 @@ all modules and does not describe the project structure per-subproject.
  -->
 <groupId>io.spine</groupId>
 <artifactId>base</artifactId>
-<version>2.0.0-SNAPSHOT.301</version>
+<version>2.0.0-SNAPSHOT.302</version>
 
 <inceptionYear>2015</inceptionYear>
 
@@ -98,6 +98,12 @@ all modules and does not describe the project structure per-subproject.
   <dependency>
     <groupId>io.spine</groupId>
     <artifactId>spine-logging-smoke-test</artifactId>
+    <version>2.0.0-SNAPSHOT.242</version>
+    <scope>test</scope>
+  </dependency>
+  <dependency>
+    <groupId>io.spine.tools</groupId>
+    <artifactId>spine-logging-testlib</artifactId>
     <version>2.0.0-SNAPSHOT.242</version>
     <scope>test</scope>
   </dependency>
--- a/src/main/proto/spine/options.proto
+++ b/src/main/proto/spine/options.proto
@@ -211,8 +211,14 @@ extend google.protobuf.FieldOptions {
     //
     //   1. The default instance of `Any` is always valid because there is nothing to unpack
     //      and validate.
-    //   2. Instances with unknown type URLs are considered valid because their Java class cannot
-    //      be determined from `io.spine.KnownTypes`.
+    //   2. Instances with type URLs that are unknown to the application are also valid.
+    //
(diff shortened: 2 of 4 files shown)


pom.xml
```diff
@@ -101,6 +101,12 @@ all modules and does not describe the project structure per-subproject.
     <version>2.0.0-SNAPSHOT.242</version>
     <scope>test</scope>
   </dependency>
+  <dependency>
+    <groupId>io.spine.tools</groupId>
+    <artifactId>spine-logging-testlib</artifactId>
```
armiol, line 106: How comes this dependency suddenly appears here now? :)

src/main/proto/spine/options.proto
```diff
@@ -127,24 +127,32 @@ extend google.protobuf.FieldOptions {
...
-    //   1. Message fields.
-    //   2. Map fields with message types as values.
-    //   3. Repeated fields of message types.
+    // When set to `true`, the field is valid only if its value satisfies the validation
+    // constraints defined in the corresponding message type:
     //
-    // When set to `true`, the marked field is considered valid only if the field's value satisfies
-    // the validation constraints defined in the corresponding message type. This means:
+    //   1. For singular message fields: the message must meet its constraints.
+    //      Note: default instances are considered valid even if the message has required fields.
```
armiol, line 142: Please have examples for each of these points.

For this particular point, it would be helpful to illustrate the difference between 
```
MyMessage msg = 42 [(validate) = true]; // implicit `(required) = false`.
```
and
```
MyMessage msg = 42 [(validate) = true, (required) = true];
```

yevhenii-nadtochii:

armiol (changes requested):
@yevhenii-nadtochii been waiting with the comments for some time. Now, that you've requested the review, here they are. Thank you!

pom.xml
yevhenii-nadtochii: I've taken `master`, built it, and `POM.xml` was updated with this dependency. Maybe something related to local caches or stale POM files of the published artifacts.

yevhenii-nadtochii: This is a direct dependency to use `tapConsole { }` I added in #859 as a review suggestion (so, after the bumped version and updated reports).

Hence, that PR was merged to `master` with a stale POM. Sooner or later, we should move this routine to CI :)

>> I followed the clues… right back to me.

yevhenii-nadtochii:
@armiol PTAL

src/main/proto/spine/options.proto
```diff
@@ -127,24 +127,92 @@ extend google.protobuf.FieldOptions {
...
+    //      }
+    //      ```
+    //
+    // If the field contains `google.protobuf.Any`, the option will first attempt to unpack
+    // the enclosed message, and only then validate it. However, unpacking is not always possible:
+    //
+    //   1. The default instance of `Any` is always valid because there is nothing to unpack
+    //      and validate.
+    //   2. Instances with unknown type URLs are considered valid because their Java class cannot
+    //      be determined from `io.spine.KnownTypes`.
```
armiol, line 215: On this level of documentation, mentioning `KnownTypes` triggers more questions than explains the state of matters.

Please try to explain it, as if the readers aren't aware of the internal machinery under the Spine's hood.

armiol (changes requested):
@yevhenii-nadtochii noticed one more thing. 

Other than that, I like the examples!

armiol (approved):

PR #867, opened 2025-04-04
Add `ChoiceOption` and `RequireOption`

yevhenii-nadtochii:
This PR deprecates `(is_required)` and `(required_fields)` in favor of `(choice)` and `(require)` options, which are message-based now.

Please note that I retained `error_msg` field instead of `if_missing` to be in sync with the used convention. All companion fields ~ `(if_missing).error_msg` and message options ~ `(pattern).error_msg` already use this name.

### DIFF
--- a/src/main/proto/spine/options.proto
+++ b/src/main/proto/spine/options.proto
@@ -1404,7 +1404,7 @@ message RangeOption {
 // Controls whether a `oneof` group must always have one of its fields set.
 //
 // Note that unlike the `(required)` constraint, this option supports any field types
-// of the `oneof` cases.
+// within the group cases.
 //
 message ChoiceOption {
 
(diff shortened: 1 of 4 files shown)


yevhenii-nadtochii:

src/main/proto/spine/options.proto
```diff
@@ -1418,3 +1400,88 @@ message RangeOption {
...
+    // Fields of other types are not supported by this option.
+    //
+    // For `oneof` groups, the restrictions above do not apply. Any `oneof` group can be used
+    // without considering the field types, and its value will be checked directly without
+    // relying on the default values of the fields within the group.
+    //
+    // Example: defining alternative required fields.
+    //
+    //     message PersonName {
+    //         option (required_field) = "given_name | honorific_prefix & family_name";
```
armiol, line 1463: This is an old text, isn't it?

armiol (changes requested):

armiol (approved):

PR #868, opened 2025-04-10
Introduce the `format` module

alexander-yevsyukov:
This PR introduces the `spine-format` artifact which is based on the parsing utilities adopted in ProtoData. 

As we migrating to the [Spine Compiler API](https://github.com/SpineEventEngine/compiler), we review the existing ProtoData code. The utilities for supporting parsing of various file formats are reusable. So, we put them into the `base` repository, introducing the `spine-format` artifact. So, the project turned from single-module to multi-module. Hence the number of changed files. Changes and new features are described in the sections below.

### Transforming proto enum `Format`

The Protobuf-based enumeration called `Format` in ProtoData was transformed into handcrafted quasi-enum Kotlin code. There is no much sense in having a Protobuf enum for parsing purposes because Protobuf already carries the knowledge about the format even when a type packed into `Any`. We need to know about a format when we need to parse some input.

Now the `io.spine.format.Format` is `sealed` class containing nested `object`s derived from it. 
This layout resembles an enumeration. We cannot use plain Kotlin `enum` here because we need generic parameters for formats. These generic parameters serve as an upper bound for types that a format supports.

### Format segregation

It turned out that Jackson cannot work with native Protobuf types. This is so because it's based on [other libraries](https://github.com/FasterXML/jackson-dataformats-binary/blob/2.19/protobuf/README.md#dependencies).

Therefore we have the following segregation of formats:
 * `Format.ProtoBinary` and `Format.ProtoJson` serve Protobuf types using native features of the Protobuf library wrapped by our convenience API.
 * `Format.Json` and `Format.Yaml` serve plain Java and Kotlin types through Jackson.

### Reading and Writing API

In addition to parsing features supported by ProtoData, `spine-format` now supports writing too. The whole user-facing API is:

 * `parse()`  functions from the `Parse.kt` file.
 * `write()` function from the `Write.kt` file.

### `SPI` for tunning Jackson

The class `JacksonSupport` allows to additionally configure an `ObjectMatter` if it's needed for introduction of a new format.

Also, there is a public `Companion.modules` list which allows to add a Jackson `Module` if it is not automatically discoverable via `ServiceLoader` API.

### `TextReader` was removed

The `TextReader` API which allowed only loading `String` was removed. The `File.readText(): String` extension function provided by Kotlin is enough for such purposes.

If or when we need to support parsing other text-based formats, we're likely to use Jackson modules.

### Other notable changes
  * Added experimental API which loads and caches a default instance of a `Message` using `ClassValue` API. Please see `DefaultInstanceValue` object in `MessageExts.kt` for details.
  * `UnknownTypeException` now extends `IllegalStateException` instead of `RuntimeException`.
  * Added `Class<Message>.parser` extension value for allowing easier access to parsing of binary data from variety of sources suppored by the `com.google.protobuf.Parser` interface, in addition to just `InputStream` that our `parse` extension functions offer. 

### Build and dependency changes
  * Dokka was bumped to `2.0.0` with related changes to the build scripts.
  * `DocumentationSettings` object was introduced under `buildSrc/.../kotlin/` for passing constants to the documentation convention scripts.
  * `Jackson` dependency object was refactored to introduce nested object grouping data format dependencies.
  * `Project.findTask()` extension was renamed to `getTask()` to highlight the strictness of the call which throws `IllegalStateException` when a task is not available.
  *

### Deprecations
 * `KClass<CodeGeneratorRequest>.parse` and `parseCodeGeneratorRequest()` were deprecated in favor of `KClass<Message>.parse` introduced in `io.spine.typeBinaryExts.kt`.
 * `File.toUnix()` was deprecated in favor of `File.toUnixPath(): String`. The deprecated extension did not really work because of forced path normalization performed by standard Java runtime.


format/src/main/kotlin/io/spine/format/parse/ProtobufParser.kt
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    /**
+     * Deserializes the given bytes into a message of the specified class.
+     */
+    abstract fun doParse(source: ByteSource, cls: Class<out Message>): Message
+}
+
+/**
+ * Settings parser for Protobuf messages encoded in the Protobuf binary format.
```
armiol, line 53: I don't understand why this object and the object below that claim they are "settings" parsers? In the scope of `format` library it seems too specific for such a generic type (and also, with quite a generic name).

Please review.

format/src/test/kotlin/io/spine/format/FormatTest.kt
```diff
@@ -0,0 +1,103 @@
...
+
+    @Test
+    fun `parse a file even if it has non-standard extension`() {
+        val nonStandardFile = file.replaceExtension("fiz")
+        write(nonStandardFile, format, instance)
+        val parsed = parse(nonStandardFile, format, instance::class.java)
+        parsed shouldBe instance
+    }
+}
+
```
armiol, line 103: This line is redundant.

format/src/test/kotlin/io/spine/format/JacksonBackedFormatTest.kt
```diff
@@ -0,0 +1,58 @@
...
+
+    override fun createInstance(): UserAccount {
+        return UserAccount.create(UUID.randomUUID().toString())
+    }
+}
+
+data class UserAccount(
+    val id: String,
+    val creationTimestamp: Instant,     // Test `JavaTimeModule`.
+    val emails: ImmutableList<String>,  // Test `GuavaModule`.
```
armiol, line 47: I suggest we have an `EmailAddress` as a type here,
Or, add another field typed as a collection of objects (`nationalities`? `genders`? :) )

format/src/test/kotlin/io/spine/format/ProtoBinarySpec.kt
```diff
@@ -0,0 +1,57 @@
...
+     * Unlike [Format.ProtoJson] an attempt to parse with another type
+     * leads to creating an empty instance of the requested type with
+     * [unknownFields][com.google.protobuf.GeneratedMessage.unknownFields]
+     * populated with the data from the parsed bytes.
+     *
+     * @see io.spine.format.parse.ProtoBinaryParser.doParse
+     * @see ProtoJsonSpec
+     */
+    @Test
+    fun `have required a matching type but it could not`() {
```
armiol, line 50: "Он должен был потребовать подходящий тип, но не смог (потребовать)"?

gradle.properties
```diff
@@ -26,3 +26,5 @@
 
 org.gradle.jvmargs=-Xmx4096m -XX:MaxMetaspaceSize=1024m -XX:+UseParallelGC
 ## -Xdebug -Xrunjdwp:transport=dt_socket,server=y,address=5005,suspend=y
+
+org.jetbrains.dokka.experimental.gradle.pluginMode=V2EnabledWithHelpers
```
armiol, line 30: Please have a comment on this :)

armiol (changes requested):
@alexander-yevsyukov please see my comments.

armiol (approved):

PR #874, opened 2025-04-22
Take the latest `config`

yevhenii-nadtochii:
This PR takes the latest `config`.

The dependencies onto the selected JUnit libs are now forced, while previously we forced the dependency onto the whole JUnit platform.


yevhenii-nadtochii:

armiol (approved):

PR #876, opened 2025-04-30
Restore `MoreCollectionsKt` class

alexander-yevsyukov:
This PR removes the `@file:JvmName("MoreCollections")` in `MoreCollections.kt` because this class is used in CoreJava and its depenencies.

The functions in the file were deprecated in favour of those moved to `MoreIterables` class which is the Java class produced from the `io.spine.collect.IterableExts.kt` file.

Also, latest `config` was applied addressing configuration issues with the `detekt` configuration.


pom.xml
```diff
@@ -164,13 +182,19 @@ all modules and does not describe the project structure per-subproject.
...
     <groupId>org.junit.jupiter</groupId>
     <artifactId>junit-jupiter-engine</artifactId>
-    <version>5.12.2</version>
+    <version>null</version>
+    <scope>test</scope>
+  </dependency>
+  <dependency>
+    <groupId>org.junit.jupiter</groupId>
+    <artifactId>junit-jupiter-params</artifactId>
+    <version>null</version>
```
armiol, line 197: @alexander-yevsyukov please see the `null` versions for JUnit (it's here and above). These are suspicious.

Something is wrong with either the version detection or our configuration (which is even worse).

armiol (approved):
@alexander-yevsyukov LGTM, but please see my comment.

pom.xml
alexander-yevsyukov: This is because of the `platform(JUnit.bom)`. The license reporter probably does not understand it.

alexander-yevsyukov: Have a look at the `BomsPlugin` which was recently introduced. I'm trying to minimize the forces and make it all a bit more lazy and manageable.

PR #902, opened 2025-11-21
[Publishing] Checkout git branch safely

armiol:
During the publishing to GitHub Pages we clone the repository and check out a specific branch, all programmatically.

However, previously, the command looked like this:
```sh
git checkout <branch_name>
```
which failed if there was no such branch.

This PR updates the command to be 

```sh
git checkout -B <branch_name>
```
which creates a branch if it was missing.

### DIFF
--- a/dependencies.md
+++ b/dependencies.md
@@ -1,6 +1,6 @@
 
 
-# Dependencies of `io.spine:spine-annotations:2.0.0-SNAPSHOT.374`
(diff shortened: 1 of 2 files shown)


armiol:
@alexander-yevsyukov PTAL.

alexander-yevsyukov (approved):
LGTM

PR #905, opened 2025-11-21
[Publishing] Specify the remote branch explicitly

armiol:
This changeset makes another attempt to fix the problem with publishing the content to GitHub Pages branch.

This time the approach is to use the branch name tracked by `Repository` instance as the current one, in `push` operation.


alexander-yevsyukov (approved):
LGTM

PR #906, opened 2025-11-21
[Publishing] Simplify publishing to GitHub Pages

armiol:
This changeset simplifies things by using `gh-pages` branch in all cases.

Also, the flow of the Dokka-generated HTML content was fixed.

Please note that the old `<gh-pages-root>/reference` folder was removed manually. This now has to be done for each repository separately.


alexander-yevsyukov (approved):
LGTM

PR #907, opened 2025-11-22
[Publishing] Adjust `gh-pages` branch checkout

armiol:
Previously, the method we used to check out some specific branches was suitable for the branches which did not exist on the remote end. That's why we used `git checkout -B <branch-name>`.

However, now as we moved to using `gh-pages` branch which exists in our repos, `-B` flag did more harm than good, since it removed the tracking of the remote branch. That's why in the recent `publishing` runs the process failed upon trying to `git pull` — there was nowhere to pull from.

This changeset updates the code to address the issue and also adds some debug logging to speed up the investigation of similar issues in the future.

### DIFF
--- a/dependencies.md
+++ b/dependencies.md
@@ -1,6 +1,6 @@
 
 
-# Dependencies of `io.spine:spine-annotations:2.0.0-SNAPSHOT.379`
+# Dependencies of `io.spine:spine-annotations:2.0.0-SNAPSHOT.380`
 
 ## Runtime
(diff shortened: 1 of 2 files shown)


armiol:
@alexander-yevsyukov FYI.

The `publish` is now [successful](https://github.com/SpineEventEngine/base-libraries/actions/runs/19594983647/job/56118393399). 

Also, this is now available (and other modules alike):

https://spine.io/base-libraries/reference/format/

PR #949, opened 2026-06-23
Introduce `QueryCriterion.isEqualTo` (Kotlin-friendly alias for `is`)

alexander-yevsyukov:
## What

Introduces `QueryCriterion.isEqualTo(V)`, carrying the behavior previously
implemented by `is(V)`, and makes `is(V)` delegate to it.

`isEqualTo` is now the canonical "equals" criterion. `is` is documented as a
short form of it — a convenience wrapper for working with a `QueryCriterion`
from Java.

## Why

Calling `is(...)` from **Kotlin** is awkward: `is` is a reserved word there, so
the call has to be wrapped in backticks:

```kotlin
builder.where(column).`is`(value)      // before — backticks required
builder.where(column).isEqualTo(value) // now — reads cleanly
```

`isEqualTo` reads naturally from both Java and Kotlin, while `is` stays
available (and convenient) for Java callers.

## Changes

- `QueryCriterion.isEqualTo(V)` — new canonical method (null-check + `EQUALS`
  parameter), inherited by `EntityCriterion` and `RecordCriterion`.
- `QueryCriterion.is(V)` — now delegates to `isEqualTo`; KDoc/Javadoc updated to
  describe it as the Java-convenient short form and to point Kotlin callers to
  `isEqualTo`.
- `RecordQueryBuilderTest` — new `IsEqualToAlias` tests: `isEqualTo` appends an
  `EQUALS` parameter, and yields the same parameter as `is`.

`is` is intentionally **not** deprecated — it remains a first-class Java-side
short form.

## Build fix (unrelated pre-existing CI flake)

While verifying this PR, CI's **Build on Ubuntu** failed with ~30–80 failures in
packages unrelated to this change (`io.spine.type`, `io.spine.base`,
`io.spine.protobuf`, `io.spine.code`) — `io.spine.type.KnownTypes` loading an
incomplete descriptor set. **Zero** `io.spine.query` tests were affected.

Root cause (reproduced deterministically, independent of the `isEqualTo`
change): `generateProto`/`generateTestProto` delete `com.google` files from
their own output directory in a `doLast` action. That mutation is not reproduced
when the task is restored from the **Gradle build cache**, so a cached
restoration leaves an inconsistent descriptor set on the test classpath and
`KnownTypes` loads a partial type set. This is why `master` (cache miss →
executed → green) and a branch that restored the cache entry (cache hit → red)
behaved differently — pure cache-population order, not the code change.

Fix: opt those tasks out of the build cache (`outputs.cacheIf { false }`) so they
always execute and the deletion is always applied. Verified: with the cache
enabled, `clean build` now passes 3/3 (previously failed deterministically when
`generateTestProto` was restored from cache).

## Testing

- `./gradlew :base:test --tests "io.spine.query.*"` → **133/133 pass** (incl. the
  two new tests).
- `./gradlew clean build` (cache enabled) → green and deterministic after the
  build fix.

## Review notes

- `spine-code-review` and `review-docs` reviewed the `isEqualTo` change;
  `spine-code-review` reviewed the build fix — all APPROVE.
- Version bumped `2.0.0-SNAPSHOT.420` → `2.0.0-SNAPSHOT.421`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/base/src/test/java/io/spine/query/RecordQueryBuilderTest.java
+++ b/base/src/test/java/io/spine/query/RecordQueryBuilderTest.java
@@ -1,11 +1,11 @@
 /*
- * Copyright 2022, TeamDev. All rights reserved.
+ * Copyright 2026, TeamDev. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- * http://www.apache.org/licenses/LICENSE-2.0
+ * https://www.apache.org/licenses/LICENSE-2.0
  *
  * Redistribution and use in source and/or binary forms, with or without
  * modification, must retain the above copyright notice and the following
@@ -251,6 +251,38 @@ class RecordQueryBuilderTest {
             var query = builder.build();
             var actualBuilder = query.toBuilder();
             assertThat(actualBuilder).isSameInstanceAs(builder);
+        }
+    }
+
+    @Nested
+    @DisplayName("treat `is` as a short form of `isEqualTo`")
+    final class IsEqualToAlias {
+
+        @Test
+        @DisplayName("appending an `EQUALS` parameter via `isEqualTo`")
+        void appendEqualsParameter() {
+            var isinValue = "JP 3496600002";
+            var parameters = queryManufacturer()
+                    .where(isin).isEqualTo(isinValue)
+                    .predicate()
+                    .parameters();
+            assertThat(parameters).hasSize(1);
+            assertHasParamValue(parameters, isin, EQUALS, isinValue);
+        }
+
+        @Test
+        @DisplayName("producing the same parameter as `is`")
+        void matchIs() {
+            var isinValue = "JP 3496600002";
+            var viaIs = queryManufacturer()
+                    .where(isin).is(isinValue)
+                    .predicate()
+                    .parameters();
+            var viaIsEqualTo = queryManufacturer()
+                    .where(isin).isEqualTo(isinValue)
+                    .predicate()
+                    .parameters();
+            assertThat(viaIsEqualTo).isEqualTo(viaIs);
         }
     }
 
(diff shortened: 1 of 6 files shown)


base/src/test/java/io/spine/query/RecordQueryBuilderTest.java
```diff
@@ -254,6 +254,38 @@ void returnSameBuilder() {
         }
     }
 
+    @Nested
+    @DisplayName("treat `is` as a short form of `isEqualTo`")
+    final class IsEqualToAlias {
```
alexander-yevsyukov: Please rename the class to `IsEqualTo`.

base/src/test/java/io/spine/query/RecordQueryBuilderTest.java
```diff
@@ -254,6 +254,38 @@ void returnSameBuilder() {
         }
     }
 
+    @Nested
+    @DisplayName("treat `is` as a short form of `isEqualTo`")
```
alexander-yevsyukov: Make the display name read: "support filtering by given values".

base/src/test/java/io/spine/query/RecordQueryBuilderTest.java
```diff
@@ -254,6 +254,38 @@ void returnSameBuilder() {
...
+            var parameters = queryManufacturer()
+                    .where(isin).isEqualTo(isinValue)
+                    .predicate()
+                    .parameters();
+            assertThat(parameters).hasSize(1);
+            assertHasParamValue(parameters, isin, EQUALS, isinValue);
+        }
+
+        @Test
+        @DisplayName("producing the same parameter as `is`")
```
alexander-yevsyukov: Please change the display name to: "providing `is` alias for `isEqualTo`".

alexander-yevsyukov (commented):
@claude[agent], please see my comments.

armiol (approved):

